Issue #1, opened 2017-02-13
import not working after local install


sjsrey:
After installing with
```
python setup.py install
```
imports are failing
```
Python 2.7.13 |Continuum Analytics, Inc.| (default, Dec 20 2016, 23:09:15)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import libpysal
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named libpysal
```

sjsrey:
Further debugging shows that `python setup.py develop` will get a working import.

However, the egg looks to be broken.

PR #3, opened 2017-02-13
add package data globbing

ljwolf:
this adds package globbing to the setup. it should make the examples work in the installation from egg.


ljwolf:
this also fixes #1

PR #11, opened 2017-05-07
adjacency list tooling (prep for wombling/disparity analysis module)

ljwolf:
Hey, I **finally** got this stuff ready to go. It was finishing the unittests and [making a clean example notebook](https://gist.github.com/ljwolf/6673b2bc7b5a45e3b6183c4240037b16) that took the longest. Always hardest to get the thread through the eye than to find the kit, I guess. 

I'd recommend checking out the notebook to see what's enabled by these tools, as well as what I plan on building up.

### DIFF
--- a/libpysal/weights/adjtools.py
+++ b/libpysal/weights/adjtools.py
@@ -1,0 +1,154 @@
+import numpy as np
+
+def adjlist_apply(vec, W=None, alist=None, func=np.subtract, skip_verify=False):
+    """
+    apply a function to an adajcency list, getting an adjacency list and result.
+
+    Parameters
+    -----------
+    vec     :   iterable
+                an N-length iterable to apply ``func'' to. 
+    W       :   pysal.weights.W object
+                a weights object that provides adjacency information
+    alist   :   pandas DataFrame
+                a table containing an adajacency list representation of a W matrix
+    func    :   callable
+                a function taking two arguments and returning a single argument. This will be
+                evaluated for every (focal, neighbor) pair, or each row of the adjacency list. 
+                Examples include: ``lambda x,y: x < y'', ``np.subtract''
+    skip_verify: bool
+                Whether or not to skip verifying that the W is the same as an adjacency list. 
+                Do this if you are certain the adjacency list and W agree and would like to
+                avoid re-instantiating a W from the adjacency list. 
+
+    Returns
+    -------
+    an adjacency list (or modifies alist inplace) with the function applied to each row. 
+    """
+    try:
+        import pandas as pd
+    except ImportError:
+        raise ImportError('pandas must be installed to use this function')
+    W,alist = _get_W_and_alist(W, alist, skip_verify=skip_verify)
+    vec = np.asarray(vec).flatten()
+    table = pd.DataFrame(np.vstack((W.id_order, vec)).T, columns=('id', 'att'))
+    alist_atts = pd.merge(alist, table, how='left', 
+                          left_on='focal', right_on='id')
+    alist_atts = pd.merge(alist_atts, table, how='left', 
+                          left_on='neighbor', right_on='id', suffixes=('_focal','_neighbor'))
+    alist_atts.drop(['id_focal', 'id_neighbor'], axis=1, inplace=True)
+    alist_atts[func.__name__] = alist_atts[['att_focal', 'att_neighbor']].apply(lambda x: func(x.att_focal, 
+                                                                                               x.att_neighbor), axis=1)
+    return alist_atts
+
+def _get_W_and_alist(W, alist, skip_verify=False):
+    """
+    Either:
+    1. compute a W from an alist
+    2. adjacencylist from a W
+    3. raise ValueError if neither are provided,
+    4. raise AssertionError if both W and adjlist are provided and don't match.
+    If this completes successfully, the W/adjlist will both be returned and are checked for equality.
+    """
+    if (alist is None) and (W is not None):
+        alist = W.to_adjlist()
+    elif (W is None) and (alist is not None):
+        from .weights import W
+        W = W.from_adjlist(alist)
+    elif (W is None) and (alist is None):
+        raise ValueError('Either W or Adjacency List must be provided')
+    elif (W is not None) and (alist is not None) and (not skip_verify):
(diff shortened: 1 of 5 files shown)


libpysal/weights/weights.py
```diff
@@ -212,6 +213,72 @@ def from_shapefile(cls, *args, **kwargs):
...
+
+        Parameters
+        -----------
+        remove_symmetric    :   bool
+                            whether or not to remove ``symmetric'' entries. If the W is symmetric,
+                            a standard ``directed'' adjacency list will contain both the forward and
+                            backward links by default because adjacency lists are a directed
+                            graph representation. If this is True, a W created from this adjacency list
+                            **MAY NOT BE THE SAME** as the original W. If you would like to 
+                            consider (1,2) and (2,1), set this to True.
```
sjsrey: Should this be "set this to False" or "Leave as False"?

ljwolf:
multivariate `adjlist_apply` is added and should pass unittesting. It will compute a multivariate distance vector between adjacent sites.

libpysal/weights/weights.py
ljwolf: "if you would like to consider (1,2) and (2,1) as the same link, set this to True" was what I had intended and forgot that "as the same link" part. 

I like "if you would like to consider (1,2) and (2,1) as distinct links, leave this as False" better.

ljwolf:
As far as inherently asymmetric `W`, I think everything should work just fine for them, as long as you don't `remove_symmetric` on them. So like, a case with:
```
focal | neighbor | weight
------------------------------------
A     |      B   | 1
B     |      A   | 1
C     |      A   | 1 
```
if you do `W.from_adjlist` on that, it'll build the correct, asymmetric `W` and `to_adjlist` should return that adjacency list when sorted on `focal`.

like, you can build a `W` correctly from a de-duped Queen/Rook, it's just not the "correct" W. it's an asymmetric, arbitrarily-directed W, and you can cycle to and from `W` and adjlists on that just fine.

libpysal/weights/adjtools.py
```diff
@@ -0,0 +1,198 @@
...
+            to_drop = adjlist.query('focal == {} and neighbor == {}'
+                                    .format(*tupe))
+        elif (tuple(reversed(tupe)) in undirected):
+            to_drop = adjlist.query('focal == {} and neighbor == {}'
+                                    .format(*tuple(reversed(tupe))))
+        else:
+            undirected.update((tupe,))
+            undirected.update((tuple(reversed(tupe),)))
+            continue
+        adjlist.drop(to_drop.index, inplace=True)
```
carsonfarmer: Are you sure inplace operations will be what the user is expecting? At the very least, I think it'd be worth noting this in the doc string.

libpysal/weights/adjtools.py
```diff
@@ -0,0 +1,198 @@
...
+    W, alist = _get_W_and_alist(W,alist)
+    fnames = set([f.__name__ for f in funcs])
+    for i, (column, function) in enumerate(zip(data.T, funcs)):
+        alist = adjlist_apply(X=column, W=W, alist=alist, skip_verify=True)
+        alist.drop(['att_focal', 'att_neighbor'], axis=1, inplace=True)
+        alist = alist.rename(columns={function.__name__ :'_'.join((function.__name__, names[i]))})
+        fnames.update((function.__name__,))
+    return alist
+
+def filter_adjlist(adjlist, focal_col = 'focal', neighbor_col = 'neighbor'):
```
carsonfarmer: This could be better 'documented' in the notebook. It took me going back and forth a few times to see what was going on with this... maybe just me though ;)

libpysal/weights/weights.py
```diff
@@ -212,6 +213,128 @@ def from_shapefile(cls, *args, **kwargs):
...
+        try:
+            import pandas as pd
+        except ImportError:
+            raise ImportError('pandas must be installed to use this method')
+        adjlist = pd.DataFrame(((idx, n,w) for idx, neighb in self 
+                                           for n,w in neighb.items()),
+                               columns = ('focal', 'neighbor', 'weight'))
+        return adjtools.filter_adjlist(adjlist) if remove_symmetric else adjlist
+
+    def to_nx(self):
```
carsonfarmer: While nx is the normal import name for networkx, might it be better to use the full name (`to_networkx`) here to avoid confusion?

carsonfarmer: P.S. This is an awesome feature 👍

libpysal/weights/weights.py
```diff
@@ -212,6 +213,128 @@ def from_shapefile(cls, *args, **kwargs):
...
+        tuple_list = []
+        for k,nw in self:
+            for n,w in nw.items():
+                tuple_list.append((k,n,w))
+        G = nx.Graph()
+        G.add_weighted_edges_from(ebunch = tuple_list)
+        return G
+    
+    @classmethod
+    def from_nx(cls, graph, weight_col='weight'):
```
carsonfarmer: See comment above

libpysal/weights/util.py
```diff
@@ -1283,6 +1283,42 @@ def isKDTree(obj):
     """
     return any([issubclass(type(obj), KDTYPE) for KDTYPE in KDTREE_TYPES])
 
+def filter_adjlist(adjlist, focal_col = 'focal', neighbor_col = 'neighbor'):
```
carsonfarmer: why same function in two places? They are identical...

libpysal/weights/adjtools.py
ljwolf: Yeah, that could be dodgy. I'll make the updates overwriting copies so that the original never gets changed.

libpysal/weights/weights.py
ljwolf: It's a relatively standard import name, and I used it for convenience. It'd be better to use `to/from_networkx` for most users, I bet.

libpysal/weights/util.py
ljwolf: Good catch! I think I put this here while attempting to avoid an import cycle, but then forgot about it when I fixed the import cycle.

ljwolf:
Should be ready to merge!

PR #13, opened 2017-05-23
add symmetrize for KNNW

ljwolf:
2. [x] This pull request is directed to the `pysal/dev` branch.
   [docstrings](https://en.wikipedia.org/wiki/Docstring#Python) *and
   [unittests](https://docs.python.org/2/library/unittest.html)? * (no unittests yet)
4. [x] You have [assigned a
I'm opening this so I don't forget this bit of code I had to write.


sjsrey:
Would this be something to have in weights more broadly, like maybe a method of W, and not just for KNN? So for example, if someone runs a symmetry check on say a contiguity W they constructed (maybe via a GAL they generated via visual inspection) only to find that there were asymmetries - this could be suggested as a way to fix those asymmetries.

Also, maybe we need to raise a warning when it is used on KNN since  the weights may no longer be KNN?

ljwolf:
Yeah, thinking about it, I don't see why it should be specific to KNN. 

I think since it returns a new W and uses `deepcopy` to avoid modifying the original, it'd be okay to not warn? Maybe changing the name from `symmetrize` to `make_symmetric` or something would help with that.

ljwolf:
I promoted this to the parent class and made it have an inplace flag. 

still needs unittesting.

ljwolf:
This also has commit [a9f1946](https://github.com/pysal/libpysal/pull/13/commits/a9f194608d30827d7a841c0d0fa13ac6eb5edc14) addressing #15

ljwolf:
#16 did, but I just didn't rebase these before making the changes. regardless, they should be merge-able once passed.

PR #20, opened 2017-06-27
higher_order typedispatch fix

ljwolf:
I'm quite frustrated I clobbered what looked like quite a bit of work in [pysal/#922](https://github.com/pysal/pysal/pull/922). 

I'm pushing this here because now Github won't let me resurrect my dev branch on pysal proper without just pushing that PR again. 

This fixes [pysal/#944](https://github.com/pysal/pysal/issues/944) by allowing any binary W to be higher-ordered. Also, it enforces that the source weights are binary. 

We need to finish this reorg soon & plug all those tickets up when we can :frowning_face:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
 script:
   - python setup.py sdist >/dev/null
   - echo "check_stable=False" >libpysal/config.py
-  - nosetests --with-coverage --cover-package=libpysal;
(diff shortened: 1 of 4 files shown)


ljwolf:
@sjsrey if this is good, this is an update of pysal-[#946](https://github.com/pysal/pysal/pull/946)

I suggest we merge the stuff (bugfixes/enhancements) here & backport at scipy?

PR #25, opened 2017-11-03
bump stable date

ljwolf:
this bumps the stable release date in the version.py file.


ljwolf:
@weikang9009 good catch on the year of that date!

Issue #26, opened 2017-11-14
two modules “Wsets.py” and "util.py" depend on each other


weikang9009:
Function `WSP2W` is defined in [**util.py**](https://github.com/pysal/libpysal/blob/master/libpysal/weights/util.py#L783) which imports the module **Wsets.py**
However, in **Wsets.py**, [WSP2W](https://github.com/pysal/libpysal/blob/master/libpysal/weights/Wsets.py#L530]) is called in the function `w_clip` without prior declaration or import. We would encounter the following error if `outSP` is set False:

![image](https://user-images.githubusercontent.com/7359284/32810537-9ee4d9c4-c958-11e7-89af-03cdd56e8ab1.png)

Since **util.py** is dependent on **Wsets.py**, we cannot import `WSP2W` from **util.py**. Any idea how to resolve this?

sjsrey:
Not sure why the [import](https://github.com/pysal/libpysal/blob/master/libpysal/weights/Wsets.py#L7) is commented out?

PR #27, opened 2017-11-14
update doctests in weights module to use libpysal instead of pysal

weikang9009:
This PR is to update doctests in `weights` module to use `libpysal` instead of `pysal`. Now all the doctests in `weights` module are passed. The [doctest](https://github.com/pysal/libpysal/blob/master/libpysal/weights/Wsets.py#L490) involving the mutual dependence issue #26 is commented for now and would be uncommented once the issue is resolved.

### DIFF
--- a/libpysal/weights/Contiguity.py
+++ b/libpysal/weights/Contiguity.py
@@ -53,10 +53,11 @@ class Rook(W):
 
         Examples
         --------
-        >>> wr=rook_from_shapefile(pysal.examples.get_path("columbus.shp"), "POLYID")
+        >>> import libpysal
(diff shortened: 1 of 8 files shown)


ljwolf:
Looks good to me. when I run:
```bash
$ git fetch upstream pull/27/head:doctests
$ git checkout doctests
$ nosetests --with-doctests libpysal/weights
```
I get passing doctests. When I don't they fail. 

I've never ran doctests before, so this is real cool.

weikang9009:
The doctests could become more relevant and important when we want to build a documentation site for libpysal using sphinx!

PR #28, opened 2017-11-15
resolve circular import in Wsets and util

weikang9009:
This PR is to resolve the circular import issue #26 of `Wsets` and `util`: module `util` import some functions of `Wsets` while `Wsets` needs the function [**WSP2W**](https://github.com/pysal/libpysal/blob/master/libpysal/weights/Wsets.py#L532) defined in `util`. 

In `Wsets`, import **WSP2W** at the top of the module won't work:
```python
from .util import WSP2W
``` 

Placing the import [at the top of function](https://gist.github.com/datagrok/40bf84d5870c41a77dc6) works.


ljwolf:
Huh, yeah, the genesis of this is strange, unless something's not unittested. 
I like this solution to resolve the import cycle.

weikang9009:
@ljwolf , you are right. There does not exit a unit test for function `w_clip` in Wsets module where`WSP2W` was called. I add a unit test for  `w_clip` in #29.

PR #29, opened 2017-11-16
add unittest for Wset.w_clip

weikang9009:
There does not exit a unit test for function `w_clip`. This PR is to add the unit test.


sjsrey:
good catch.

PR #30, opened 2017-11-21
add a function to attach the nearest neighbor to island

weikang9009:
This PR is to add the function ```attach_islands``` which attaches the nearest neighbor to each island of a given pysal/libpysal spatial weight object.


ljwolf:
Yeah... I think they all should be sets.

We'd have to modify and validate the input `neighbors` dictionary to the `W` instance to make sure this is type-stable. 

Regardless, this looks fine to me.

sjsrey:
@ljwolf i think there is an issue with using sets for the neighbors and/or weights values as for the non-contiguity cases, the unordered nature of sets means we likely have the ids misaligned with the values.

ljwolf:
Ah, yeah, you're right: between the weights dictionary and the neighbors dictionary, they may get out of sync. Good thinking!

PR #36, opened 2018-04-04
Adding Voronoi generator for cg as well as Voronoi weights from 2-d points

sjsrey:
- generate Voronoi polygons for 2-d point arrays
- build voronoiW for neighbors
- notebook example


libpysal/weights/user.py
```diff
@@ -18,7 +19,43 @@
            'threshold_binaryW_from_shapefile', 'threshold_continuousW_from_array',
            'threshold_continuousW_from_shapefile', 'kernelW', 'kernelW_from_shapefile',
            'adaptive_kernelW', 'adaptive_kernelW_from_shapefile',
-           'min_threshold_dist_from_shapefile', 'build_lattice_shapefile']
+           'min_threshold_dist_from_shapefile', 'build_lattice_shapefile', 'voronoiW']
+
+
+def voronoiW(points):
```
ljwolf: From an API perspective, idk if this should go in `user`, since the api is moving away from user... in fact, you can't get `libpysal.weights.user` unless you explicitly import it as:
`from libpysal.weights import voronoi`. 
could we do this as  `queen.from_voronoi`

ljwolf: If we put this in util, then we can have that class method as:
```
def from_voronoi(self, points, **kwargs):
    return self.from_dataframe(util.voronoi_frame(points), **kwargs)
```
would that be ok? I don't want to overload that `from` method, but I really think it's better semantically to keep that all nested underneath the class.

ljwolf:
gist checking with `baltimore` [here](https://gist.github.com/ljwolf/b54abac21c5ba505c45980297809e65b). 

Looks real good!! 

voronoi

The adjacency due to those corner polygons with really distant limits caught me off guard, but when you inspect, that's right. 

The move from `user` to `Queen.from_voronoi` is all that's required? I that's a long time coming, we can just merge this and I can deal with that later.

libpysal/weights/user.py
sjsrey: This is great. Moving away from `libpysal.weights.user` is the better way to go. I put it in there as I just wanted to point out a use case for `libpysal.cg.voronoi` and that was the quickest one I thought of. With other eyes we can come up with other uses, but for weights I like moving it to `util`.

sjsrey:
@ljwolf we could merge this as is, then think through migration out of `user`.

There are a couple of options we could then explore. For example, we could clip by a bounding rectangle to trim and minimize the effects of the distant limits. This might be useful when those distant limits are actually in "real" space, say the generator points are for tracts in one county, but the limits reach into neighboring counties. In those cases, we likely want to trim at the county boundary and redo the region polygons before defining W.

ljwolf:
Sure, yeah, that trim is a future fix.  was only thinking about adding `Queen.from_voronoi` for visibility, since `libpysal.weights.user` is no longer exposed, so this'll be invisible unless you know to look for it. We can solve that later.

knaaptime:
KNN maybe?

sjsrey:
@darribas sorry for merging before you commented. I saw an earlier email responder indicating you were on some university trip so i reckoned you were out for a bit.

I agree on the next steps for the weights interface should be to get it integrated into the API.

PR #44, opened 2018-05-12
close the door on 2 for libpysal

ljwolf:
with this, we're commiting the auto-converted code for libpysal that only works in two. 

If someone else wants to deal with trying to get libpysal into common subset, they need to:
1. refactor/extract the weights readers from `io` and kill the `FileIO` metaclass
2. see 1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ branches:
 only:
   - master
 python:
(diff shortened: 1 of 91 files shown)


sjsrey:
Happily merging and moving onto the future!

PR #45, opened 2018-05-25
add guerry example dataset

slumnitz:
Add Guerry example dataset to examples in libpysal

add README.md file with information of content and source

@sjsrey


slumnitz:
There is one test failure, which seems unrelated to this pull request:

```
======================================================================
FAIL: test_voronoi_frames (libpysal.cg.tests.test_voronoi.Voronoi)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/steffie/code/libpysal/libpysal/cg/tests/test_voronoi.py", line 27, in test_voronoi_frames
    import geopandas as df
ImportError: dlopen(/Users/steffie/anaconda3/lib/python3.6/site-packages/fiona/ogrext.cpython-36m-darwin.so, 2): Library not loaded: @rpath/libpng16.16.dylib
  Referenced from: /Users/steffie/anaconda3/lib/libgdal.20.dylib
  Reason: Incompatible library version: libgdal.20.dylib requires version 51.0.0 or later, but libpng16.16.dylib provides version 49.0.0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/steffie/code/libpysal/libpysal/cg/tests/test_voronoi.py", line 30, in test_voronoi_frames
    self.assertTrue(isinstance(region, cg.shapes.Polygon))
AssertionError: False is not true
```

When I run `nosetests .` locally I get the same failure, in both my branch and in the master branch. 

I can import `Guerry` from `libpysal` with the changes here, so that all seems to work.

Please let me know if you don't want to merge this, because then I won't change my notebooks in `splot` yet to import `Guerry` from `libpysal`, but will keep a copy of `Guerry` in `splot/examples/`.

sjsrey:
The failures are due to some underlying change in gdal, which is a dependency for geopandas. Once [this gets sorted](https://github.com/ContinuumIO/anaconda-issues/issues/9449), the libpysal tests should pass.

slumnitz:
seems like the test failure got sorted

Issue #48, opened 2018-06-12
doctests on weights are failing across the board


sjsrey:
```
Ran 736 tests in 14.677s

FAILED (SKIP=50, errors=6, failures=62)

```

sjsrey:
Addressed in #110

PR #50, opened 2018-06-13
Wplot

sjsrey:
This is addressing #48 and fixes the docstring examples for weights.

It was started as I was unable to get w.plot working from the example, and then realized all the docstring examples were in need of updating.


ljwolf:
I will do a more thorough review after my 24 hour odyssey to the US tomorrow, but... 

I don't think we should check in the pickle for libpysal/examples/georgia/XB.p... since that's an array @TaylorOshan  can we dump it in csv?

TaylorOshan:
Its safe to just trash it. No longer using those pickled objects since they were from a much older  version of mgwr. Am currently storing the necessary example files locally in the mgwr directory and need to update the mgwr-related examples in libpysal.

PR #56, opened 2018-06-29
add berlin example

TaylorOshan:
The justification for this PR is to add a new example dataset of airbnb data  for the Prenzlauer Berg neighborhood of Berlin.


ljwolf:
checked this out locally; do we need both the `prenzlauer.csv` and `prenzlauer.shp`? pick one, I think. 

Also @sjsrey, do we want to move to zipping up the shapefiles & using geopandas's `zip://`?

or geopackages?

For instance, on south.shp, if I zip it up, it's about 700kb, but raw the dbf is ~1.1mb and shp is ~800kb. 

[One check (plus a flag, if we want to preserve raw paths)](https://github.com/ljwolf/libpysal/blob/pberg/libpysal/examples/__init__.py#L18) can make `libpysal.examples.get_path` return filepaths for zips with `zip://` in front, which I believe works both on windows and linux/osx.

sjsrey:
@ljwolf I'm +1 on the zip idea.

TaylorOshan:
Removed the csv and and switched to zipped folder for shapefile contents.

ljwolf:
@TaylorOshan I pushed up the zipcheck to your branch so it'd get added to this PR. 

Can you pull this down and make sure it works like you expect? By default, it gives `zip://` in front of any zip in the examples, but will give the raw path if you pass `raw = True` to `get_path`?

By default, nothing should change; `get_path` should give you a path that you can immediately read with geopandas.

TaylorOshan:
@ljwolf Looks looks good to me and seems to behave as intended.

PR #51, opened 2018-06-15
ENH: Neighbor detection for nonplanar enforced polygon collections.

sjsrey:
This addresses https://github.com/pysal/pysal/issues/889#issuecomment-396693495.

PR #57, opened 2018-06-30
Fuzzy contiguity

sjsrey:
This implements fuzzy contiguity detection for all features in a GeoDatatFrame. 

It is intended as a complement to #51 which focuses only on fixing island features. This is a more general treatment in that the geometric operations are applied to all features, not just island features.

### DIFF
--- a/libpysal/weights/util.py
+++ b/libpysal/weights/util.py
@@ -14,13 +14,19 @@ from warnings import warn
 import numbers
 import copy
 from collections import defaultdict
+from ..common import requires
+
+try:
+    import geopandas as gpd
+except ImportError:
+    _warnings.warn('geopandas not available. Some functionality will be disabled.')
 
 __all__ = ['lat2W', 'block_weights', 'comb', 'order', 'higher_order',
            'shimbel', 'remap_ids', 'full2W', 'full', 'WSP2W',
            'insert_diagonal', 'get_ids', 'get_points_array_from_shapefile',
            'min_threshold_distance', 'lat2SW', 'w_local_cluster',
            'higher_order_sp', 'hexLat2W', 'regime_weights', 'attach_islands',
-           'nonplanar_neighbors']
+           'nonplanar_neighbors', 'fuzzy_contiguity']
 
 
 KDTREE_TYPES = [scipy.spatial.KDTree, scipy.spatial.cKDTree]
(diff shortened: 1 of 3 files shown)


libpysal/weights/util.py
```diff
@@ -1469,6 +1475,122 @@ def nonplanar_neighbors(w, geodataframe, tolerance=0.001):
...
+        new_geometry = gpd.GeoSeries([feature.buffer(buffer) for feature in gdf.geometry])
+        gdf['_buffer'] = new_geometry
+        old_geometry_name = gdf.geometry.name
+        gdf.set_geometry('_buffer', inplace=True)
+
+    tree = gdf.sindex
+    neighbors = {}
+    n,k = gdf.shape
+    for i in range(n):
+        geom = gdf['geometry'].iloc[i]
```
ljwolf: This column will not always be called `"geometry"`, but I think the `gdf.geometry` attribute will always be the geometry. So, like:

```python
import pysal as ps
import geopandas as gpd

frame = gpd.read_file(ps.examples.get_path("south.shp"))
frame['the_geom'] = frame.geometry
del frame['geometry']
frame = frame.set_geometry('the_geom')
frame.geometry #still exists, returns geoseries named `"the_geom"`
frame['geometry'] #key errors
```

libpysal/weights/util.py
```diff
@@ -1469,6 +1475,122 @@ def nonplanar_neighbors(w, geodataframe, tolerance=0.001):
...
+        hits = list(tree.intersection(geom.bounds))
+        possible = gdf.iloc[hits]
+        ids = possible.intersects(geom).index.tolist()
+        ids.remove(i)
+        neighbors[i] = ids
```
ljwolf: So, this is only fuzzy-contiguity if `buffering==True & tolerance>0`? Could we just do `buffer=0` and have it be fuzzy when `buffer > 0`, retaining the behavior of `tolerance`?

libpysal/weights/util.py
```diff
@@ -1469,6 +1475,122 @@ def nonplanar_neighbors(w, geodataframe, tolerance=0.001):
...
+        new_geometry = gpd.GeoSeries([feature.buffer(buffer) for feature in gdf.geometry])
+        gdf['_buffer'] = new_geometry
+        old_geometry_name = gdf.geometry.name
+        gdf.set_geometry('_buffer', inplace=True)
+
+    tree = gdf.sindex
+    neighbors = {}
+    n,k = gdf.shape
+    for i in range(n):
+        geom = gdf[gdf.geometry.name].iloc[i]
```
ljwolf: I think this should just be `gdf.geometry.iloc[i]`.

libpysal/weights/util.py
ljwolf: Actually, nvm. this way we get "sane" defaults set with tolerance. Disregard.

sjsrey: I think there are two general cases:

a) sloppy digitizing resulting in intersecting polygons (such that a point could be contained in more than a single polygon).

b) "curb" rather than "centerline" digitizing.


So my logic is:
a) is handled without buffering.

b) is handled with buffering. But turning buffering on can also induce contiguities when there really shouldn't be any (as well as fixing those that were missed by assuming planar enforcement).

since b) is more aggressive, i thought it should not be the default.

ljwolf: Yeah, I agree, I think you made the right choice. I didn't consider these distinct cases initially, but it makes sense.

ljwolf:
if @weikang9009's cool with it, let's merge this as the last substantive feature before the freeze.

PR #58, opened 2018-07-10
[WIP] Alpha shapes (2D) code

darribas:
This is an (in progress) PR to merge code for alpha shapes. Relates to [`pysal/pysal#1036`](https://github.com/pysal/pysal/issues/1036).

Items the PR should have before merge:

- [X] Working code
- [X] Documentation + doc examples
- [X] Unit tests


ljwolf:
Excellent, thanks for posting this! I'll take a look and try to start sketching out some simple regression tests.

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,165 @@
+
+import numba as nb
```
ljwolf: this should use the same stuff we use in mapclassify for jenks:
```python
try:
    from numba import jit
    HAS_JIT = True
except ImportError:
    def jit(function, **kwargs):
        return function
    HAS_JIT = False
from warnings import warn
...
def alpha_shape(xys, alpha):
    warn('This wlil be incredibly slow if you do this')     
```

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,165 @@
...
+    radii_sorted_i = radii.argsort()
+    triangles = triangulation.simplices[radii_sorted_i][::-1]
+    radii = radii[radii_sorted_i][::-1]
+    geoms_prev = alpha_geoms((1/radii.max())-1e-10, triangles, radii, xys)
+    xys_bb = np.array([*xys.min(axis=0), *xys.max(axis=0)])
+    if verbose:
+        print('Step set to %i'%step)
+    for i in range(0, len(radii), step):
+        radi = radii[i]
+        alpha = (1 / radi) - 1e-100
```
ljwolf: if I'm understanding what you mean correctly here, this should be `np.finfo(type(alpha)).eps`, the smallest difference relevant for the alpha parameter supplied.

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,165 @@
...
+    a_pts = triangles[:, 0, :]
+    b_pts = triangles[:, 1, :]
+    c_pts = triangles[:, 2, :]
+    radii = r_circumcircle_triangle(a_pts, b_pts, c_pts)
+    radii[np.isnan(radii)] = 0 # "Line" triangles to be kept for sure
+    del triangles, a_pts, b_pts, c_pts
+    radii_sorted_i = radii.argsort()
+    triangles = triangulation.simplices[radii_sorted_i][::-1]
+    radii = radii[radii_sorted_i][::-1]
+    geoms_prev = alpha_geoms((1/radii.max())-1e-10, triangles, radii, xys)
```
ljwolf: here too, maybe, it should be eps, not an arbitrarily small constant?

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,165 @@
...
+    bc = nb_dist(b, c)
+    ca = nb_dist(c, a)
+
+    num = ab * bc * ca
+    den = np.sqrt( (ab + bc + ca) * \
+                   (bc + ca - ab) * \
+                   (ca + ab - bc) * \
+                   (ab + bc - ca) )
+    if den == 0:
+        print(("Warning: Zero denominator found when calculating"\
```
ljwolf: I think this should raise a ValueError if it's a null result, or it should avoid this print. 

Is this an error state? or just a rare branch?

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,165 @@
...
+                mask[k] = False
+                mask[k+1] = False
+    return faces[mask]
+
+def get_single_faces(triangles_is):
+    num_faces_single = 3
+    num_triangles = triangles_is.shape[0]
+    num_faces = num_triangles * num_faces_single
+    faces = np.zeros((num_faces, 2), np.int_)
+    mask = np.ones((num_faces,), np.bool_)
```
ljwolf: can these use `dtype=np.int_` or `dtype=np.bool_`? just to be explicit about what that second argument is.

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,165 @@
...
+
+    faces = build_faces(faces, triangles_is, 
+                        num_triangles, num_faces_single)
+
+    orderlist = ["x{}".format(i) for i in range(faces.shape[1])]
```
ljwolf: this section is a bit weird; let's chat about what it's doing and see if we can do something a bit more intelligible.

libpysal/cg/alpha_shapes.py
darribas: Addressed in 50f3776dc1807e473b41b91d05a2777131268448

ljwolf:
Make sure to add this to https://en.wikipedia.org/wiki/Alpha_shape

libpysal/cg/alpha_shapes.py
```diff
@@ -0,0 +1,325 @@
...
+        return function
+    HAS_JIT = False
+from warnings import warn
+import numpy as np
+import scipy.spatial as spat
+from shapely.geometry import LineString
+from shapely.ops import polygonize
+from geopandas import GeoSeries
+
+eps = np.finfo(type(alpha)).eps
```
ljwolf: this isn't defined outside of the function in which alpha is passed.

We can define `eps` as `np.finfo(type(float)).eps` if alpha should always be a float?

ljwolf:
last blocker is to do unittests for this

libpysal/cg/alpha_shapes.py
darribas: This is addressed in 603ce88

ljwolf:
OK, there are errors now in the nonplanar neighbors stuff. 

@sjsrey 

since we are currently using `pip install -r requirements_dev.txt`, we get geopandas from pypi, not conda. 
this means that we don't get `libspatialindex` on travis, so all geopandas dataframes have `sindex` as `None`.
Then, when we *use* sindex in `nonplanar_neighbors`, we can't use the `sindex`. 

I would like to fix this by insisting that the sindex exists, and not proceeding if the sindex does not exist. 

In the long run, I think it's critical we get geopandas from conda, not pip. 

I will append these changes here.

ljwolf:
Errors have been resolved. All green.

darribas:
@sjsrey do you wanna go ahead and merge?

PR #65, opened 2018-07-14
add check for disconnected components

knaaptime:
This adds a check and raises a warning if the W matrix is not fully connected. We can use this to add check in cases like https://github.com/pysal/pysal/issues/1037#issuecomment-393549165 for algos like maxp that require a fully connected graph and can raise an error if `W.components >1`


ljwolf:
now this needs to rebase & rerun

ljwolf:
the error persists...

ljwolf:
irrelevant testing errors were resolved by #74, so we're good to go with this if a test is written.

knaaptime:
Great. I’ll finish this test I started

libpysal/weights/weights.py
```diff
@@ -188,6 +198,8 @@ def __init__(self, neighbors, weights=None, id_order=None,
             else:
                 warnings.warn("There are %d disconnected observations" % ni)
```
ljwolf: could we combine these warnings?

libpysal/weights/weights.py
```diff
@@ -158,8 +167,9 @@ class W(object):
     """
 
     def __init__(self, neighbors, weights=None, id_order=None,
-        silent_island_warning=False, ids=None):
```
ljwolf: and these flags?

ljwolf: maybe call it `silent=True`.

PR #70, opened 2018-07-14
Swm

knaaptime:
This brings https://github.com/pysal/pysal/commit/1ba0b9234a9f77fa472f64cf2cb23bdc7c6fa5de into libpysal


ljwolf:
sorry, can you run python 2to3 on this so that it's 3 only? the errors are because basestring isn't a builtin in python3 .

PR #79, opened 2018-07-21
increment version number and add bugfixes, api changes

ljwolf:
this adds api changes and bugfixes


jGaboardi:
@ljwolf I just pulled down your branch and tested the changes. No problems on my end. Would you like me to merge this or wait for another review?

ljwolf:
@jGaboardi Not yet; we need to examine how this hits downstream packages, too. I think specifically `splot` and `spreg` will be affected.

ljwolf:
Perfect example of a great implementation complexity this metaclass thing hides:

If the module `IOHandlers` is not imported at least once, then none of the subclassers are run in `IOHandlers` at initialization. Unfortunately, this means that the `FileIO.FileIO` statement doesn't know anything about our custom readers, so it treats everything as a standard file handle in Python. 

Unfortunately, our tests for some reason do not catch this case; as you can see from the commit log above, tests were green for 30aada5, even though no `dbf` could use its `by_col` method, since we never actually test `by_col` in the `libpysal` module. 

I have added a test for the `by_col` method in `test_FileIO` now.

PR #95, opened 2018-08-08
Ensure consistency in `from .module import *` in components of libpysal

ljwolf:
This does what we chatted about on gitter.im/pysal/pysal today. 

Please double check that this is consistent over my fork. 

If you'd prefer, please [check it out locally.](https://help.github.com/articles/checking-out-pull-requests-locally/) If your reference to the `pysal/pysal` remote repository is called `upstream`, then you can:

```
git fetch upstream pull/95/head:ljw-imports
git checkout ljw-imports
```

### DIFF
--- a/libpysal/weights/user.py
+++ b/libpysal/weights/user.py
@@ -10,12 +10,7 @@ from ..io.FileIO import FileIO as ps_open
 from .. import cg
 import numpy as np
 
-__all__ = ['queen_from_shapefile', 'rook_from_shapefile', 'knnW_from_array',
-           'knnW_from_shapefile', 'threshold_binaryW_from_array',
-           'threshold_binaryW_from_shapefile', 'threshold_continuousW_from_array',
-           'threshold_continuousW_from_shapefile', 'kernelW', 'kernelW_from_shapefile',
-           'adaptive_kernelW', 'adaptive_kernelW_from_shapefile',
-           'min_threshold_dist_from_shapefile', 'build_lattice_shapefile', 'voronoiW']
+__all__ = ['min_threshold_dist_from_shapefile', 'build_lattice_shapefile', 'spw_from_gal']
 
 def spw_from_gal(galfile):
     """
--- a/libpysal/io/__init__.py
+++ b/libpysal/io/__init__.py
@@ -1,3 +1,5 @@
 from . import FileIO
-from . import IOHandlers
+from .Tables import *
+from .IOHandlers import *
+from .util import *
 open = FileIO.FileIO
--- a/libpysal/io/util/__init__.py
+++ b/libpysal/io/util/__init__.py
@@ -1,2 +1,3 @@
 from .wkt import *
+from .wkb import *
 from .shapefile import *
--- a/libpysal/io/util/wkb.py
+++ b/libpysal/io/util/wkb.py
@@ -28,8 +28,8 @@ SOURCE: http://webhelp.esri.com/arcgisserver/9.3/dotNet/index.htm#geodatabases/t
 
 
 """
-from cStringIO import StringIO
-from pysal import cg
+from io import StringIO
+from ... import cg
 import sys
 import struct
 
--- a/libpysal/weights/__init__.py
+++ b/libpysal/weights/__init__.py
@@ -1,8 +1,8 @@
-from .weights import W
-from .Distance import KNN, Kernel, DistanceBand
-from .Contiguity import Queen, Rook
-from . import spintW as spatial_interaction
-from . import util
-from . import user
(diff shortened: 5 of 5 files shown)


libpysal/weights/user.py
```diff
@@ -10,12 +10,7 @@
 from .. import cg
 import numpy as np
 
-__all__ = ['queen_from_shapefile', 'rook_from_shapefile', 'knnW_from_array',
```
weikang9009: I just noticed that a lot of convenient weight building functions were removed. Not sure why though...

weikang9009: They were removed 17 days ago https://github.com/pysal/libpysal/commit/e03a3aa036385d845d6038aabf3d94d61378e356#diff-a467e753fd77aae9068ef14f1f7a56a0

ljwolf: This is not removed, but it is moved and de-duplicated.

All of this functionality is directly accessible within the methods of KNN, DistanceBand, Queen, and Rook. In some cases (e.g. the knn and kernel stuff), this duplicated those implementations, and in some ways behaved subtly different from the classes themselves.

This generated a bug, fixed in #82, because the classes did not match what these were doing. 

To prevent more bugs in the future, I removed all of the duplicates here.

weikang9009: I do not see these functions as duplicate implementations since they were calling the methods of KNN, DistanceBand, Queen, and Rook instead of re-implementing. They are more like wrappers?

If these functions are removed, I wonder whether it would be necessary to have some deprecation warnings? Some of these functions such as `queen_from_shapefile` and `rook_from_shapefile` seem to be widely used.

knaaptime: I wouldn't mind a warning, though I'm happy to see these go because I've long been confused about which to use. Along those lines though, something something a little similar. Since we flatten everything in the new structure I can do both

`lps.weights.Contiguity.Queen.from_dataframe`
and
`lps.weights.Queen.from_dataframe`

I'm in favor of *not* having multiple ways to do something;  it's 'more pythonic',  but more practically its a big source of confusion when following examples... scattered notebooks from different instructors with different preferences, etc.

Personally I think more than one level of the hierarchy can make sense for cases like this. But--if we're going to go flat within weights, then I would vote to remove the levels like `Contiguity` and `Distance` and their analogues elsewhere because if we're not asking users to actually traverse the levels of the API, then those symbolic groupings are superfluous and will do more harm than good (since, by definition, we can't get rid of the higher-level versions)

weikang9009: This goes back to the [`del` discussion](https://github.com/pysal/libpysal/pull/89)

weikang9009: For a completely flat api structure as in seaborn, they do not delete the "middle" layer. We can do 
```python
import seaborn
seaborn.categorical.boxplot
```
as well as 
```python
import seaborn
seaborn.boxplot
```
I think what matters more is how we document and deliver the functions/classes to the general public.

knaaptime: what about then lowercasing the middle layer to make clear that its just part of the namespace and not a class?

ljwolf: I agree; upper-cased files appear like classes and are confusing.

I disagree that these user functions did not duplicate stuff in class `__init__`s. For instance, I think [this](https://github.com/pysal/libpysal/blob/7273a320c4336c00d2b07036c6c20f92b6c8be39/libpysal/weights/user.py#L728) duplicates [this](https://github.com/pysal/libpysal/blob/7273a320c4336c00d2b07036c6c20f92b6c8be39/libpysal/weights/Distance.py#L248). 
1. the class `__init__` already handles arrays
2. the class `__init__` has a broader API that `knnW_from_array` obscures & changes.  
3. the class also has a `.from_array` method, which retains the visual similarity to `user.knnW_from_array`, but resists duplication by passing the array & all options straight down. 
4. Given how long-lived our other deprecation have been, I am loath to add deprecation warnings that simply delay this decision indefinitely into the future.

ljwolf: Sorry; wrong links, those should refer to `Kernel` and `kernelW`

weikang9009: Right. The implementation of Kernel weights does seem a little messy. Thanks for showing them to me!

Then I will vote for the current strategy of deleting the duplicates in `user.py`. Since we are not going to raise a deprecation warning for these deleted functions and some of these could be widely used, it would be great to include the deletion in the changelog of the next release of libpysal.

All the unittests pass on my local machine (doctests do not). I think we are good to go.

ljwolf:
I think so, but let's do that in another specific PR.

PR #96, opened 2018-08-09
remove function regime_weights

weikang9009:
Resolve https://github.com/pysal/libpysal/issues/94


weikang9009:
This is not a unittest for function `regime_weights` , probably because it is basically [calling the other function `block_weights`](https://github.com/pysal/libpysal/blob/master/libpysal/weights/util.py#L262) under the hood.

PR #98, opened 2018-08-15
Lower case module names

sjsrey:
Following pep 8 conventions.


libpysal/weights/contiguity.py
```diff
@@ -336,7 +336,7 @@ def Voronoi(points):
     --------
     >>> np.random.seed(12345)
     >>> points= np.random.random((5,2))*10 + 10
-    >>> w = voronoiW(points)
+    >>> w = Voronoi(points)
```
weikang9009: Should not the name of this function lowercase? And its name is too similar to the other function [voronoi ](https://github.com/pysal/libpysal/blob/master/libpysal/cg/voronoi.py#L16) which determines finite Voronoi diagram for a 2-d point set. It seems to be better to stick to "voronoiW" since it is building a spatial weight?

sjsrey: https://github.com/pysal/libpysal/blob/master/libpysal/weights/Contiguity.py#L314

ljwolf: I think it's fine to name functions as if they're classes, although I guess the "proper" thing to do would be to wrap this as:
```python
class Voronoi(Queen):
    def __init__(self, points, *args, **kwargs):
        region_df, _ = voronoi_frames(points)
        return Queen.from_dataframe(region_df, *args, **kwargs)
```
and make it a "real" class.

weikang9009:
Name of module [Wsets.py](https://github.com/pysal/libpysal/blob/master/libpysal/weights/Wsets.py) might also need to be changed?

ljwolf:
@weikang9009 is definitely right; I aliased that to `set_operations`, and think that's a good name for it if we're cool with renaming things.

weikang9009:
And also some modules in [io/](https://github.com/pysal/libpysal/tree/master/libpysal/io) directory

libpysal/weights/contiguity.py
ljwolf: sorry; a conceptually "smaller" way (and one I like more) that I just thought of to do this is to extend, not subclass, the `Queen` class.
```
class Queen(W): 
    ... # do all the things we used to do here
    @classmethod
    def from_points(cls, points, *args, **kwargs):
        region_df, _ = voronoi_frames(points)
        return cls.from_dataframe(region_df, *args, **kwargs)
```
I believe, but would have to verify, that this is portable to `Rook` without duplication:
```python
class Rook(W):
    ....
    from_points = Queen.from_points
```

weikang9009: I would vote for this!

sjsrey: -1 we are bikeshedding at this point.

sjsrey: or let's move the discussion of the Voronoi class/method to a different issue.

ljwolf: sure, I think that makes more sense than continuing here.

PR #101, opened 2018-08-16
Weights/util/get ids gdf

jGaboardi:

- Add functionality to accept a `geopandas.GeoDataFrame` in `weights.util.get_ids()`
- update doc strings in the function and added an example
   - the function does not `@requires geopandas`, so will the example still be appropriate?
- add test as `weights.tests.test_util.test_get_ids_gdf()`

### DIFF
--- a/libpysal/weights/tests/test_util.py
+++ b/libpysal/weights/tests/test_util.py
@@ -187,12 +187,19 @@ class Testutil(unittest.TestCase):
         w_newdneighborsa = ['c', 'b']
         self.assertEqual(w_newdneighborsa, w_new.neighbors['a'])
 
-    def test_get_ids(self):
+    def test_get_ids_shp(self):
         polyids = util.get_ids(
             examples.get_path('columbus.shp'), "POLYID")
         polyids5 = [1, 2, 3, 4, 5]
         self.assertEqual(polyids5, polyids[:5])
 
+    @unittest.skipIf(not HAS_GEOPANDAS, "Missing geopandas, cannot test get_ids with gdf")
+    def test_get_ids_gdf(self):
+        gdf = gpd.read_file(examples.get_path('columbus.shp'))
+        polyids = list(gdf["POLYID"])
+        polyids5 = [1, 2, 3, 4, 5]
+        self.assertEqual(polyids5, polyids[:5])
+    
     def test_get_points_array_from_shapefile(self):
         xy = util.get_points_array_from_shapefile(
             examples.get_path('juvenile.shp'))
--- a/libpysal/weights/tests/test_util.py
+++ b/libpysal/weights/tests/test_util.py
@@ -196,7 +196,7 @@ class Testutil(unittest.TestCase):
     @unittest.skipIf(not HAS_GEOPANDAS, "Missing geopandas, cannot test get_ids with gdf")
     def test_get_ids_gdf(self):
         gdf = gpd.read_file(examples.get_path('columbus.shp'))
-        polyids = list(gdf["POLYID"])
+        polyids = util.get_ids(gdf, "POLYID")
         polyids5 = [1, 2, 3, 4, 5]
         self.assertEqual(polyids5, polyids[:5])
     
--- a/libpysal/weights/util.py
+++ b/libpysal/weights/util.py
@@ -936,16 +936,20 @@ def remap_ids(w, old2new, id_order=[]):
             return W(new_neigh, new_weights)
 
 
-def get_ids(shapefile, idVariable):
+def get_ids(in_shps, idVariable):
     """
-    Gets the IDs from the DBF file that moves with a given shape file.
+    Gets the IDs from the DBF file that moves with a given shape file or
+    a geopandas.GeoDataFrame.
 
     Parameters
     ----------
-    shapefile    : string
-                   name of a shape file including suffix
-    idVariable   : string
-                   name of a column in the shapefile's DBF to use for ids
+    in_shps      : str or geopandas.GeoDataFrame
+                   The input geographic data. Either
+                   (1) a path to a shapefile including suffix (str); or
+                   (2) a geopandas.GeoDataFrame.
+    idVariable   : str
+                   name of a column in the shapefile's DBF or the 
+                   geopandas.GeoDataFrame to use for ids.
 
     Returns
     -------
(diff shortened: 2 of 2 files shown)


libpysal/weights/util.py
```diff
@@ -959,23 +963,39 @@ def get_ids(shapefile, idVariable):
...
+            var = list(in_shps[idVariable])
         return var
+    
     except IOError:
-        msg = 'The shapefile "%s" appears to be missing its DBF file. The DBF file "%s" could not be found.' % (
-            shapefile, dbname)
+        msg = 'The shapefile "%s" appears to be missing its DBF file. '\
+              + ' The DBF file "%s" could not be found.' % (in_shps, dbname)
         raise IOError(msg)
     except AttributeError:
```
ljwolf: Oh man, this is a great instance of using multiple try/except. But! there is one more exception needed, and here I think pysal maybe mixed up GIS terminology with concepts in Python. 

If a "name" of a given attribute does not exist on a python object (i.e. it's not accessible using `object.attribute`) then an `AttributeError` is needed. 
If a "name" of a given *object* is not found in a look up table (i.e. you can't find it in the set of keys in a dictionary or column name in a table using `object[attribute]`, where attribute conceptually is a name/key/string, not a slice/index/int), then a `KeyError` is needed. 

`pandas` raises a `KeyError` when you grab `dataframe[nonexistent_column]`. But, we `raise AttributeError` when doing `filehandler.by_col[nonexistent_column]`. Conceptually, that's not right in Python; there is no point at which the user requests `object.nonexistent_column`. However, we *do* fail to find `nonexistent_column` in the header dictionary, so that the conceptual equivalent of `object.get(nonexistent_column)` is `None`. To me, this is clearly a `KeyError`, not an `AttributeError` in pythonspeak. 

The confusion arises (I think) from GIS terminology, where a covariate/column in a table is an "attribute" of the shape, so it makes sense to call this an `AttributeError` when an (GIS-sense) attribute is missing. 

So, I think it's fine to change l993 to `except (AttributeError, KeyError)`, which would get the case for the dataframe lookup failing, and the case where we raise an (imo) erroneous `AttributeError`.

jGaboardi: Well, I can't take the credit for implementing the multiple try/except statements, those were already part of the function :smile:. Though I am happy to receive the explanation for different errors. Good feedback.

libpysal/weights/tests/test_util.py
```diff
@@ -187,12 +187,19 @@ def test_remap_ids(self):
...
         self.assertEqual(w_newdneighborsa, w_new.neighbors['a'])
 
-    def test_get_ids(self):
+    def test_get_ids_shp(self):
         polyids = util.get_ids(
             examples.get_path('columbus.shp'), "POLYID")
         polyids5 = [1, 2, 3, 4, 5]
         self.assertEqual(polyids5, polyids[:5])
 
+    @unittest.skipIf(not HAS_GEOPANDAS, "Missing geopandas, cannot test get_ids with gdf")
```
jGaboardi: @ljwolf would this be better to have `util.get_ids()` rather than directly reading in the `gdf["POLYID"]` column?

libpysal/weights/tests/test_util.py
```diff
@@ -187,12 +187,19 @@ def test_remap_ids(self):
...
+    def test_get_ids_gdf(self):
```
jGaboardi: @ljwolf would `test_get_ids_gdf` be better to have `util.get_ids()` rather than directly reading in the `gdf["POLYID"]` column?

ljwolf: yes, it should be testing `util.get_ids` against its known result.

libpysal/weights/util.py
```diff
@@ -959,23 +963,39 @@ def get_ids(shapefile, idVariable):
     >>> polyids = get_ids(libpysal.examples.get_path("columbus.shp"), "POLYID")
     >>> polyids[:5]
     [1, 2, 3, 4, 5]
+    
+    >>> from libpysal.weights.util import get_ids
```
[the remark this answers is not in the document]
jGaboardi: @weikang9009 I made this conceptually as two separate examples, one for the `shp` input and one for the `gdf` input.
```python
>>> from libpysal.weights.util import get_ids
>>> import libpysal
>>> polyids = get_ids(libpysal.examples.get_path("columbus.shp"), "POLYID")
>>> polyids[:5]
[1, 2, 3, 4, 5]
```
and
```python 
>>> from libpysal.weights.util import get_ids
>>> import libpysal
>>> import geopandas as gpd
>>> gdf = gpd.read_file(libpysal.examples.get_path("columbus.shp"))
>>> polyids = gdf["POLYID"]
>>> polyids[:5]
[1, 2, 3, 4, 5]
```

Do you think it would work better as:
```python
>>> from libpysal.weights.util import get_ids
>>> import libpysal
>>> polyids = get_ids(libpysal.examples.get_path("columbus.shp"), "POLYID")
>>> polyids[:5]
[1, 2, 3, 4, 5]
>>> import geopandas as gpd
>>> gdf = gpd.read_file(libpysal.examples.get_path("columbus.shp"))
>>> polyids = gdf["POLYID"]
>>> polyids[:5]
[1, 2, 3, 4, 5]
```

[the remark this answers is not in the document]
weikang9009: Sorry, did not unfold to view the complete code. Either is good!

jGaboardi:
@ljwolf Are there any other concerns with this PR or can it be merged?  (or should we wait for serge review?)

libpysal/weights/tests/test_util.py
jGaboardi: @ljwolf I think you are looking at an older commit ([11e9ed0](https://github.com/pysal/libpysal/pull/101/files/11e9ed023eb93763c2754e0fecd757b972d48107#diff-3378068a2f75204ca6ff319938554ae2)). I corrected the test in [this commit](https://github.com/pysal/libpysal/pull/101/commits/ceb643165ec47d43ae822db2bde2a9d78cc0e429) a few weeks ago.

ljwolf: yep! thanks for calling that out; not sure why my commit view wasn't updated.

PR #105, opened 2018-08-22
Rel.4.0.0

sjsrey:
This is an API change release so we bump to 4.0.0


CHANGELOG.md
```diff
@@ -1,3 +1,74 @@
+# Changes
+
+Version 3.0.8 (2018-08-22)
```
weikang9009: version is wrong

sjsrey: yup. realized that the moment I pushed.

PR #109, opened 2018-09-23
Doctests and start of documentation for libpysal

sjsrey:
- Major pruning of stale code
- Refactoring doctests to pass locally
- Start of sphinx docs for libpysal


sjsrey:
@ljwolf most of this is updating the import scheme and ensuring doctests pass. I prunned some code and removed a few stale modules. One I thought you might have a look at is the removal of [io/geotables/config.py](https://github.com/pysal/libpysal/pull/109/commits/ebfd0744d648c8df81a18c3bd63e0ca378366456).   It imports [ops](https://github.com/pysal/libpysal/blob/8b324f9fe8965558d506434d5e42cce4113ab52f/libpysal/io/geotable/config.py#L1) but there is no ops module there?

ljwolf:
`ops` was a bunch of duplication of `geopandas` back when we were committed to maintaining the option that users should be able to use either `pysal` or `shapely` objects. Not sure when it was removed, but since it is no longer there, `config` should be removed as well.

doc/_static/references.bib
```diff
@@ -0,0 +1,195 @@
+%% This BibTeX bibliography file was created using BibDesk.
```
weikang9009: need to be updated.

doc/api.rst
```diff
@@ -0,0 +1,62 @@
...
+Spatial Weights
+---------------
+
+.. autosummary::
+   :toctree: generated/
+
+   libpysal.weights.adjtools.adjlist_apply
+   libpysal.weights.adjtools.adjlist_map
+   libpysal.weights.adjtools.filter_adjlist
+   libpysal.weights.contiguity.Queen
```
weikang9009: Since the api is structured in such a way that the "middle" layer (modules inside `weights`, `cg`, `io`) is not exposed to users, (e.g. `import libpysal.weights.Queen` is recommended compared to `import libpysal.weights.contiguity.Queen`), I am thinking that it could make the recommended api usage clearer if we hide the middle layers here (sth similar to [seaborn](https://seaborn.pydata.org/api.html)).

A tentative example would be to change this line to `weights.Queen`. Then we would have the following:
image
image

doc/_static/references.bib
sjsrey: Good catch. I took out the references page for libpysal, but we probably want to figure out a system for shared references across all pysal projects. Not sure the best way to do this. Maybe a good item for the next dev meeting (and template)?

doc/api.rst
sjsrey: Another careful reading :+1: 

I agree, this is cleaner, but in working on this, the sanctioned approach was not clear to me (I didn't see any prior documentation to go from  - if we have discussed this, please add the link) and I had to go back to the source code. So the full path import was a reflection of my experience with the code base.

What you suggest is cleaner on the reading the docs (and I like it), but maybe at a cost of not fully grokking the code base. Queen lives in libpysal.weights.contiguity.Queen, but we lift it up to libpysal.weights.Queen.  It is a bit inconsistent since we decided to jettison the .api modules which lifted up things to the top-level, but now we are still lifting (albeit, not to the same extent). 

So if a potential new contributor wants to add an enhancement to say Queen, they are going to search for  libpysal.weights.Queen (using our docs) and wonder what is happening? 

If I had to vote on how we do the docs, I would go with:

 "Explicit is better than implicit." = `libpysal.weights.contiguity.Queen` 
over:
"implicit" = `libpysal.weights.Queen`

@ljwolf @darribas @weikang9009 @knaaptime please add to the discussion on this.

ljwolf: I [did nearly what you describe in this compromise](https://github.com/pysal/libpysal/commit/8525ba5fe124333b88b78068c3dedb804d793107#diff-18cf7dd0605c52e7d2fd64afc1f50aac) back in july. and @weikang9009, @jGaboardi, @knaaptime and I decided to [reverse it](https://github.com/pysal/libpysal/commit/65ab80740d68e7154c15d6f8c13692d3422a3d32) after discussion in august. 

After thinking about it with those folks, I changed my mind because using explicit imports in `__init__.py` *and* using `__all__` means every new thing that you want to expose to users must be added in two places. If I change `foo.py` to add class `Bar`, I now have to add `Bar` to the `__all__` list in `foo.py` *and* add `from foo import ..., Bar` to the end of the module's `__init__.py` import. 

I find `*` imports very difficult to keep track of, but in this context, they're useful to let us define things in one place only.

sjsrey: Thanks for clarifying. I'll update to match the agreed upon scheme for imports.

PR #112, opened 2018-10-06
enh: add doctests to travis (#2)

sjsrey:


ljwolf:
I believe the `shapley_ext.py` errors are coming from the last line, where the `for method in __all__` copies over all the shapely docstrings carte blanche. This is fine to merge as is. 

But, consider:
1. these aren't unit-tested (in `libpysal/cg/tests`)
2. as written, they won't be doctested anymore (they borrow shapely's doctests, so have no control over their passing)
3. we tend to recommend folks go straight to geopandas/shapely for data read/write and GIS in our workshops
4. they're only depended on by the `wkb` reader for testing, which, again `shapely.wkb.loads` should be recommended.
Like at the Friday call, I really think we should delete `libpysal/cg/shapely_ext.py` rather than baking it in to be ignored in testing. I think we should remove functionality if it's not tested, not used, and not recommended to be used.

sjsrey:
I agree with the idea of removing `libpysal/cg/shapely_ext.py`. 
Should we deprecate this first, or just remove? Either way, it looks like, in addition to being used by the `wkb` reader, it is also used in [`cg/ops`](https://github.com/pysal/libpysal/blob/master/libpysal/cg/ops/_shapely.py#L49)?

ljwolf:
good check; that is a soft dependency in cg/ops. 
It's probably a good idea to deprecate... I know my tendency is to want to just remove immediately, but that's not fair to others using it. 

Let's merge this. I'll try to get a PR in adding deprecation warnings to the shapely_ext or make sure they are unittested.

Issue #111, opened 2018-10-04
alphashapes & n<4


ljwolf:
in theory, the "delaunay" triangulation for cases where the number of points is less than 4 is still "known." N=3 is the triangle, n=2 is the line, and n=1 is the point. 

as it stands, `Qhull` will error out when passed a collection with less than four points. It'd be nice if we checked this & returned something sane in these cases, rather than bailing out from qhull.

ljwolf:
I think this is sufficient:

```python
def safe_ashape(xys, **ashape_kws):
    if xys.shape[0] < 4:
        return ops.cascaded_union([geom.Point(xy) for xy in xys]).convex_hull
    return alpha_shape(xys, **ashape_kws)
```

darribas:
I've ran into this issue too. I think that solution might do it. Just a question, would that return a polygon object?

jGaboardi:
Addressed in #115

PR #115, opened 2018-10-07
[WIP] ensure safe returns for small n alphashapes

ljwolf:
this addresses #111.

My testing stack is broken at the moment, so I'm relying on travis for unittests. 

Will tag a reviewer once/if they pass.

### DIFF
--- a/libpysal/cg/alpha_shapes.py
+++ b/libpysal/cg/alpha_shapes.py
@@ -453,6 +453,11 @@ def alpha_shape(xys, alpha):
     '''
     if not HAS_JIT:
         warn("Numba not imported, so alpha shape construction may be slower than expected.")
+    if xys.shape[0] < 4:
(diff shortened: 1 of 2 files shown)


ljwolf:
tests now pass. I've added the small-N alpha shaping to any data with less than 4 points, since qhull should take care of anything with four or more points.

ljwolf:
bumping this... would like to get this into the next libpysal.

ljwolf:
FYI on this, I still run into some kind of hairy behaviour here. The `buffer(0)` trick ensures that all return types are `POLYGON`. But, if you're `n == 2` or `n==1`, then this returns `POLYGON EMPTY`. This isn't the friendliest thing, especially since [`geopandas` doesn't like spatial joints with empty polygons](https://github.com/geopandas/geopandas/issues/944)...

jGaboardi:
hmmmm

Issue #116, opened 2018-10-21
weights.distance.KNN.from_dataframe ignoring radius 


sjsrey:
![2018-10-20-1806-1540084013-sel](https://user-images.githubusercontent.com/118042/47262075-160c8380-d493-11e8-80bb-463700554060.png)
[files.zip](https://github.com/pysal/libpysal/files/2498705/files.zip)

ljwolf:
I remember hitting this in Riverside. I thought I fixed it in the first commit [here](https://github.com/pysal/libpysal/pull/82).

ljwolf:
Hmm, maybe it's that, if `radius` is passed, we also need to set `distance_metric="arc"`

sjsrey:
fixed with #117

PR #117, opened 2018-10-21
Always make spherical KDTrees if radius is passed

ljwolf:
This makes it so that if `radius` is provided, `distance_metric='arc'` is implied. 

This should ensure the API behaves as expected in #116. 

Currently, you have to pass `distance_metric='Arc'` in order for `radius` to be used. This is because, when we pass a `radius` but keep `distance_metric='euclidean'`, `KDTree` ignores the radius, and returns the `KDTree` according to `distance_metric`


ljwolf:
Oh yeah, also, I made it so that the comparison is on `distance_metric.lower()` always, which is in line with the rest of the library so far as I can tell.

sjsrey:
Looks good to me. Thanks for tracking down and fixing.

PR #122, opened 2018-10-30
update interactive examples in inline docstrings

weikang9009:
update interactive examples in the inline docstrings to comply with the api introduced in online docs


libpysal/weights/util.py
```diff
@@ -1151,12 +1136,12 @@ def lat2SW(nrows=3, ncols=5, criterion="rook", row_st=False):
...
     --------
 
-    >>> from libpysal.weights import lat2W
+    >>> from libpysal.weights import lat2SW
     >>> w9 = lat2SW(3,3)
-    >>> w9[0,1]
-    1
-    >>> w9[3,6]
-    1
+    >>> w9[0,1] == 1
```
weikang9009: @jGaboardi this is working locally and on travis ci
If you can review this PR, and merge it if there is not any issue, the python version bumping PR should pass.

jGaboardi: So then I should remove the test adjustment in #154?

```
>>> w9[0,1]
array(1, dtype=int8)
>>> w9[3,6].tolist()
1
```

weikang9009: I think so. The test adjustment in #154 is not passing locally. I am not sure why...
But `numpy.array(1, dtype=numpy.int8)==1 ` actually awalys return `True`.

jGaboardi: hmmmm. The tests for #154 pass locally for me, both for Python `3.6` and `3.7.3` with `numpy 1.16.3`.

jGaboardi: Might it be an issue with your install?

weikang9009: Not sure. Maybe we can have a third person testing that.

jGaboardi: @weikang9009 and I are trying to figure out the relationship between the failing doctest (which I addressed in #154) and `numpy`. We need at least one more dev to weigh in here.

jGaboardi: @weikang9009 #154 has been merged into `master` by Serge.

jGaboardi: #155 reverting to pre-#154 until we get this issue worked out.

sjsrey: I had the same local failures as @weikang9009 so I don't think we have a good solution yet.

sjsrey: I think this is due to scipy going from 1.2.1 to 1.3.0. When I update to the latter I see
```
>>> w9[0,1]
array(1, dtype=int8)
```
but on 1.2.1 it is
```
>>> w9[0,1]
1
```

jGaboardi:
#156

libpysal/weights/util.py
weikang9009: Thanks for solving the mystery. I did not pay attention to the scipy version on my mac. Thought it was a numpy thing. Just noticed [`conda-forge` released `scipy 1.3.0`](https://github.com/conda-forge/scipy-feedstock/commit/d852089026f7b04110af5b67618110b4bef96a3a) and I did not update my local version.

Issue #131, opened 2019-01-21
addressing DeprecationWarning: fromstring()


jGaboardi:
Mimic `DeprecationWarning ` solution found in pysal/pysal#1073

jGaboardi:
#152

jGaboardi:
Closed via #132

PR #132, opened 2019-01-21
addressing DeprecationWarning: fromstring()

jGaboardi:
 - Addressing #131


jGaboardi:
`Python3.6`, `PYSAL_PLUS` failing

jGaboardi:
@sjsrey @sjsrey @weikang9009 @knaaptime 

Do any of you have thoughts on this?

jGaboardi:
#152

.travis.yml
```diff
@@ -24,7 +24,7 @@ before_install:
 install:
   - conda install --yes pip
   - if [ "$PYSAL_PLUS" == true ]; then
-        echo 'plus testing'; pip install -r requirements_plus.txt;
+        echo 'plus testing'; conda install --yes --file requirements_plus_conda.txt; pip install -r requirements_plus_pip.txt;
```
ljwolf: I've missed a few development calls recently; have we confirmed this choice as a library?

ljwolf: I should say, I think this is a good idea, and will help us in the long run to create reproducible environments. Indeed, we should probably move this all the way to an `environmentl.yml`.

jGaboardi: @ljwolf This is actually already [part of the `.travis.yml`](https://github.com/pysal/libpysal/blob/aa7882e7877b962f4269ea86a612dfc58152e5c6/.travis.yml#L25-L32). It must have snuck in since I made this PR back in January.

jGaboardi: yeah, this was part of merge #140

.travis.yml
```diff
@@ -24,7 +24,7 @@ before_install:
...
         else conda install --yes --file requirements.txt;
     fi;
   - conda install --yes geopandas
```
weikang9009: we can put `geopandas` in requirements_plus_conda.txt?

jGaboardi: @weikang9009 [This is already the case](https://github.com/pysal/libpysal/blob/master/requirements_plus_conda.txt). It was merged with #140 .

weikang9009: Cool! I think we should delete this line since  [requirements_dev.txt](https://github.com/pysal/libpysal/blob/master/requirements_dev.txt) also have `geopandas`.

jGaboardi: Rather than removing
```
conda install --yes --file requirements_plus_conda.txt
```
we should remove `geopandas` from `requirements_dev.txt` because that is using `pip` for installation, not `conda`.

Issue #99, opened 2018-08-15
weights.Voronoi is a function, not a class. 


ljwolf:
[This thread](https://github.com/pysal/libpysal/pull/98/files/34a898821a6be25edf54366f07884534d6032209#r210394634) discusses this rough edge. 

Whether or not we should document for the 2.0 release, then change (mandating another subsequent change to the documentation) or change (attempting to avoid spalling out into the rest of the library), then document the 2.0 release is up for grabs.

jGaboardi:
#135

PR #135, opened 2019-02-08
Automatically voronoi input point dataframes to Queen/Rook

ljwolf:
This is some very dank code smell, so there may be a better way of doing this. But, I wanted there to be a `from_dataframe` constructor for the `Voronoi` function. We made this function `Voronoi` look like a class, but it will output either a `Rook` or a `Queen` instance depending on the criterion requested. But, we wanted this "function" to look like a class, since all our other constructors are classes. So, we gave the `Voronoi` function an upper case name, to make it look like the `Queen` and `Rook` constructors. 

Now, I'd really like to add a *method* to the function that calls an additional constructor/accessor to this function, namely the `from_dataframe` method, since popping back down into arrays in a `geopandas`-focused world requires you to unpack the `geodataframe.geometry.x, geodataframe.geometry.y` and then column-stack them together. 

So, this PR adds a method to the `Voronoi` function. I'm not sure if this is the "right" way to solve the problem, which I can see as being said two ways:

#### 1  *we want to keep a distinct constructor `Voronoi` with many different `from_*` constructors, but that returns objects of type `Rook` or `Queen`, not `Voronoi` itself*. 

I've no idea how to do this... I tend to think that a class's `__init__` should return objects of that class, but maybe I'm old fashioned. But, to avoid this "function with methods" hack, we would need to:
1. make `Voronoi` its own subclass of `W`
2. figure out how to make `Voronoi.__init__` return an object that isn't of class(self), but instead of some other class, of which `Voronoi` isn't a subclass. I think we could do this using `Voronoi.__new__`, but I would need to look into that. Basically, we never want an actual instance of a `Voronoi` class. 

#### 2 *we want to push `Voronoi` underneath `Queen` and `Rook`, so that passing points to `Queen` and `Rook` implicitly constructs the contiguity graph of the voronoi diagram*

The more I think about the solution I've implemented in this PR, the more I like 2. 

Please let me know what folks think!

### DIFF
--- a/libpysal/cg/voronoi.py
+++ b/libpysal/cg/voronoi.py
@@ -303,7 +303,7 @@ def clip_voronoi_frames_to_extent(regions, vertices, clip='extent'):
     except ImportError:
         raise ImportError('geopandas is required to clip voronoi cells')
     clipped_regions = overlay(regions, clipper, how='intersection')
-    return clipped_region, clip=clips
+    return clipped_regions
     
     
 
--- a/libpysal/cg/voronoi.py
+++ b/libpysal/cg/voronoi.py
@@ -301,7 +301,7 @@ def clip_voronoi_frames_to_extent(regions, vertices, clip='extent'):
                                                            .convex_hull])
     elif clip.lower() in ('ahull','alpha hull', 'alpha_hull', 
                                'ashape','alpha shape', 'alpha_shape'):
-        from .ashapes import alpha_shape_auto
+        from .alpha_shapes import alpha_shape_auto
         from ..weights import get_points_array
         coordinates = get_points_array(vertices.geometry)
         clipper = geopandas.GeoDataFrame(geometry=[alpha_shape_auto(coordinates)])
--- a/libpysal/cg/voronoi.py
+++ b/libpysal/cg/voronoi.py
@@ -173,7 +173,7 @@ def as_dataframes(regions, vertices, points):
 
     return region_df, point_df
 
-def voronoi_frames(points, radius=None, clip=False, tolerance=.01):
+def voronoi_frames(points, radius=None, clip='extent'):
     """
     Composite helper to return Voronoi regions and generator points as individual dataframes
 
@@ -187,9 +187,21 @@ def voronoi_frames(points, radius=None, clip=False, tolerance=.01):
                   distance to "points at infinity" used in 
                   building voronoi cells
 
-    clip        : bool
-                  whether or not to clip the voronoi cells
-                  back to the original map extent (default: True)
+    clip        : str, or shapely.geometry.Polygon
+                  an overloaded option about how to clip the voronoi
+                  cells. The options are:
+                  - 'none'/None: no clip is applied. Voronoi cells may be arbitrarily
+                                 larger that the source map. Note that this may lead
+                                 to cells that are many orders of magnitude larger
+                                 in extent than the original map. Not recommended. 
+                  - 'bbox'/'extent'/'bounding box': clip the voronoi cells to the 
+                                                    bounding box of the input points.
+                  - 'chull'/'convex hull': clip the voronoi cells to the convex hull
+                                           of the input points.
+                  - 'ashape'/'ahull': clip the voronoi cells to the tightest hull that
+                                      contains all points (e.g. the smallest alphashape,
+                                      using libpysal.cg.alpha_shape_auto)
+                  - Polygon: clip to an arbitrary Polygon
 
     tolerance   : float
                   percent of map width to use to buffer the extent
@@ -230,10 +242,10 @@ def voronoi_frames(points, radius=None, clip=False, tolerance=.01):
     if clip:
         regions = clip_voronoi_frames_to_extent(regions, 
                                                 vertices,
-                                                tolerance=tolerance)
+                                                clip=clip)
     return regions,vertices 
 
-def clip_voronoi_frames_to_extent(regions, vertices, tolerance = .01):
+def clip_voronoi_frames_to_extent(regions, vertices, clip='extent'):
     """
     Arguments
     ---------
@@ -241,30 +253,57 @@ def clip_voronoi_frames_to_extent(regions, vertices, tolerance = .01):
                 dataframe containing voronoi cells to clip
     vertices:   geopandas geodataframe
                 dataframe containing vertices used to build voronoi cells
-    tolerance:  float
-                percent of map width to buffer the bounds of the vertices
-                in order to ensure that vertices are unambiguously
-                contained within regions. Default is .01, or 1 percent
-                of the total map width. 
-    """
-    min_x, min_y, max_x, max_y = vertices.total_bounds
-    try:
-        from shapely.geometry import Polygon
-    except ImportError:
-        raise ImportError('shapely is not found, and is required to'
-                          ' clip voronoi polygons.')
-    try:
-        import geopandas
-    except ImportError:
-        raise ImportError('geopandas is not found, and is required to'
-                          ' clip voronoi polygons')
-    bounding_poly = Polygon([(min_x, min_y), (min_x, max_y), 
-                             (max_x, max_y), (max_x, min_y),
-                             (min_x, min_y)]).buffer(abs(max_x - min_x)*tolerance)
-                            
-    bounding_frame = geopandas.GeoDataFrame(geometry=[bounding_poly])
+    clip        : str, or shapely.geometry.Polygon
+                  an overloaded option about how to clip the voronoi
+                  cells. The options are:
+                  - 'none'/None: no clip is applied. Voronoi cells may be arbitrarily
+                                 larger that the source map. Note that this may lead
+                                 to cells that are many orders of magnitude larger
+                                 in extent than the original map. Not recommended. 
+                  - 'bbox'/'extent'/'bounding box': clip the voronoi cells to the 
+                                                    bounding box of the input points.
+                  - 'chull'/'convex hull': clip the voronoi cells to the convex hull
+                                           of the input points.
+                  - 'ashape'/'ahull': clip the voronoi cells to the tightest hull that
+                                      contains all points (e.g. the smallest alphashape,
+                                      using libpysal.cg.alpha_shape_auto)
+                  - Polygon: clip to an arbitrary shapely.geometry.Polygon
+    
+    Returns
+    -------
+    a geodataframe of clipped voronoi regions
 
-    return geopandas.overlay(regions, bounding_frame, how='intersection')
+    """
+    if clip_type is None:
+        return regions
+    elif clip_type.lower() == 'none':
+        return regions
+    elif clip_type.lower() in ('bounds','bounding box', 'bbox', 'extent'):
+        min_x, min_y, max_x, max_y = vertices.total_bounds 
+        bounding_poly = Polygon([(min_x, min_y), (min_x, max_y), 
+                                 (max_x, max_y), (max_x, min_y),
+                                 (min_x, min_y)])
+        clipper = geopandas.GeoDataFrame(geometry=[bounding_poly])
+    elif clip_type.lower() in ('chull','convex hull', 'convex_hull'):
+        clipper = geopandas.GeoDataFrame(geometry=[vertices.geometry\
+                                                           .unary_union\
+                                                           .convex_hull])
+    elif clip_type.lower() in ('ahull','alpha hull', 'alpha_hull', 
+                               'ashape','alpha shape', 'alpha_shape'):
+        from .ashapes import alpha_shape_auto
+        from ..weights import get_points_array
+        coordinates = get_points_array(vertices.geometry)
+        clipper = geopandas.GeoDataFrame(geometry=[alpha_shape_auto(coordinates)])
+    else:
+        raise ValueError('clip type "{}" not understood. Try one '
+                         ' of the supported options: [None, "extent", '
+                         '"chull", "ahull"]'.format(clip_type))
+    try:
+        from geopandas import overlay
+    except ImportError:
+        raise ImportError('geopandas is required to clip voronoi cells')
+    clipped_regions = overlay(regions, clipper, how='intersection')
+    return clipped_region, clip=clips
     
     
 
--- a/libpysal/cg/voronoi.py
+++ b/libpysal/cg/voronoi.py
@@ -302,7 +302,7 @@ def clip_voronoi_frames_to_extent(regions, vertices, clip='extent'):
     elif clip.lower() in ('ahull','alpha hull', 'alpha_hull', 
                                'ashape','alpha shape', 'alpha_shape'):
         from .alpha_shapes import alpha_shape_auto
-        from ..weights import get_points_array
+        from ..weights.distance import get_points_array
         coordinates = get_points_array(vertices.geometry)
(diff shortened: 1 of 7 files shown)


jGaboardi:
The `from_dataframe*` idea seems like a good idea to me, but I will defer to more senior judgment on this.

> 1. make `Voronoi` its own subclass of `W`

In `pysal` we only generate voronoi diagrams from points (to my knowledge), though a voronoi diagram may be generated from any geometric object ([Okabe et al. 2000: Ch. 3](https://www.worldcat.org/title/spatial-tessellations-concepts-and-applications-of-voronoi-diagrams/oclc/25205363)). Would having this as a subclass for now make it more tractable to incorporate voronoi generation from lines and polygons in the future? If so, this would be a great idea.

* Okabe, A., B. Boots, K. Sugihara, and S. N. Chiu. 2000. *Spatial Tessellations*. 2nd ed. Chichester: John Wiley.

libpysal/weights/contiguity.py
```diff
@@ -347,7 +348,45 @@ def Voronoi(points):
...
+
+    Returns
+    -------
+    w           :   W
+                    instance of spatial weights.
+    """
+    try:
+        x,y = df.geometry.x.values, df.geometry.y.values
+    except ValueError:
+        raise NotImplementedError('Voronoi weights are only'
```
jGaboardi: See [comment](https://github.com/pysal/libpysal/pull/135#issuecomment-461979931) regarding points, lines, polygons.

ljwolf:
If we *do* do weights from voronoi diagrams from other spatial objects, we'd likely still need to have a separate `Voronoi` weights constructor, as well as matching implementations for those geometries in `libpysal.cg.voronoi_frames`. 

But, I still think we'd need to decide whether that weights constructor is a class whose `__init__` doesn't return an object of its own type, or if we're cool with it being a function with methods, like I've done here. .

jGaboardi:
@ljwolf Can we be cool with it like you have done here for now and alter it in the future if need be?

ljwolf:
Yeah, the semantics remain the same.

knaaptime:
the nice thing about the Voronoi pseudoclass is that it makes explicit to the users what's happening. Conceptually, though, i think 2 makes a lot of sense, since anyone using this function is really after Queen or Rook weights. So I think i would vote for 2, as long as there's a demo and users are aware they can pass points to Queen and Rook 

 you do lose the ability to provide this tip, though

`Voronoi weights are only implemented for point geometries. You may consider using df.centroid.`

ljwolf:
I agree that for what we have now (only doing Voronoi for Points), it makes a *ton* of sense to just stuff everything underneath Rook/Queen and dispatch based on the geometry type. 

But, if (in the future) someone wants to get contiguity for the Voronoi diagram of lines/polygons (that are not in a lattice), then that strategy becomes very ambiguous very quickly. 

Thinking about this, I'm on team [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), and would like to just stuff everything inside of Queen/Rook. In the future, to support @jGaboardi's linear/polygonal voronoi diagrams, we'd support a `voronoi=False[True]` flag for any constructor that 
1. checks the geometry type
2. builds a voronoi diagram for all point input and any case where `voronoi=True`,
3. then makes the contiguity graph of that diagram.

ljwolf:
I've stuffed the voronoi constructor under the `Rook`/`Queen` types for point-like shapes. 

This needs:
1. unittesting
2. confirmation in a no-shapely environment.

ljwolf:
I've added a postprocessing pass (optional) that clips voronoi cells to the map extent.

libpysal/cg/voronoi.py
```diff
@@ -215,4 +226,45 @@ def voronoi_frames(points, radius=None):
...
+        raise ImportError('shapely is not found, and is required to'
+                          ' clip voronoi polygons.')
+    try:
+        import geopandas
+    except ImportError:
+        raise ImportError('geopandas is not found, and is required to'
+                          ' clip voronoi polygons')
+    bounding_poly = Polygon([(min_x, min_y), (min_x, max_y), 
+                             (max_x, max_y), (max_x, min_y),
+                             (min_x, min_y)]).buffer(abs(max_x - min_x)*tolerance)
```
sjsrey: If the bounding rectangle is either wide or narrow (i.e., not square) maybe we should buffer on the shorter dimension?

```                             (min_x, min_y)]).buffer(min(abs(max_x - min_x),abs(max_y-min_y))*tolerance)  ```

Comment (ljwolf) [reply libpysal/cg/voronoi.py]: yeah... i've thought about this, and I think the right/best thing to do might be to use our new `alpha_shape_auto` code to restrict the diagram to neighbors within that hull... I'm just finding some odd results with adjacencies at the boundary.. basically, you can get some real nonlocal connections between vertices because of the edge effects of the voronoi lattice.

Comment (ljwolf) [reply libpysal/cg/voronoi.py]: For instance, the stuff I'm using this on is for the IJGIS revision of [spenc](https://github.com/ljwolf/spenc)

is just the difference between the convex hull & the alphahull for some airbnbs in brooklyn: 
test_different_clips

lines in black are clipped to the bounding box. Lines in red are clipped to the convex hull. Lines in yellow are clipped to the alpha shape from `cg.alpha_shape_auto`. 

Zooming into the top left corner of the map, we get surprising nonlocal links (drawn in green below), even for the convex hull-clipped boundaries:
surprising_nonlocal_links_between_dumbo_and_wburg
So... I think the notion of what we intend to model with the delaunay-style neighbor graph is best modeled by clipping to the alpha-hull and using the `Rook` contiguity.

I'd make a graph comparison, but it's pretty dirty, and for some reason the `splot` and `W.plot()` methods are both failing for me right now...

Comment (sjsrey) [reply libpysal/cg/voronoi.py]: I think clipping to the alpha shapes definitely makes for the best solution.

Comment (sjsrey) [reply libpysal/cg/voronoi.py]: But what about the idea of allow for an option on the clipping geometry? Default could be alpha, but we could support other options, hull, bb, general poly passed in by the user?

Comment (ljwolf) [reply libpysal/cg/voronoi.py]: Great idea! I've already got the code written for most of that in order to make the images above. I'll contrib that here for the functionality you describe.

File: libpysal/cg/voronoi.py [hunk 3]
```diff
@@ -230,41 +242,68 @@ def voronoi_frames(points, radius=None, clip=False, tolerance=.01):
...
-    bounding_poly = Polygon([(min_x, min_y), (min_x, max_y), 
-                             (max_x, max_y), (max_x, min_y),
-                             (min_x, min_y)]).buffer(abs(max_x - min_x)*tolerance)
-                            
-    bounding_frame = geopandas.GeoDataFrame(geometry=[bounding_poly])
-
-    return geopandas.overlay(regions, bounding_frame, how='intersection')
+        raise ImportError('geopandas is required to clip voronoi cells')
+    clipped_regions = overlay(regions, clipper, how='intersection')
+    return clipped_region, clip=clips
```
jGaboardi: @ljwolf 
 * I may be missing something, but I am not seeing where `clips` defined.
 * doesn't having an assignment in a return statement raise a `SyntaxError`?

ljwolf: yessir! staged the wrong hunk.

libpysal/cg/voronoi.py
```diff
@@ -215,4 +238,72 @@ def voronoi_frames(points, radius=None):
...
+                                 to cells that are many orders of magnitude larger
+                                 in extent than the original map. Not recommended. 
+                  - 'bbox'/'extent'/'bounding box': clip the voronoi cells to the 
+                                                    bounding box of the input points.
+                  - 'chull'/'convex hull': clip the voronoi cells to the convex hull
+                                           of the input points.
+                  - 'ashape'/'ahull': clip the voronoi cells to the tightest hull that
+                                      contains all points (e.g. the smallest alphashape,
+                                      using libpysal.cg.alpha_shape_auto)
+                  - Polygon: clip to an arbitrary shapely.geometry.Polygon
```
ljwolf: this aint supported yet.

jGaboardi:
#99

knaaptime:
since we're using kernel weights a lot in `segregation` lately, I've been thinking about whether the "inverse" of this idea would be useful too... like if you pass polygons to a distance-based weights constructor it operates on the centroids?

ljwolf:
you should be able to send `DistanceBand.from_dataframe(df.centroid, threshold=2, geom_col='centroid')`?

knaaptime:
ah cool. thats basically what i was doing but in an extra step since i didnt realize the `geom_col` option was there

in that case, disregard

ljwolf:
Uhh wait that's already the behavior of all the distance methods applied to polygons. You should be able to do this no problem:
```python
DistanceBand.from_dataframe(df, threshold=XXX)
```
`get_points_array` [grabs the centroid of polygons](https://github.com/pysal/libpysal/blob/master/libpysal/weights/util.py#L1007), and this is what's used on all of the geometries in `distance.py`

knaaptime:
doh! 🤦‍♂️

didnt catch that in my cursory search before this comment

ljwolf:
Yes, I figured this would be relevant... this started with me looking into the `Delaunay` object for this graph, since that is the dual graph of the voronoi lattice. This code does build on top of the voronoi functionality (@sjsrey's code to build shapely polygons from the [scipy voronoi object](https://github.com/pysal/libpysal/blob/master/libpysal/cg/voronoi.py#L60)). And, it's definitely going to be slower, building the voronoi polygons first and then construct the weights from that.

But, the whole point of this PR is that the scipy voronoi diagram extends to an arbitrary distance, albeit, controlled by a parameter. This PR (and, processing the graph as the contiguity graph from polygons), allows us to acknowledge that **the constructed graph will be different if cells are clipped differently**. [See the comment/discussion above.](https://github.com/pysal/libpysal/pull/135#discussion_r255312575) The graph you've constructed there is actually different from the graph if we clip the voronoi cells to a reasonable extent, like the convex hull of the points, or their bounding box.

ljwolf:
For instance, with this PR:
```python
import numpy, matplotlib.pyplot as plt, libpysal
numpy.random.seed(12345)
points=numpy.random.random((5,2))*10 + 10
w = libpysal.weights.Voronoi(points)
w_noclip = libpysal.weights.Voronoi(points, clip=None)
w.neighbors
# is: {0: [2, 3, 4], 1: [2], 2: [0, 1, 4], 3: [0, 4], 4: [0, 2, 3]}
w_noclip.neighbors
# is: {0: [1, 2, 3, 4], 1: [0, 2], 2: [0, 1, 4], 3: [0, 4], 4: [0, 2, 3]}

region_df, generators = libpysal.cg.voronoi_frames(points, clip='ahull')
region_df_noclip, generators_noclip = libpysal.cg.voronoi_frames(points, clip=None)

f,ax = plt.subplots(1,1,figsize=(10,10))
region_df_noclip.boundary.plot(ax=ax, color='r')
region_df.boundary.plot(ax=ax)
wnoclip.plot(region_df, ax=ax, color='limegreen')
w.plot(region_df,ax=ax)
```
tmp
There, we can see that the blue diagram (clipped to the alpha shape, which is identical to the convex hull in this case) avoids that "extra" link shown in lime green. Clipping can change the graph, but in a lot of applied cases, that Voronoi cell goes *way* far out.

ljwolf:
I totally admit that my preference for clipping the diagram to the alpha shape hull might be too severe. But I think it's important to clip *at least* to the map extent, to avoid some edge effects that we can see in the the discussion topic above. What do folks think?

ljwolf:
On the testing, it still looks like its's mainly environment/install failures on geopandas. I have no idea how to fix that right now...

weikang9009:
> I totally admit that my preference for clipping the diagram to the alpha shape hull might be too severe. But I think it's important to clip _at least_ to the map extent, to avoid some edge effects that we can see in the the discussion topic above. What do folks think?

Map extent sounds more reasonable. Can we check which option Geoda is using so that GeoDa and pysal (default option) can be consistent?

knaaptime:
> On the testing, it still looks like its's mainly environment/install failures on geopandas. I have no idea how to fix that right now...

we're [hitting this](https://github.com/pysal/segregation/issues/110) too. I think it's an apt thing?

ljwolf:
> can we check which option GeoDa is using...
Great idea! it looks like [they are using something close to the map extent.](https://geodacenter.github.io/workbook/4b_dist_weights/lab4b.html#creating-thiessen-polygons-from-a-point-layer)

But, even in their example, that causes some really nonlocal links! I think this is *very* surprising, especially when you get very non-rectangular point patterns. Even in that example I link:

the point in the large cell near Burke Lakefront Airport gets connected with that other point down by Edgewater park... I just don't think that's a smart default.

ljwolf:
> > On the testing, it still looks like its's mainly environment/install failures on geopandas. I have no idea how to fix that right now...
> 
> we're [hitting this](https://github.com/pysal/segregation/issues/110) too. I think it's an apt thing?

yeah.... no idea how to fix this yet :/ tests pass locally for me on this.

weikang9009:
Yeah. This does look surprising. Can we understand this as incorporating the closeness to the lake as a factor in constructing the spatial weight? 😄

ljwolf:
> Yeah. This does look surprising. Can we understand this as incorporating the closeness to the lake as a factor in constructing the spatial weight?

Haha, in this case, sure. I worry in others with non-rectangular shapes, a convenient interpretation may not be so forthcoming. I think the convex hull or ashape is the most reasonable clip, because (like in Brooklyn) you really don't know *how* nonlocal these links will get if we let the boundaries get big. 

To minimize user surprise, we should try to avoid nonlocal links as stringently as possible, and document the case for users who feel they want to relax this condition. I personally would never use this without clipping at least to the convex hull.

weikang9009:
Clipping by the convex hull could result in only one neighbor for points at the boundary (like the small example you showed earlier).

ljwolf:
Yes, definitely right.... And the clip to the bounding box would still retain 2 neighbors.

sjsrey:
> Clipping by the convex hull could result in only one neighbor for points at the boundary (like the small example you showed earlier).

One option would be to expand the hull a bit to internalize the points on the original hull boundary.

sjsrey:
I think we should clip with the default being ashape, and supporting options for hull, bbox, or maybe a user specified clipping polygon (as an optional arg that would take precedence over the clip='ashape' argument).

ljwolf:
The current PR allows for a geometry argument to `clip`. It's in the docstring, iirc? This means that, in total:
1. you can pass a string ('chull','ahull','bbox') to clip to a "known" boundary
2. you can pass your own geometry to clip to, say like a municipal boundary.
3. you can pass None and do no clipping.

weikang9009:
Some tests (on libpysal.weights.contiguity.Voronoi) are failing on travis-CI https://travis-ci.org/pysal/libpysal/jobs/557627273

ljwolf:
Will fix this afternoon

ljwolf:
Ah! that's a doctest failure. Sorry, forgot to update those. Thanks for the catch.

ljwolf:
Green now

sjsrey:
Is this WIP or ready to merge? (Looks great to me, but I didn't want to merge if the WIP is binding).

weikang9009:
Looks good to me!

ljwolf:
It's ready to merge!

PR #150, opened 2019-03-26
[ENH][WIP] Adding a `rasterW` to extract `W` from raster and align values

darribas:
This PR adds a new method to build `W` object from a (potentially masked) 2D array, returning the `W` *and* a 1D vector with the original values aligned to the `W` created. The method is currently housed at `util.py`, together with `lat2W`, etc. Alternative locations might be more desirable but I can't think of a better one.

The motivation for this is to open up the use of spatial weights and related methods such as ESDA/spatial regression to raster data.

**NOTE**: This is work in progress and the PR is open to kick off discussion. Any suggestions of comments are most welcome. If it is decided this should go into `libpysal`, the following would need to be added:

- [ ] Adding tests
- [ ] Adding example on the docstring
- [ ] Adding example to the `notebooks` folder
- [ ] Anything else I'm missing?

### DIFF
--- a/libpysal/weights/util.py
+++ b/libpysal/weights/util.py
@@ -23,7 +23,7 @@ __all__ = ['lat2W', 'block_weights', 'comb', 'order', 'higher_order',
            'insert_diagonal', 'get_ids', 'get_points_array_from_shapefile',
            'min_threshold_distance', 'lat2SW', 'w_local_cluster',
            'higher_order_sp', 'hexLat2W', 'attach_islands',
-           'nonplanar_neighbors', 'fuzzy_contiguity']
+           'nonplanar_neighbors', 'fuzzy_contiguity', 'rasterW']
 
 
 KDTREE_TYPES = [scipy.spatial.KDTree, scipy.spatial.cKDTree]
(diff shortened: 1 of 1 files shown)


ljwolf:
Nothing else you're missing. 

The main code looks fine to me, let's flag for test development!

libpysal/weights/util.py
```diff
@@ -1562,7 +1562,48 @@ def fuzzy_contiguity(gdf, tolerance=0.005, buffering=False, drop=True):
...
+                      Spatial weights matrix connecting. The first observation
+                      corresponds to the first (non-masked) value starting
+                      from the top-left (first row, first column)
+    y               : ndarray
+                      Values from `raster` as a vector of dimension (`w.n` x
+                      None) aligned with `w`
+    '''
+    w = lat2W(*raster.shape, rook=rook, id_type=id_type)
+    y = raster.flatten(order='C')
+    if type(raster) is np.ma.core.MaskedArray:
```
ljwolf: This should use `isinstance()` I think

libpysal/weights/util.py
```diff
@@ -1562,7 +1562,48 @@ def fuzzy_contiguity(gdf, tolerance=0.005, buffering=False, drop=True):
...
+    -------
+    w               : W
+                      Spatial weights matrix connecting. The first observation
+                      corresponds to the first (non-masked) value starting
+                      from the top-left (first row, first column)
+    y               : ndarray
+                      Values from `raster` as a vector of dimension (`w.n` x
+                      None) aligned with `w`
+    '''
+    w = lat2W(*raster.shape, rook=rook, id_type=id_type)
```
sjsrey: I was thinking a generator might be worth looking into for W on rasters? since you know the 4-8 neighbor pattern, why store all of it rather than do it via a lazy eval or something along those lines. Taking the mask into account might introduce some complexities, but maybe the mask is a special case, while the generator would be for the general raster case?

ljwolf: We could do it by defining the relevant properties on the fly and then caching them once computed?

That said, I'm not sure if this is going to save much in the long run? Would need perftests...

sjsrey: Yes, that is what I was thinking in terms of on the fly. 
The context I had in mind was for _large_ rasters where we might be memory bound.

darribas: That's a cool idea, the generator, it'd make some use cases a *lot* more lightweight.

On the other hand, I had in mind this as a bridge to allow ESDA, etc. on rasters. In that context, you would need a full object right? Related to this, I was also thinking for that context it might be better to have `rasterWSP` rather than `rasterW`?

From annecdotal experience, I think the mask will be more the norm rather than the exception as most images will be rotated in some way? Maybe I'm wrong on that though. Even if the full image is not rotated, for cases where you read in only part of the raster from a shape (e.g. [here](https://rasterio.readthedocs.io/en/stable/topics/masking-by-shapefile.html)), the read in element will almost always be masked.

PR #154, opened 2019-05-23
bump supported Python versions and correct lat2SW doctest

jGaboardi:
Bumping  supporting Python versions in TravisCI to match pysal/pysal#1095

{3.5, 3.6} --> {3.6, 3.7}

PR #156, opened 2019-05-25
BUG: fix for scipy bump #154

sjsrey:
Should address #122 as well.


sjsrey:
> +1 on my end. Should we wait for @weikang9009 or go ahead and merge?

let's wait for @weikang9009 as #122 was waiting for a bit for me and there could be other issues she has insights on.

PR #175, opened 2019-08-24
ENH: fromstring has been deprecated

sjsrey:
This would silence verbose warnings in various test suites in downstream packages. [Example](https://travis-ci.org/sjsrey/geosnap/jobs/575622418#L5600).

Documentation on the [deprecation](https://docs.python.org/3/library/array.html#array.array.frombytes).


sjsrey:
I missed #132. I think this handles that, but want to check with @jGaboardi

sjsrey:
Closing as #132 fixed this issue.

Issue #174, opened 2019-07-31
Big tarball


ocefpaf:
The PyPI tarball is ~25 MB, most of is the examples folder.
If the original design is to ship that folder with the tarball please close the issue, if not I can send a PR excluding that folder from the MANIFEST to produce a smaller tarball.

sjsrey:
There is a start of this in my [fetch](https://github.com/sjsrey/libpysal/tree/fetch) branch.

Starting with the largest example `nat` and following the logic of scikit-learn.

We probably have to refactor [examples/__init__.py](https://github.com/sjsrey/libpysal/blob/fetch/libpysal/examples/__init__.py) so `file_2_dir` can be updated after a fetch.

Was thinking anything > 1m has to go in the fetch category (and will be removed from the source distribution).

sjsrey:
![](https://github.trello.services/images/mini-trello-icon.png) [Fetch](https://trello.com/c/EbMY3vct/1-fetch)

jGaboardi:
@sjsrey Do we need a trello account to access that link?

sjsrey:
Addressed in #176

PR #176, opened 2019-08-24
Fetch

sjsrey:
This would remove some of the larger example datasets out of the source distribution and provide for remote fetching of those examples when needed.

It is a start towards addressing #174.

This would have implications for any packages that rely on these larger datasets for integration testing. An example of how this could be done is [here](https://github.com/pysal/libpysal/commit/cfd200d0447914b369c16694e218f485d43ad037 ).

This has started with the nat and south datasets.

There is some refinement that needs to be done, but I wanted to get feedback on the general direction.


sjsrey:
At this point, we reduce the size of the source distribution from 25mb to 3.7mb.

weikang9009:
One potential issue is related to the `available` function. 
When I import `libpysal` and call this function for the first time, the large datasets are displayed as "None":
image

After I fetch one large dataset ("sacramento2") and call this function again, the other unfetched datasets are gone:
image

Is this behavior intended? If so, I think it makes sense to provide a message reminding the users that we have other example datasets available for fetch?

sjsrey:
> One potential issue is related to the `available` function.
> When I import `libpysal` and call this function for the first time, the large datasets are displayed as "None":
> image
> 
> After I fetch one large dataset ("sacramento2") and call this function again, the other unfetched datasets are gone:
> image
> 
> Is this behavior intended? If so, I think it makes sense to provide a message reminding the users that we have other example datasets available for fetch?

It isn't the intended behavior. I think it might be that those stale directories were around, but empty (since the descriptions are all None). That is a guess though.

Great idea on adding some information about the remote datasets.

ljwolf:
from the Sept. 2019 development call, we thought it might be necessary to add json with meta data for all examples that allows us to distinguish local and remote datasets and allows for emphemeral use of a remote.

taken from [development call notes](https://hackmd.io/CTUWbJgxQjWvgy9xcw0Eog?both)

sjsrey:
It has been some time since the last time this pr was looked at. I refactored the approach pretty substantially, but I think this is ready to go. 

It would be very helpful if maintainers of upstream packages could test against this to see if any breakage is introduced.

.travis.yml
```diff
@@ -28,6 +28,8 @@ script:
   - python setup.py sdist >/dev/null
   - echo "check_stable=False" >libpysal/config.py
   - python setup.py install
+  - python -c 'import libpysal; libpysal.examples.fetch_all()'
```
weikang9009: Will adding this slow down the travis testing for libpysal? Maybe only fetching those which are used by libpysal tests?

.travis.yml
```diff
@@ -28,6 +28,8 @@ script:
...
+  - rm -rf /home/travis/pysal_data/StLouis
```
weikang9009: Any particular reason why only this data set is removed?

.travis.yml
sjsrey: Good point. Will modify.

.travis.yml
sjsrey: Yes, we added some files in the builtin dataset, as well as modified some of the original files (added attributes) so the original one causes problems if installed with the builtin.

.travis.yml
sjsrey: After further thought, it doesn't really slow things down on travis (4 minutes total). Going to leave it in as this could catch problems with the remote sources.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 scipy>=0.11
 numpy>=1.3
 pandas
+bs4
```
QuLogic: This really should be `beautifulsoup4`. `bs4` is an alias to prevent squatting https://pypi.org/project/bs4/

jGaboardi: Thanks! This is being addressed in #239

Issue #178, opened 2019-08-25
development guidelines link failure


jGaboardi:
Since [switching from `readthedocs` to GH hosted docs](https://github.com/pysal/pysal/pull/1134), the [development guidelines](https://pysal.readthedocs.io/en/latest/developers/index.html) link is broken. 

Should the `development guidelines` link be:

- pointing directly to the [wiki](https://github.com/pysal/pysal/wiki), as it on the [Community Guidelines](http://pysal.org/community) page, or 
- simply to the Community Guidelines page?

jGaboardi:
resolved in #181

PR #181, opened 2019-09-01
DOCS: moving off rtd

sjsrey:


jGaboardi:
#178 resolved by f86bbe5

Issue #183, opened 2019-09-02
Weights constructor behaving unexpectedly with dataframe IDs


darribas:
Attached full (reproducible, zipped) notebook, but the gist of it is:

![Annotation 2019-09-02 162802](https://user-images.githubusercontent.com/417363/64124342-acd9ea00-cd9e-11e9-9eeb-fd0054082fd3.png)

From the docstring:

```
idVariable  : string
                    the name of the column to use as IDs. If nothing is
                    provided, the dataframe index is used
```

Shoudn't `[5]` output be the same as `[4]`?

[weights_ids.zip](https://github.com/pysal/libpysal/files/3566588/weights_ids.zip)

sjsrey:
This is not working as documented.

jGaboardi:
In the case of `[5]` the conditionals in [`line 325`](https://github.com/pysal/libpysal/blob/e6998425b917d342ece7a569abe5df6a15af220a/libpysal/weights/contiguity.py#L325), [`line 335`](https://github.com/pysal/libpysal/blob/e6998425b917d342ece7a569abe5df6a15af220a/libpysal/weights/contiguity.py#L335), [`line 337`](https://github.com/pysal/libpysal/blob/e6998425b917d342ece7a569abe5df6a15af220a/libpysal/weights/contiguity.py#L337) are `False` so the only [`line 339`](https://github.com/pysal/libpysal/blob/e6998425b917d342ece7a569abe5df6a15af220a/libpysal/weights/contiguity.py#L339)(below) in running.

```python
w = cls.from_iterable(df[geom_col].tolist(), ids=ids, id_order=id_order, **kwargs)
```

Should the documentation change or the functionality?

sjsrey:
I'm working on changing the functionality (in libpysal) so we get something like:

![2019-09-02-1642-1567467775-sel](https://user-images.githubusercontent.com/118042/64135760-d431b680-cda0-11e9-910f-2d23387db6bb.png)

jGaboardi:
#184

PR #184, opened 2019-09-03
BUG: Queen and Rook from_dataframe do not match docs

sjsrey:
This address #183

### DIFF
--- a/libpysal/weights/contiguity.py
+++ b/libpysal/weights/contiguity.py
@@ -161,25 +161,20 @@ class Rook(W):
         :class:`libpysal.weights.weights.W`
         :class:`libpysal.weights.contiguity.Rook`
         """
-        if id_order is not None:
-            if id_order is True and ((idVariable is not None) 
-                                     or (ids is not None)):
-                # if idVariable is None, we want ids. Otherwise, we want the
-                # idVariable column
-                id_order = list(df.get(idVariable, ids))
-            else:
-                id_order = df.get(id_order, ids)
-        elif idVariable is not None:
-            ids = df.get(idVariable).tolist()
-        elif isinstance(ids, str):
-            ids = df.get(ids).tolist()
+        if idVariable is not None:
+            ids = df.get(idVariable).tolist() #  idVariable takes precedent 
+            if id_order is None:
+                id_order = ids
+        else:
+            ids = df.index.tolist()
+            id_order = ids
         return cls.from_iterable(df[geom_col].tolist(), ids=ids,
                                  id_order=id_order, **kwargs)
 
 class Queen(W):
     """
     Construct a weights object from a collection of pysal polygons that share at least one vertex.
-
+ 
     Parameters
     ----------
     polygons    : list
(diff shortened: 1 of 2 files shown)


sjsrey:
> Nice. I was wondering about the `Rook.from_dataframe` and `Queen.from_dataframe` not matching.
> 
> As a side note, would it be possible to source the internals for the `Rook` and `Queen` classes from one set of methods? As is stands, aren't the `Rook` and `Queen` class methods identical?

We need to hear from @ljwolf before doing much more as he was the original author and I'm not sure the changes reflect his original intent.

ljwolf:
Doesn't this fail if idVariable is none but ids is passed?

ljwolf:
In general, this works better. The behavior I was going for was (0) use the index if nothing else is provided, (1) use idvariable to build IDs if it was provided, (2) use IDs if provided and, (3) keep the order of the now-resolved IDs if id_order was not provided  

This was awkward because of the id_order behavior and pysal's silent sorting of the input weights if you don't force their order.  Id_order basically always needs to be set to either range(0,n) (to avoid silent lexicographic sorting of the IDs) or the user-specified order that's different from their input dataframe. 

Splitting id_order from the order in which the inputs are provided adds so much complexity... Sorry I missed that twice...

ljwolf:
Here's a bit of a test matrix. 

I still think the main issue is that PySAL separates the ids from how the ids are ordered. We talked about this back in GSOC. It should be the responsibility of stuff *outside* the class to make sure the input data is ordered correctly, because *the worst surprise* is that your weights are silently incorrect w.r.t. the ordering of your input data. By default, PySAL does this because (at the lowest level) weights are sorted lexicographically unless forced to retain the original ordering ([docstring](https://github.com/pysal/libpysal/blob/master/libpysal/weights/weights.py#L34), [location in code](https://github.com/pysal/libpysal/blob/master/libpysal/weights/weights.py#L147))
Sorting weight keys by default will make the weights out of alignment with the input dataframe *unless* the user knows this and presorts their data or happens to have it sorted by default. I raised this back in my GSOC, but we decided to keep it around, rather than remove it. I understand how this arose, but I still think the "right" fix is to just stop sorting things on input. Things become dramatically simpler, then:
1. take the ids from the index of the dataframe. If a user wants to use a variable, make them *set that as the index on the dataframe*)
2. for collections without indices, use the `ids` input in the order it's provided or default to range(0,n). 

Quite a few PRs have attempted to address this inconsistency (#137, [pysal#988](https://github.com/pysal/pysal/pull/988), [pysal#951](https://github.com/pysal/pysal/pull/951) (refresh of [pysal#922](https://github.com/pysal/pysal/pull/922)), [pysal#946](https://github.com/pysal/pysal/pull/946)), and this solution, while a lot better than current, is still inconsistent, just within the Rook class:

```python
import geopandas, numpy
from libpysal import weights, examples

full = geopandas.read_file(examples.get_path('south.shp'))
numpy.random.seed(11121)
part = full.sample(frac=.1)

nothing = weights.Rook.from_dataframe(part)
nothing.id_order == part.index # makes sense

ids_from_seq = weights.Rook.from_iterable(part.geometry, 
                                          ids=part.index.tolist())
print((ids_from_seq.full()[0] == nothing.full()[0]).all())
# False, because of sorting in seq, surprising.

fips_ids_from_df = weights.Rook.from_dataframe(part, 
                                               ids=part.FIPS.tolist())
fips_ids_from_seq = weights.Rook.from_iterable(part.geometry, 
                                               ids=part.FIPS.tolist())
fips_ids_from_seq.id_order # is a list of strings, makes sense
fips_ids_from_df.id_order # is part.index, surprising
(fips_ids_from_seq.full()[0] == fips_ids_from_df.full()[0]).all() 
# False, because of sorting in seq, surprising.


fips_ido_from_df = weights.Rook.from_dataframe(part, 
                                               id_order=part.FIPS.tolist())
fips_from_reindex = weights.Rook.from_dataframe(part.set_index('FIPS'))
fips_idvar_from_df = weights.Rook.from_dataframe(part,
                                                 idVariable='FIPS')
# fails!
#fips_ido_from_seq = weights.Rook.from_iterable(part.geometry, 
#                                               id_order=part.FIPS.tolist())
# fails!
#fips_ido_from_seq = weights.Rook.from_iterable(part.geometry, 
#                                               ids=part.FIPS.tolist(), 
#                                               id_order = part.index.tolist())
# fails! 
#fips_idvar_from_seq = weights.Rook.from_iterable(part.geometry, 
#                                                 idVariable='FIPS')

fips_idvar_from_seq = weights.Rook.from_iterable(part.set_index('FIPS').geometry)
fips_idvar_from_seq.id_order # is a list of integers, makes sense
fips_ido_from_df.id_order # is a list of integers, surprising
fips_idvar_from_df.id_order # integers in the order of the sorted idVariable!

## We never set id_order, so all string-keyed weights would be lexsorted according to docs

(fips_idvar_from_seq.full()[0] == fips_ido_from_df.full()[0]).all() 
# True, because now the order is respected, surprising iff you assume keys are always lexsorted!
(fips_idvar_from_seq.full()[0] == fips_idvar_from_df.full()[0]).all() 
# True, because now the order is respected, surprising iff you assume keys are always lexsorted!
(fips_from_reindex.full()[0] == fips_ido_from_df.full()[0]).all()
# True, because now the order is respected, surprising iff you assume keys are always lexsorted!
```

The issue is the surprise caused by sorting that forces the complexity up. now, you've always gotta set `ids` and `id_order` correctly and propagate them always.

ljwolf:
I still think the "correct" solution here is to stop sorting string indices by default. When provided, take the index in the input order.

Issue #189, opened 2019-10-09
Voronoi results in weights of different shape than input points


dillongardner:
```
from libpysal.weights import Voronoi
import numpy as np

np.random.seed(12345)
points= np.random.random((500,2))*10 + 10
w_test = Voronoi(points)
assert len(w_test.neighbors) == len(points)
```

This fails and I believe it should not. 

- Platform information:
```
posix darwin
posix.uname_result(sysname='Darwin', nodename='xxxx', release='18.6.0', version='Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64', machine='x86_64')
```
- Python version: 
```
3.6.5 (default, Mar 30 2018, 06:41:53)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
```
- SciPy version:
```
1.0.1
```
- NumPy version:
```
1.14.2
```

jGaboardi:
Which version of `libpysal` are you on?

dillongardner:
4.1.1

jGaboardi:
hmmmm. `assert len(w_test.neighbors) == len(points)` passes on:

OS: `Darwin-19.0.0-x86_64-i386-64bit`
Python: `3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 02:16:08) [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]`
Libpysal: `4.1.1`
SciPy: `1.2.1`
NumPy: `1.16.3`

jGaboardi:
>SciPy version:
3.6.5 (default, Mar 30 2018, 06:41:53)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]

Which scipy version are you on?

dillongardner:
1.0.1. Updated initial post.

dillongardner:
It is a correct assumption that these _should_ be the same size always, correct?

sjsrey:
> It is a correct assumption that these _should_ be the same size always, correct?

Yes, that is the correct assumption.

Can you let us know if `len(w_test.neighbors) >  len(points)` ?

jGaboardi:
I do believe that is a correct assumption for a planar Voronoi, but may not be the case for a non-planar Voronoi. @ljwolf, correct me if I am wrong?

[...]

PR #190, opened 2019-10-09
BUG: alpha_shape_auto can fail to contain all points in the set.

sjsrey:
In trying to debug #189 the following will reproduce the issue:
Screen Shot 2019-10-09 at 7 44 33 PM

The issue appears to be that `alpha_shape_auto` can fail to contain all the points in a set as one of the conditions was that the bounding box for the alpha hull and the point set were equal. That was necessary, but not sufficient to ensure all points were contained in the hull.

As a result, when clipping the voronoi regions using the alpha hull, a point (and region) could be omitted and thus the number of points would not match the length of the resulting voronoi W. Notice the point in the north east of this diagram. The light green is the alpha shape clipper, blue are the voronoi regions, and in red are the points:

Screen Shot 2019-10-09 at 7 54 37 PM

The PR has fixes for this.


ljwolf:
This is very weird... I run:
```python
import geopandas
from libpysal.weights import Voronoi
import numpy
from tqdm import tqdm

numpy.random.seed(12345)
for iteration in tqdm(range(1000)):
    points = numpy.random.random((50,2))*10+10
    w_test = Voronoi(points, clip='ahull')
    if len(w_test.neighbors) != len(points):
        print('bad: {}'.format(iteration))
        break
```
and get 1000 good iterations...? That's no proof, but I'd like to play around with an MWE... Can someone please serialize a set of coordinates that show the bug?

libpysal/cg/alpha_shapes.py
```diff
@@ -27,9 +27,19 @@ def partial_inner(func):
...
 EPS = np.finfo(float).eps
+from shapely.geometry.point import Point
 
 __all__ = ['alpha_shape', 'alpha_shape_auto']
 
+def _single_hull(geoms, points):
+    if geoms.shape[0] != 1:
+        return False
+    for point in points:
+        if not (point.within(geoms[0]) or point.touches(geoms[0])):
```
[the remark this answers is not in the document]
sjsrey: Yes, I can  swap intersects in for the conjunction.

And will delete the second triangulation.

sjsrey:
> This is very weird... I run:
> 
> ```python
> import geopandas
> from libpysal.weights import Voronoi
> import numpy
> from tqdm import tqdm
> 
> numpy.random.seed(12345)
> for iteration in tqdm(range(1000)):
>     points = numpy.random.random((50,2))*10+10
>     w_test = Voronoi(points, clip='ahull')
>     if len(w_test.neighbors) != len(points):
>         print('bad: {}'.format(iteration))
>         break
> ```
> 
> and get 1000 good iterations...? Ive tried a few other seeds & adding an extreme point near the boundary of the bounding box like in the image shown & still can't get a failing example. That's no proof, but I'd like to play around with an MWE... Can someone please serialize a set of coordinates that show the bug?
> 
> I don't wanna block... I'd like to make the changes outlined above w/o worrying they break the fix.



> This is very weird... I run:
> 
> ```python
> import geopandas
> from libpysal.weights import Voronoi
> import numpy
> from tqdm import tqdm
> 
> numpy.random.seed(12345)
> for iteration in tqdm(range(1000)):
>     points = numpy.random.random((50,2))*10+10
>     w_test = Voronoi(points, clip='ahull')
>     if len(w_test.neighbors) != len(points):
>         print('bad: {}'.format(iteration))
>         break
> ```
> 
> and get 1000 good iterations...? Ive tried a few other seeds & adding an extreme point near the boundary of the bounding box like in the image shown & still can't get a failing example. That's no proof, but I'd like to play around with an MWE... Can someone please serialize a set of coordinates that show the bug?
> 
> I don't wanna block... I'd like to make the changes outlined above w/o worrying they break the fix.

gist: https://gist.github.com/sjsrey/6d420dd4c59ec86d52ceeb2fb078f8ed

ljwolf:
Thanks!

darribas:
Sorry for the delay, I wanted to make sure I understood the rationale for this. I think I get it now, and the fix looks great. +1 from me.

libpysal/cg/alpha_shapes.py
```diff
@@ -551,10 +561,16 @@ def alpha_shape_auto(xys, step=1, verbose=False):
             print('%.2f%% | Trying a = %f'\
 		  %((i+1)/radii.shape[0], alpha))
         geoms = alpha_geoms(alpha, triangles, radii, xys)
-        if (geoms.shape[0] != 1) or not (np.all(xys_bb == geoms.total_bounds)):
-            break
-        else:
+        if verbose:
+            print(geoms.shape)
```
darribas: Do we want to leave this print statement?

libpysal/cg/alpha_shapes.py
```diff
@@ -551,10 +561,16 @@ def alpha_shape_auto(xys, step=1, verbose=False):
...
+        if _valid_hull(geoms, points):
             geoms_prev = geoms
+        else:
+            if verbose:
+                print(i, geoms.shape[0], xys_bb, geoms.total_bounds)
+            break
+    if verbose:
+        print(geoms_prev.shape)
```
darribas: For the production version, do we want this print statement in? If so, maybe add a message of what it represents?

libpysal/cg/alpha_shapes.py
sjsrey: Good catch. Will remove.

libpysal/cg/alpha_shapes.py
sjsrey: Good catch - removed.

Issue #147, opened 2019-03-19
remove `distribute_setup.py`?


jGaboardi:
[`distribute_setup.py`](https://github.com/pysal/libpysal/blob/master/distribute_setup.py) is no longer used in the `libpysal` [`setup.py`](https://github.com/pysal/libpysal/blob/master/setup.py). Should it be considered for removal?

jGaboardi:
@sjsrey @weikang9009 What do you think about this?

jGaboardi:
closed via #193

PR #193, opened 2019-10-15
[WIP] removing unused relics

jGaboardi:
 - This PR addresses #147 & #194 
   - [`distribute_setup.py`](https://github.com/pysal/libpysal/blob/master/distribute_setup.py) is no longer used in the `libpysal` [`setup.py`](https://github.com/pysal/libpysal/blob/master/setup.py) and seems to be a relic from the early days of the PySAL project.
    - [`README.txt`](https://github.com/pysal/libpysal/blob/master/README.txt) refers to the structure of pre-reorganization of PySAL (`pysal.spatial_dynamics`, `pysal.network`, etc...) and seems not to be used for any purpose.

Although these are not large files, if they are no longer relevant we should remove them to cut down on the clutter.


jGaboardi:
Tying in #194

jGaboardi:
@weikang9009 Thanks! Before we merge I am going to review more stuff in the directory to see f anything else can be removed. At first (second?) glance it seems that [`INSTALL.txt`](https://github.com/pysal/libpysal/blob/master/INSTALL.txt) can also go.

Issue #197, opened 2019-10-17
[BUG] alpha_shapes/shapely import error


jGaboardi:
Although tests passed locally and on Travis CI for #190 a bug somehow snuck through causing Travis failures downstream (i.e. [`spaghetti`](https://travis-ci.com/jGaboardi/spaghetti/jobs/246506682)). The offender in from line 33 in [`cg.alpha_shapes.py`](https://github.com/pysal/libpysal/blob/d0ff1e0b462bc4df109a1530649c9364e9f6c1c7/libpysal/cg/alpha_shapes.py#L33):

```python
from shapely.geometry.point import Point
```

jGaboardi:
Closing because this is not a bug, it is an Allowed Failure.

PR #198, opened 2019-10-17
[BUG] correcting shapely import bug

jGaboardi:
 - Patches the minor bug introduced in #190 
 - Addresses #197 
 - The bug causes [Travis failures in `spaghetti`](https://travis-ci.com/jGaboardi/spaghetti/jobs/246506682), but was not caught when originally merged.

### DIFF
--- a/libpysal/cg/alpha_shapes.py
+++ b/libpysal/cg/alpha_shapes.py
@@ -30,7 +30,7 @@ import scipy.spatial as spat
 from ..common import requires
 
 EPS = np.finfo(float).eps
-from shapely.geometry.point import Point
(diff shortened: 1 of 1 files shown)


jGaboardi:
I was wrong about this. This is not a bug:

```python
>>> from shapely.geometry import Point                                                                                                  
>>> from shapely.geometry.point import Point as pPoint                                                                                 
>>> Point(1,1) == pPoint(1,1)                                                                                                                                                                                                                                             
True
```

The actual cause is an "Allowed Failure", as is expected due to `shapely` not being installed in that Travis build.

Closing this PR and #198.

Issue #199, opened 2019-10-17
rearrange shapely import in cg.alpha_shapes


jGaboardi:
Following my false alarm yesterday (#197, #198), I was wondering if the `shapely` import statement from [`cg.alpha_shapes`](https://github.com/pysal/libpysal/blob/master/libpysal/cg/alpha_shapes.py) could be moved so that it doesn't result in an [allowed failure downstream](https://travis-ci.com/jGaboardi/spaghetti/jobs/246506682).

This can be accomplished by:
1. [removing the initial import statement at the top of the script](https://github.com/pysal/libpysal/blob/c01fce5293629f361b5e6f6616cf63104fd1b1b7/libpysal/cg/alpha_shapes.py#L33)
2. [moving the import statement outside the try block of in `alpha_shapes_auto`](https://github.com/pysal/libpysal/blob/c01fce5293629f361b5e6f6616cf63104fd1b1b7/libpysal/cg/alpha_shapes.py#L536)
3. [adding the `geom.` prefix to `Point` in this list comprehension](https://github.com/pysal/libpysal/blob/c01fce5293629f361b5e6f6616cf63104fd1b1b7/libpysal/cg/alpha_shapes.py#L553)

jGaboardi:
Addressing in #200

PR #200, opened 2019-10-17
fix quasi-redundant import of shapely

jGaboardi:
- Addressing #199 
- ~~Adjusting docstring examples~~
- Removing extra spaces


jGaboardi:
hmmmm. Doctrsing examples from [`alpha_shapes`](https://github.com/pysal/libpysal/blob/c01fce5293629f361b5e6f6616cf63104fd1b1b7/libpysal/cg/alpha_shapes.py#L445-L453) and [`alpha_shapes_auto`](https://github.com/pysal/libpysal/blob/c01fce5293629f361b5e6f6616cf63104fd1b1b7/libpysal/cg/alpha_shapes.py#L519-L524) passed locally with integer coordinates... Will swap back for original example output.

Issue #204, opened 2019-11-01
AttributeError: 'Queen' object has no attribute 'silent_island_warning'


TransUMD:
Hello,

I was using libpysal to conduct esda. I was doing the spatial weights using continuity weights but it pops up the following errors:

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-189-9a78d6d6df16> in <module>
     13             break
     14         c_type = crime_type[i*2+j]
---> 15         li_bg = esda.moran.Moran_Local(blgp_crime[c_type], qW_blgp)
     16         mi_bg = esda.moran.Moran(blgp_crime[c_type], qW_blgp)
     17         sig = 1 * (li_bg.p_sim < 0.05)

E:\Anaconda\envs\incenTrip\lib\site-packages\pysal\explore\esda\moran.py in __init__(self, y, w, transformation, permutations, geoda_quads)
    870         np.seterr(**orig_settings)
    871         self.z = z
--> 872         w.transform = transformation
    873         self.w = w
    874         self.permutations = permutations

E:\Anaconda\envs\incenTrip\lib\site-packages\libpysal\weights\weights.py in set_transform(self, value)
    943                     row_sum = sum(wijs) * 1.0
    944                     if row_sum == 0.0:
--> 945                         if not self.silent_island_warning:
    946                             print(('WARNING: ', i, ' is an island (no neighbors)'))
    947                     weights[i] = [wij / row_sum for wij in wijs]

AttributeError: 'Queen' object has no attribute 'silent_island_warning'

I check the raw code and can't find things about silent_island_warning.

Thank you.

Best,

sjsrey:
Could you include your example code and data so we can debug?

TransUMD:
[SampleCode&Data.zip](https://github.com/pysal/libpysal/files/3801154/SampleCode.Data.zip)

TransUMD:
Hello,

I attached the sample code and data. It is basically a hexagon-based shapefile with the value attached in each hexagon, and I am trying to apply queen weight but it reported the silent_island_warning message. 

The kernel weights works.

jakeandersonbell:
Hi, I have also been experiencing this error.

I think the issue lies within weights.py of libpysal. At line 941-946, where the script checks for items that have no neighbours, it checks the bool of the attribute 'silent_island_warning'. This attribute is not mentioned anywhere else in the module, pehaps the devs were planning on implementing it at some point. 

[...]

PR #206, opened 2019-11-24
(bug) replace silent_island_warning with silence_warnings for weights

weikang9009:
`silent_island_warning` was removed from the attribute list of `libpysal.weights.W` and was replaced with the name `silence_warnings`, but one occurence was not updated to reflect such change, leading to errors reported in https://github.com/pysal/libpysal/issues/204 and https://github.com/pysal/spreg/pull/27

### DIFF
--- a/libpysal/weights/weights.py
+++ b/libpysal/weights/weights.py
@@ -942,7 +942,7 @@ class W(object):
                     wijs = self.weights[i]
                     row_sum = sum(wijs) * 1.0
                     if row_sum == 0.0:
-                        if not self.silent_island_warning:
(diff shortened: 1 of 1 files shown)


jGaboardi:
I thought this was maybe failing due to the new `scipy` version ([`1.3.3`](https://docs.scipy.org/doc/scipy-1.3.3/reference/))? But seems like we are grabbing the [`v1.4.0rc1`](https://travis-ci.org/pysal/libpysal/jobs/616372763#L1866-L1867).

weikang9009:
The issue with scipy seems to occur starting today. I am fine with merging the PR now.

PR #209, opened 2019-12-05
DOC: math rendering in sphinx, and members included for W

sjsrey:
@weikang9009 see the use of the `members` option in the [W rst file](https://github.com/pysal/libpysal/compare/master...sjsrey:math?expand=1#diff-b4c964f775631bd479f5aebd399659b9).

PR #210, opened 2019-12-05
(docs) automatically generate docstrings for class members

weikang9009:
This PR is to update the configuration file for sphinx to:
* automatically generate docstrings for class members including methods, attributes, properties
* avoid display class members twice 

this would be a more convenient fix compared with the solution suggested in https://github.com/pysal/libpysal/pull/209

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -263,10 +263,15 @@ texinfo_documents = [
 
 # Generate the API documentation when building
 autosummary_generate = True
-numpydoc_show_class_members = True
-class_members_toctree = True
-numpydoc_show_inherited_class_members = True
+
+# avoid to show members twice
+numpydoc_show_class_members = False
 numpydoc_use_plots = True
+
+# automatically document class members
+autodoc_default_options = {
+    'members': True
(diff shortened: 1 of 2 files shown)


sjsrey:
> This PR is to update the configuration file for sphinx to:
> 
>     * automatically generate docstrings for class members including methods, attributes, properties
> 
>     * avoid display class members twice
> 
> 
> this would be a more convenient fix compared with the solution suggested in #209

Agreed.

Can you add the suggestion about updating the bootstrap theme  that @knaaptime mentioned in #209 ? We can then close that one and merge this.

weikang9009:
> > This PR is to update the configuration file for sphinx to:
> > ```
> > * automatically generate docstrings for class members including methods, attributes, properties
> > 
> > * avoid display class members twice
> > ```
> > 
> > 
> > this would be a more convenient fix compared with the solution suggested in #209
> 
> Agreed.
> 
> Can you add the suggestion about updating the bootstrap theme that @knaaptime mentioned in #209 ? We can then close that one and merge this.

Just did.

PR #231, opened 2020-01-04
move to splot implementation of gdf plotting

ljwolf:
This moves weights plotting logic out of the custom implementation and to `splot`.


martinfleis:
The method is splot has a slightly different behaviour that the current one in libpysal. I would suggest to update first the splot one to allow similar result as we can currently get here as discussed a bit in https://github.com/pysal/splot/issues/159

jGaboardi:
Should this be closed? – https://github.com/pysal/splot/issues/188

Issue #121, opened 2018-10-30
some weights util functions are lost in __ini__.py


weikang9009:
Some weights util functions are lost in the initialization as well as the online docs, including [`neighbor_equality `](https://github.com/pysal/libpysal/blob/master/libpysal/weights/util.py#L1213) and [`write_gal `](https://github.com/pysal/libpysal/blob/master/libpysal/weights/util.py#L1200)

PR #232, opened 2020-01-04
add neighbor_equality and write_gal to the imports from util

ljwolf:
This addresses bug #121


weikang9009:
Should these also be added to the api docs?

PR #233, opened 2020-01-04
[WIP] stop sorting ids by default

ljwolf:
This is a stub to start looking at stop sorting weights IDs by default. 
I want to see what fails in the CI when we stop doing the sorting.


knaaptime:
resolved by Graph

PR #244, opened 2020-02-09
Fix and simplify filter_adjlist.

QuLogic:
3. [x] This pull introduces new functionality covered by    [docstrings](https://en.wikipedia.org/wiki/Docstring#Python) and [unittests](https://docs.python.org/2/library/unittest.html)? 
4. [N/A] You have [assigned a reviewer](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) and added relevant [labels](https://help.github.com/articles/applying-labels-to-issues-and-pull-requests/) <- not possible
5. [x] The justification for this PR is: The docs say `filter_adjust` should remove elements in insertion order, but it doesn't do that because `set` is unordered. It is dependent on the order Python decides, which does in fact change between versions, and the related test breaks on 3.8.
   The test itself is also wrong because it removed elements in a 'nice' manner, not insertion order.

### DIFF
--- a/libpysal/weights/adjtools.py
+++ b/libpysal/weights/adjtools.py
@@ -180,21 +180,15 @@ def filter_adjlist(adjlist, focal_col = 'focal', neighbor_col = 'neighbor'):
     -------
     an adjacency table with reversible entries removed. 
     """
-    directed = set(map(tuple, adjlist[[focal_col, neighbor_col]].values))
+    edges = adjlist.loc[:, [focal_col, neighbor_col]]
     undirected = set()
-    for tupe in directed:
-        #print(tupe)
-        if (tupe in undirected):
-            to_drop_mask = ((adjlist[focal_col] == tupe[0])
-                            & (adjlist[neighbor_col] == tupe[-1]))
-            to_drop = adjlist[to_drop_mask]
-        elif (tuple(reversed(tupe)) in undirected):
-            to_drop_mask = ((adjlist[focal_col] == tupe[-1])
-                            & (adjlist[neighbor_col] == tupe[0]))
-            to_drop = adjlist[to_drop_mask]
+    to_remove = []
+    for index, *edge in edges.itertuples(name=None):
+        edge = tuple(edge)
+        if edge in undirected or edge[::-1] in undirected:
+            to_remove.append(index)
         else:
-            undirected.update((tupe,))
-            undirected.update((tuple(reversed(tupe),)))
-            continue
-        adjlist = adjlist.drop(to_drop.index)
+            undirected.add(edge)
+            undirected.add(edge[::-1])
+    adjlist = adjlist.drop(to_remove)
     return adjlist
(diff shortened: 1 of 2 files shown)


libpysal/weights/adjtools.py
```diff
@@ -180,21 +180,15 @@ def filter_adjlist(adjlist, focal_col = 'focal', neighbor_col = 'neighbor'):
...
-    directed = set(map(tuple, adjlist[[focal_col, neighbor_col]].values))
+    edges = adjlist.loc[:, [focal_col, neighbor_col]]
     undirected = set()
-    for tupe in directed:
-        #print(tupe)
-        if (tupe in undirected):
-            to_drop_mask = ((adjlist[focal_col] == tupe[0])
-                            & (adjlist[neighbor_col] == tupe[-1]))
-            to_drop = adjlist[to_drop_mask]
-        elif (tuple(reversed(tupe)) in undirected):
```
QuLogic: This condition was redundant with https://github.com/pysal/libpysal/pull/244/files#diff-ab25b3d21ac8613bd5cb0998c1f9f2a7L197 which added the reverse to the set.

libpysal/weights/adjtools.py
```diff
@@ -180,21 +180,15 @@ def filter_adjlist(adjlist, focal_col = 'focal', neighbor_col = 'neighbor'):
...
-                            & (adjlist[neighbor_col] == tupe[0]))
-            to_drop = adjlist[to_drop_mask]
+    to_remove = []
+    for index, *edge in edges.itertuples(name=None):
+        edge = tuple(edge)
+        if edge in undirected or edge[::-1] in undirected:
+            to_remove.append(index)
         else:
-            undirected.update((tupe,))
-            undirected.update((tuple(reversed(tupe),)))
```
QuLogic: But there was a bug here as the comma was within the `tuple` call, so it really added the two elements to the set, _not_ a reversed tuple.

libpysal/weights/tests/test_adjlist.py
```diff
@@ -34,8 +34,8 @@ def test_filter(self):
             badgrid = weights.W.from_adjlist(alist)
             np.testing.assert_allclose(badgrid.sparse.toarray(),
                                        grid.sparse.toarray())
-        assert set(alist.focal.unique().tolist()) == set(list(range(4)))
```
QuLogic: The adjacency list as created in the top of this test is:
```
   focal  neighbor  weight
0      0         2     1.0  *
1      0         1     1.0  *
2      1         0     1.0
3      1         3     1.0  *
4      2         0     1.0
5      2         3     1.0  *
6      3         1     1.0
7      3         2     1.0
```
I have marked the unique edges _in order_ with`*`. As you can see, the documented behaviour could not have produced `{0, 1, 2, 3}` for either `focal` or `neighbor`.

ljwolf: Sure. This test is ok to drop. The main point was to test the edges.

Issue #254, opened 2020-02-18
Error in the internal hack for the Arc_KDTree class inheritance and the KDTree function


joshuawagner93:
The internal hack that is used in libpysal.cg.kdtree / pysal.lib.cg.kdtree has an error and defaults back to using scipy.spatial.KDTree instead of scipy.spatial.cKDTree for scipy version 1.* with *<12.
This is also present in the definition of KDTree for euclidean distance.

#### Currently in pysal.lib.cg.kdtree:

\# internal hack for the Arc_KDTree class inheritance
if int(scipy.version.version.split(".")[1]) < 12:
&nbsp;&nbsp;&nbsp;temp_KDTree = scipy.spatial.KDTree
else:
&nbsp;&nbsp;&nbsp;temp_KDTree = scipy.spatial.cKDTree

#### possible solution:
scipy_version = scipy.version.version.split(".")
\# internal hack for the Arc_KDTree class inheritance
if int(scipy_version[1]) < 12 and int(scipy_version[0]) == 0:
&nbsp;&nbsp;&nbsp;temp_KDTree = scipy.spatial.KDTree
else:
&nbsp;&nbsp;&nbsp;temp_KDTree = scipy.spatial.cKDTree

#### Platform and versions:
- Platform information: nt win32

- Python version: 3.6.7 (default, Dec  6 2019, 07:03:06) [MSC v.1900 64 bit (AMD64)]

- SciPy version: 1.3.1

- NumPy version: 1.17.5

- pysal version: 2.1.0

PR #255, opened 2020-02-18
Bugfix

joshuawagner93:
fixes the bug that libpysal.cg.kdtree uses scipy.spatial.KDTree instead of scipy.spatial.cKDTree for scipy version 1.* with *<12


joshuawagner93:
fixes #254

Issue #263, opened 2020-03-24
examples directory prevents installing with pyInstaller


bbenne10:
- Platform information: Deployment target is Windows. Dev host is macOS and Linux.
- Python version: 3.7.7
- SciPy version: Irrelevant to this report
- NumPy version: Irrelevant to this report

When importing libpysal, the examples are parsed and exported at module import time.
This means that the example files must be present in all deployments of libpysal,
which is problematic for us as we're deploying under pyInstaller on a Windows host. 

Furthermore, the usage of `__file__` means we cannot simply add the examples to our pyinstaller spec, as pyinstaller has an odd split of sourcecode files and data files and so the files will not be where you expect them to be.

My suggestion (and hope) is that you will abandon the implicit importing of examples in the name of namespace unification (or will provide a function that gathers the builtin examples explicitly when called rather than implicitly upon import. Essentially a function that does what lines 62 - 74 of builtin.py do).

OR you might consider simply rewriting the code to be much more defensive and not throwing exceptions when the files are missing. 

I understand that this is a targeted use case, but moving away from implicit filesystem reads (and import time side effects in general) I think warrants opening an issue. Until then, I believe that my organization will need to maintain an in-house forked version of libpysal which removes the examples directory entirely.

I'm happy to open a PR for the behavior above, but would like to talk about the preferred direction rather than guessing.

bbenne10:
Upon further inspection, it looks like simply removing `from . import examples` from `libpysal/__init__.py` might be sufficient. Examples would still work (deferring to the `__init__.py` inside `libpysal.examples`) but would require an extra import e.g. `import libpysal.examples`

sjsrey:
Thanks for raising this. Please open a PR for discussion along the lines you suggest and we will explore ways to handle the side-effects of the recent changes.

Issue #259, opened 2020-03-03
[rough edge] libpysal.examples w/o internet?


jGaboardi:
Currently an error is thrown when trying to import `examples` when no internet connection is available. Most functionality of `libpysal` should persist for users and a warning should be provided that remote datasets cannot be accessed without an internet connection, unless previously downloaded. See traceback below for the error thrown with no internet.

```
../miniconda3/envs/env/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
spaghetti/__init__.py:7: in <module>
    from .network import Network, PointPattern, SimulatedPointPattern
spaghetti/network.py:8: in <module>
    from . import util
spaghetti/util.py:3: in <module>
    from libpysal import cg
../miniconda3/envs/env/lib/python3.7/site-packages/libpysal/__init__.py:30: in <module>
    from . import examples
../miniconda3/envs/env/lib/python3.7/site-packages/libpysal/examples/__init__.py:5: in <module>
    from .remotes import datasets as remote_datasets
../miniconda3/envs/env/lib/python3.7/site-packages/libpysal/examples/remotes.py:14: in <module>
    page = requests.get(url)
../miniconda3/envs/env/lib/python3.7/site-packages/requests/api.py:75: in get
    return request('get', url, params=params, **kwargs)
../miniconda3/envs/env/lib/python3.7/site-packages/requests/api.py:60: in request
    return session.request(method=method, url=url, **kwargs)
../miniconda3/envs/env/lib/python3.7/site-packages/requests/sessions.py:533: in request
    resp = self.send(prep, **send_kwargs)
../miniconda3/envs/env/lib/python3.7/site-packages/requests/sessions.py:646: in send
    r = adapter.send(request, **kwargs)
../miniconda3/envs/env/lib/python3.7/site-packages/requests/adapters.py:516: in send
    raise ConnectionError(e, request=request)
E   requests.exceptions.ConnectionError: HTTPSConnectionPool(host='geodacenter.github.io', port=443): Max retries exceeded with url: /data-and-lab// (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x11c8fd940>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```

ljwolf:
This would still occur if examples are lazy loaded, but it would only raise an issue when importing `pysal.examples`. If we want to deal with this properly, we need to think about & address what happens when:
[...]

PR #264, opened 2020-03-24
GH-263: Don't implicitly import examples when importing base library

bbenne10:
As #263 states:
> When importing libpysal, the examples are parsed and exported at module import time.
> This means that the example files must be present in all deployments of libpysal,
> which is problematic for us as we're deploying under pyInstaller on a Windows host.
> 
> Furthermore, the usage of __file__ means we cannot simply add the examples to our pyinstaller > spec, as pyinstaller has an odd split of sourcecode files and data files and so the files will not be > where you expect them to be.

This addresses the above by simply making the import of `libpysal.examples` explicit. This passes all tests (I'm erroring on two locally due to missing geopandas, but I expect them to pass on Travis)

There are other ways of resolving this, but this is the simplest and will get discussion started.

### DIFF
--- a/libpysal/examples/base.py
+++ b/libpysal/examples/base.py
@@ -45,17 +45,20 @@ def get_list_of_files(dir_name):
     create a list of file and sub directories in dir_name
     """
     #names in the given directory
-    file_list = os.listdir(dir_name)
     all_files = list()
-    # Iterate over all the entries
-    for entry in file_list:
-        # Create full path
-        full_path = os.path.join(dir_name, entry)
-        # If entry is a directory then get the list of files in this directory
-        if os.path.isdir(full_path):
-            all_files = all_files + get_list_of_files(full_path)
-        else:
-            all_files.append(full_path)
+    try:
+        file_list = os.listdir(dir_name)
+        # Iterate over all the entries
+        for entry in file_list:
+            # Create full path
+            full_path = os.path.join(dir_name, entry)
+            # If entry is a directory then get the list of files in this directory
+            if os.path.isdir(full_path):
+                all_files = all_files + get_list_of_files(full_path)
+            else:
+                all_files.append(full_path)
+    except FileNotFoundError:
+        pass
 
     return all_files
 


bbenne10:
I don't seem to have permissions to assign someone to the task or to assign appropriate labels. I'll investigate the failure and see if there's something I can do to resolve it.

bbenne10:
I have to admit that I don't understand why conda info would fail here. Can someone point me in the right direction>

knaaptime:
the conda recipe does an [explicit check](https://github.com/conda-forge/libpysal-feedstock/blob/416b1894fc958288ae3a9ffd175cba50ff2e3444/recipe/meta.yaml#L35) to ensure the submodues can be imported, so if we adopt this change, we'll just need to update the recipe

jGaboardi:
@bbenne10 The most important for now is passing Travis and Appveyor. We are recently trying out GitHub Actions and it still looks like we need a better understanding.

sjsrey:
> As #263 states:
> 
> > When importing libpysal, the examples are parsed and exported at module import time.
> > This means that the example files must be present in all deployments of libpysal,
> > which is problematic for us as we're deploying under pyInstaller on a Windows host.
> > Furthermore, the usage of **file** means we cannot simply add the examples to our pyinstaller > spec, as pyinstaller has an odd split of sourcecode files and data files and so the files will not be > where you expect them to be.
> 
> This addresses the above by simply making the import of `libpysal.examples` explicit. This passes all tests (I'm erroring on two locally due to missing geopandas, but I expect them to pass on Travis)
> 
> There are other ways of resolving this, but this is the simplest and will get discussion started.

The change to handle this particular use case is straightforward, and has the added benefit of increasing the cross-platform profile of libpysal.

The change would, however, be an api breakage as a lot of downstream code relies on the current implicit import scheme. I'm wondering if some of the other ways of resolving this that you hinted at above might offer a fix for this use case without the api break?

bbenne10:
I apologize for what will be a slight delay here. I'm not the one facing the failure (our build guy reported the problem and I'm unable to properly recreate on macOS or linux). I'm waiting on a traceback so that I might change the examples import behavior to fail softly and warn the user rather than throw an exception at the problematic point. 

I'll revert the current commit and get that in sometime today. Thanks for the feedback so far. I'm sure we'll come to a good compromise here.

bbenne10:
Okay - new approach. 
Inside `examples.base`, `get_list_of_files` (which is really a re-implementation of `os.walk` as far as I could tell) assumes (in master now) that the directory name it was passed exists before it attempts to walk it.

I could think of two solutions:
1) **add a call to `os.isdir(dir_name)`.**
This didn't seem particularly Pythonic ("Easier to ask forgiveness than permission")
2) **wrap the whole body in a `try/except` and log in the failure case.**
This was easier and more Pythonic, but I didn't see a `logging.logger` anywhere (just a few calls to `print` inside the `requires` decorator) so I just silently returned an empty list for directories that don't exist. 

This seemed adequate: It doesn't change the effective API (it doesn't change anything at all during normal setup.py backed installations) and it makes a simple pyinstaller deployment work properly. However, it doesn't fix every case...

If a pyinstaller user wants to make use of the examples directory in their deployment, a new (slightly more invasive) solution needs to be invoked. In `examples.builtin`, `__file__` is used to find the directory containing the current file - this doesn't work with pyinstaller. Instead something like the below is necessary:

```python
# instead of line 62, we do:
import sys
if getattr(sys, 'frozen', False):
    builtin_root = os.path.join(sys._MEIPASS, "examples")  # I think. This is untested.
else:
    builtin_root = os.path.dirname(__file__)    
```

I can ALSO add a commit that does this and test it, but that seemed like overkill for our specific scenario, but may be useful in the generic sense.

bbenne10:
If desired, I am also happy to add a specific test for this case once I re-familiarize myself with nose. Just let me know.

bbenne10:
ping?
Failure on conda 3.6 looks to be caused by missing geopandas in the installed environment. Nothing I can do to fix that. 

Does this approach look reasonable? What about the addon approach I mentioned?

sjsrey:
Will be on this tomorrow and get back to you after reviewing.

jGaboardi:
@bbenne10 The failure of the Python 3.6 GitHub Action is a non-issue. I have had success the last couple days on a "complete solution" CI workflow with GitHub Actions in [`spaghetti`](https://github.com/pysal/spaghetti/blob/master/.github/workflows/unittests.yml) and will do a PR here when I have time. The most important part is @sjsrey 's review and recommendation about the API breakage, etc.

sjsrey:
> Okay - new approach.
> Inside `examples.base`, `get_list_of_files` (which is really a re-implementation of `os.walk` as far as I could tell) assumes (in master now) that the directory name it was passed exists before it attempts to walk it.
> 
> I could think of two solutions:
> 
>     1. **add a call to `os.isdir(dir_name)`.**
>        This didn't seem particularly Pythonic ("Easier to ask forgiveness than permission")
> 
>     2. **wrap the whole body in a `try/except` and log in the failure case.**
>        This was easier and more Pythonic, but I didn't see a `logging.logger` anywhere (just a few calls to `print` inside the `requires` decorator) so I just silently returned an empty list for directories that don't exist.
> 
> 
> This seemed adequate: It doesn't change the effective API (it doesn't change anything at all during normal setup.py backed installations) and it makes a simple pyinstaller deployment work properly. However, it doesn't fix every case...
> 
> If a pyinstaller user wants to make use of the examples directory in their deployment, a new (slightly more invasive) solution needs to be invoked. In `examples.builtin`, `__file__` is used to find the directory containing the current file - this doesn't work with pyinstaller. Instead something like the below is necessary:
> 
> ```python
> # instead of line 62, we do:
> import sys
> if getattr(sys, 'frozen', False):
>     builtin_root = os.path.join(sys._MEIPASS, "examples")  # I think. This is untested.
> else:
>     builtin_root = os.path.dirname(__file__)    
> ```
> 
> I can ALSO add a commit that does this and test it, but that seemed like overkill for our specific scenario, but may be useful in the generic sense.

This would work against portability (MEIPASS) which we want to avoid if at all possible.

bbenne10:
@sjsrey: not entirely true. Without the earlier `getattr(sys, "frozen", False) `, it would. But the check against `sys.frozen` should prevent any problems, as that is only inserted by pyinstaller when running under a frozen/deployed instance. 

Additionally, I could rework to instead use `getattr(sys, "_MEIPASS", False)` for free which would avoid ever using `sys.frozen` and have no effect on portability.

sjsrey:
@bbenne10 another alternative is a lazy [load of examples](https://github.com/pysal/libpysal/compare/master...sjsrey:noexamples). 

Does this cover the issues your are facing?

ljwolf:
Looking over the current solution, it bails whenever *any* example is not found where we expect it to be found, pyinstaller or no. This is my preferred solution. I do not think lazy loading is worth it here, and I don't think it makes sense to assume that all frozen instances are compliant with pyInstaller's behavior ehre. @bbenne10 why specifically is the try/except wrap here not sufficient for your usecase? 

Second, on the off-chance the simple things have been exhausted, can this be solved on the installation end using existing environment variable checks in `base.py` ([l36](https://github.com/pysal/libpysal/blob/master/libpysal/examples/base.py#L36)) to set the data location?

bbenne10:
> @bbenne10 why specifically is the try/except wrap here not sufficient for your usecase?

The try/catch is perfectly sufficient for our use case.
We're using my fork of this for a deployment very shortly.

However, it fails late and doesn't allow the deployment of examples.
(to be clear: We're not using any of the examples, so our use case is quite simple compared to what is possible and available.)
I was making the point above so that we can support general use with pyinstaller rather than just our limited use-case. 

I'm also happy to work the `sys._MEIPASS` (read: pyinstaller specific) code into a commit if you'd like without generalizing to the `sys.frozen` attribute. 
This allows simple deployment with pyinstaller and doesn't make a generalized decision about frozen installs. 
If anyone wants that - please request it.
I think it is the right solution, but can understand not wanting more invasive changes than necessary.

> Second, on the off-chance the simple things have been exhausted, can this be solved on the installation end using existing environment variable checks in base.py (l36) to set the data location?

This *may* be feasible in a general sense, but as we are not deploying the examples at all, we have not investigated this approach. 
Further, I'm unsure as to whether you can predict where the data files will be deployed before the pyinstaller unpack step (they're unpacked into a temp directory). 
This may be a solution for the additional commit I was referring to above (just set `PYSALDATA`  to `${sys._MEIPASS}/examples` or something before importing libpysal) if you want the examples), but I haven't explored this as we'd still need something to handle the `FileNotFoundError` in the case that there's simply nothing deployed.

bbenne10:
> @bbenne10 another alternative is a lazy [load of examples](https://github.com/pysal/libpysal/compare/master...sjsrey:noexamples).
> > Does this cover the issues your are facing?

This actually looks pretty elegant and simple - It fixes the problem we have and generalizes well to any other import problems (maybe allowing use the geopandas dependent parts too if implemented thoughtfully). I'd happily accept this compromise, but it sounds like others may have problems with the approach.

ljwolf:
> The try/catch is perfectly sufficient for our use case.

Great. I'd prefer to ship that, then, if it doesn't break downstream stuff. @sjsrey, we talked a little on Friday about this... but I can't recall if this current solution breaks downstream packages? I do not see how this does, unless the `libpysal` installation is itself broken. 

> This allows simple deployment with pyinstaller and doesn't make a generalized decision about frozen installs. If anyone wants that - please request it.

That's very considerate to offer. Thanks! I think that would indeed be useful in abstract. 

But @sjsrey @jGaboardi, I don't see the use of example datasets in pyinstaller setups as a generally useful feature. PyInstaller is usually for standalone applications; I would suggest that standalone applications probably would focus on using fresh (not canned) data. We'd have to maintain this specific instance, which I don't personally want to do since that would entail checking that a pyinstaller deployment correctly detects/runs examples code. 

> lazy loading

If this addresses the issue & does not break downstream, then it's fine to merge if you're cool to maintain it @sjsrey! I don't understand its benefit of this over the try/catch specifically for solving the source of the error: example datasets may be stripped or relocated in some deployments. Lazy loading examples still retains the issue that, if imported, datasets may break/be missing. You're just changing when the error is encountered.

jGaboardi:
This may also be related to #259 ?

bbenne10:
> This may also be related to #259 ?

Without beginning a :+1: farm, this issue is VERY related to #259

bbenne10:
Okay. Upon even further review, I think that there's nothing more to do than the try/catch I have committed. pyinstaller has a notion of `hooks` that are used to modify third-party libraries (there's a pretty extensive scipy hook to ship all of its static files, for instance). I believe that if anyone wants to use the examples, they should not modify libpysal but rather write a hook for pyinstaller to ship the examples directory. This shouldn't require more changes in libpysal (but if they do, they'll be quite simple) and allows libpysal to remain ignorant of pyinstaller's internals. 

However, I think the try/catch is still meaningful: it is more defensive generally and allows those of us without the need for the examples at all to avoid shipping them. 

If that is an acceptable compromise, I believe that the work done here is complete. Do you all have any thoughts?

sjsrey:
I agree that the try/catch is the way to go on this. Thanks everyone for the discussion to hone this.

PR #266, opened 2020-03-25
TEST: turning off 3.6 on github actions

sjsrey:
3.6 job grabs an older version of geopandas which is causing the voronoi tests to fail (those pass on 3.6 on travis).


sjsrey:
will be handled in #267

PR #273, opened 2020-03-31
Mac builds seem to take longer — bump up timeout

jGaboardi:
Bump up timeout due to Mac builds seems to take longer. This seems to have solved the issue with `spaghetti` unittests, currently [all 36 are passing](https://github.com/pysal/spaghetti/runs/548801586).


sjsrey:
Looks good

PR #269, opened 2020-03-28
Polish up GitHub Action residuals

jGaboardi:
* See #268 for list up items
 * See #267 for original PR

PR #274, opened 2020-04-01
Handle connection errors for remote datasets

sjsrey:
Addressing #259

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -27,13 +27,17 @@
             environment-file: ${{ matrix.environment-file }}
             activate-environment: test
        - shell: bash -l {0}
-         run: |
-           conda info
-           conda list
-           conda config --show-sources
-           conda config --show
-           python -c 'import libpysal; libpysal.examples.fetch_all()'
-           py.test -v libpysal --cov=libpysal --cov-report=xml
+         run: conda info --all
+       - shell: bash -l {0}
+         run: conda list
+       - shell: bash -l {0}
+         run: conda config --show-sources
+       - shell: bash -l {0}
+         run: conda config --show
+       - shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


.github/workflows/unittests.yml
```diff
@@ -27,13 +27,13 @@
...
             activate-environment: test
        - shell: bash -l {0}
-         run: conda info --all
-       - shell: bash -l {0}
-         run: conda list
-       - shell: bash -l {0}
-         run: python -c 'import libpysal; libpysal.examples.fetch_all()'
-       - shell: bash -l {0}
-         run: py.test -v libpysal --cov=libpysal --cov-report=xml
+         run: |
```
jGaboardi, line 30: Is there a benefit for including these commands in one `shell` bash command? I had split these into separate bash commands for log clarity, which ultimately helped with the debugging of the `geopandas` import issue (#269). I would recommend we keep the current split commands and add in the show commands. Thoughts?
```
- shell: bash -l {0}
 run: conda info --all
- shell: bash -l {0}
 run: conda list
- shell: bash -l {0}
 conda config --show-sources
- shell: bash -l {0}
 conda config --show
- shell: bash -l {0}
 run: python -c 'import libpysal; libpysal.examples.fetch_all()'
- shell: bash -l {0}
 run: py.test -v libpysal --cov=libpysal --cov-report=xml
```

sjsrey:
I thought keeping them together might have been a way to fix the missing gp issue. So that didn't work and we can go back to separating them.

jGaboardi:
> I thought keeping them together might have been a way to fix the missing gp issue. So that didn't work and we can go back to separating them.

Let's keep them separate. Having them separate is better for debugging (so that each command is visible in the actions log) and does not affect the activated Python environment.

jGaboardi:
I'll just go ahead make changes to the `.yml` in this PR or do you want to?

PR #278, opened 2020-04-06
Return alpha option & use pygeos for alphashaping if available

ljwolf:
This provides options to return the alpha radius from `alpha_shape_auto`, as well as the detected bounding disks.

```python
import libpysal, geopandas, matplotlib.pyplot as plt, numpy
numpy.random.seed(2478879)
testpoints = numpy.random.normal(size=(100,2))
testalpha, testradius, testx = libpysal.cg.alpha_shapes.alpha_shape_auto(testpoints, return_circles=True)
ax = geopandas.GeoDataFrame(geometry=[testalpha]).plot()
ax.scatter(*numpy.row_stack(testalpha.boundary.coords).T, marker='.', color='k')
[ax.add_patch(plt.Circle(c, radius=testradius, edgecolor='r', facecolor='none')) for c in testx]
plt.scatter(*numpy.row_stack(testx).T, color='r', marker='x')
```

ashape_bounding

It still needs
- [X] confirmation that it "works" for n < 4 (*edit: now has decent behavior in these cases. When n=3, returns the alpha shape (a triangle) that triangle's radius, and the circles defined by each pair of boundary points. When n=2, it returns the empty hull, a radius of half the distance between the points, and the circle with those points as diameters. When n=1, it returns the point, zero, and the point.*)
- [x] tests

I keep with the rest of the package using `radius` to refer to the radius of the circles, `alpha` to refer to `1/radius`, and `circle` (not `disk`) to refer to the shapes defining the `alpha_shape`.  Strictly speaking, Edselbrunner et al use `alpha` to mean `radius` itself, since when alpha = infty, the alpha shape is the convex hull (the space bounded by circles with infinite radius). Two points are "alpha-neighbors" when they admit a circle of radius alpha with no other points. etc. But, since we're consistent, I think this is fine. 

In theory, I suppose this should be exposed for the individual `alpha_shape` method, but since `alpha` is specified in that case, it's not very hard to do:

```python
ahull = alpha_shape(points, alpha)
circles = construct_bounding_circles(ahull, 1/alpha)
```


libpysal/cg/alpha_shapes.py
```diff
@@ -564,46 +576,128 @@ def alpha_shape_auto(xys, step=1, verbose=False):
...
+            return alpha_shape
+        elif xys.shape[0] == 2:
+            if return_radius:
+                r = spat.distance.euclidean(xys[0], xys[1])/2
+                if return_circles:
+                    circle = _construct_centers(xys[0], xys[1], r)
+                    return [alpha_shape, r, circle]
+                return [alpha_shape, r]
+            return alpha_shape
+        elif return_radius:
```
jGaboardi, line 605: Should this be an `if`? Is is part of the `xys.shape[...]` logic?

libpysal/cg/alpha_shapes.py
```diff
@@ -564,46 +576,128 @@ def alpha_shape_auto(xys, step=1, verbose=False):
...
     if verbose:
         print(geoms_prev.shape)
-    return geoms_prev[0] # Return a shapely polygon
+    if return_radius:
+        out = [geoms_prev[0], radi_prev]
+        if return_circles:
+            out.append(construct_bounding_circles(out[0], radi_prev))
+        return out
+    return geoms_prev[0]
+    # Return a shapely polygon
```
jGaboardi, line 647: Small nitpick: Let's move the comment above the `return` statement.

libpysal/cg/alpha_shapes.py
```diff
@@ -564,46 +576,128 @@ def alpha_shape_auto(xys, step=1, verbose=False):
...
-        print('Step set to %i'%step)
+        print('Step set to %i' % step)
     for i in range(0, len(radii), step):
         radi = radii[i]
         alpha = (1 / radi) - EPS
         if verbose:
-            print('%.2f%% | Trying a = %f'\
-            %((i+1)/radii.shape[0], alpha))
+            print('%.2f%% | Trying a = %f'
+                  % ((i+1)/radii.shape[0], alpha))
```
jGaboardi, line 632: Was `black` run on this line?

libpysal/cg/alpha_shapes.py
```diff
@@ -564,46 +576,128 @@ def alpha_shape_auto(xys, step=1, verbose=False):
...
+    if return_radius:
+        out = [geoms_prev[0], radi_prev]
+        if return_circles:
+            out.append(construct_bounding_circles(out[0], radi_prev))
+        return out
+    return geoms_prev[0]
+    # Return a shapely polygon
+
+
+def construct_bounding_circles(alpha_shape, radius):
```
jGaboardi, line 650: How does this affect the `numba` performance? Or is that even an issue?

libpysal/cg/alpha_shapes.py
ljwolf: No, it's part of the shape logic. Since we know we have `n < 4` and have exhausted `n==2` and `n==1`, this is the bit for `n==3`. We can deal with that implicitly. Because of that, we can just use this `elif` to deal with `n==3 & return_radius`.

ljwolf: A better solution might be to split off the standard behavior to a separate function, like `_alpha_shape_search` which returns alpha shape and radius, so that we can collapse this branch & the branching after the search into a single branching.

libpysal/cg/alpha_shapes.py
ljwolf: yeah, this is from black

libpysal/cg/alpha_shapes.py
ljwolf: done!

libpysal/cg/alpha_shapes.py
jGaboardi: Good explanation. Maybe needs a short inline comment?

jGaboardi: > A better solution might be to split off the standard behavior to a separate function, like _alpha_shape_search which returns alpha shape and radius, so that we can collapse this branch & the branching after the search into a single branching.

I am in favor of this idea.

libpysal/cg/alpha_shapes.py
jGaboardi: Huh. Maybe it displays as split line formatting for the comparison only.

libpysal/cg/alpha_shapes.py
jGaboardi: Maybe the `diff` isn't updating; the comment (line 647) is still below the `return` statement (line 646)

libpysal/cg/alpha_shapes.py
ljwolf: This could be numba'd, but it's not important imho. The hull is usually orders of magnitude fewer points. 

I've forced it into `jit(nopython=True)` mode with some simple unpacking and rearranging.

jGaboardi: To be honest, when doing to performance testing on `alpha_shapes` with and without `numba`, it did not see a significant difference in computation time for any sized problems.

libpysal/cg/alpha_shapes.py
ljwolf: Actually... looking at it, since the method to compute the radius varies for each n (1,2,3,4+) we can't really get any savings from doing that. I think it's ok as is.

libpysal/cg/alpha_shapes.py
ljwolf: really? that's surprising... our earlier performance checks suggested so.

jGaboardi: Then I probably did something wrong.

libpysal/cg/alpha_shapes.py
jGaboardi: How about the following inline comment:
```python
# implicitly deal with n==3 scenarios
```

darribas:
This looks good to me too. My only question would be: have you done any check on whether performance when `return_disks=False` deteriorates? I'm OK with minor things but if it's a notable one, it might be worth rethinking how we implement this. Could you provide a couple of timing experiments?

ljwolf:
> have you done any check on whether performance with `return_disks=False` deteriorates? 

Sure, I can run timing experiments now. But, for what it's worth I think it's fine for two reasons. 
1. For any hull, the number of points on the boundary is usually quite small compared to the number of points. The alpha shape is no different. This code only computes the limiting circles when the hull is completed, so we're only using the boundary points to build the limiting circles. 
2. On @jGaboardi's suggestion, the code to build the limiting circles now uses `jit(nopython=True)` and has no for loops nor branching internally, so it should be quite fast.

> rethinking how we implement this

There isn't really another way to implement this from an algorithmic perspective. 

Solely for the API/UX, we could only support building the circles in a second call using the constructor I outlined in the first post, `construct_bounding_circles(alpha_shape, radius)`. But, if we did, we would still need to have `return_radius` in `alpha_shape_auto`. I don't think forcing two calls here is a good design decision from a UX perspective... `numpy` has many `return_xxx` options for functions to do similar things as this. Further, circle construction is not on by default, so there is no performance penalty there, either.

ljwolf:
whoops did not mean to close!

ljwolf:
# performance

## size
For point 1, the average number of points on the alpha-hull of a pointcloud with 1 million points (normally-distrbuted, though) is around 70. 
alpha_shape_boundary_size

Weirder shapes will see this flex a little, but it should remain approximately logarithmic. So, we're talking **very** few points on the boundary compared to the number of triangles constructed elsewhere. Run time will always be dominated by constructing the Delaunay Triangulation. 

## time

|  n points | time with circles       | time without circles | difference                            | significant? | 
| -----------  | -------------------------- | --------------------------| ------------------------------------| ----------------| 
| 100         | 33.3 ms +/- 2.37 μs | 33.2 ms +/- 31 μs   | +.01 ms, or .00001 s           | :ballot_box_with_check:  |
| 1000       | 437 ms +/- 30.7 ms | 440ms +/- 30.7 ms |  - 3 ms, or -.003 s (n.s.) | :x: |
| 10000    |  8.22s +/- 555 ms    | 7.81 s +/- 104 ms    | +.41 s | :x: | 
| 100000 | 
| 1000000| 

**lprun at n=10000 goes here**

ljwolf:
For comparison, the timing table with pygeos:

|  n points | time with circles       | time without circles | diff. w/ circles | diff. with pygeos |
| -----------  | -------------------------- | --------------------------| ------------------|------------------| 
| 100         | 26.6 ms +/- 1.68 ms | 25.7 ms +/- 1.14 ms   | +.9 ms (+3%)  | -8 ms (-22%) | 
| 1000       | 278 ms +/- 9.62 ms | 270 ms +/- 12.5 ms |  +8 ms (+9%) | -159 ms (-36%) |
| 10000    |  3.67 s +/- 163 ms    |  3.53s +/- 32.7 ms    | +.14 s (+3%) |  -4.55 s (-55%) | 
| 100000 |  29.7 s +/- 731 ms | 30.2 s +/- 994 ms | -.5 s (-1%) |  -75.3 s (-71.71%)|

darribas:
This is amazing!!! Super good report too! I think since it does not seem to affect much, and it's a nice feature, we can pull it in. The `pygeos` speed up is awesome!

Very minor thing for consideration: maybe better to add a gpkg file for the tests, instead of a shapefile?

ljwolf:
I've migrated the data in the alpha shape tests to use geopackages.

PR #280, opened 2020-04-09
ENH: allow specific buffer in fuzzy_contiguity

martinfleis:
Allow specific distance for buffer in `fuzzy_contiguity` instead of a fraction of bbox length. There are cases in which it is better to specify tolerance/buffer as a set known distance instead of a fraction of the df extent. 

I have also change buffering to vectorised operation, so it should be significantly faster, esp. once pygeos will land in GeoPandas.

With this change it could be directly used in https://github.com/pysal/mapclassify/pull/61.


sjsrey:
Thanks!

PR #285, opened 2020-04-29
remove non-coincident points correctly from knn weights

ljwolf:
This fixes self neighbor issues by moving away from for loops in de-duplicating the neighbor lists.


ljwolf:
code has also been blackened. We may want to consider running black all over the library so that substantive commits can be sussed out from ones that are not.

libpysal/weights/distance.py
```diff
@@ -4,22 +4,27 @@
...
+)
 import copy
 from warnings import warn as Warn
 from scipy.spatial import distance_matrix
 import scipy.sparse as sp
 import numpy as np
 
 
-def knnW(data, k=2, p=2, ids=None, radius=None, distance_metric='euclidean'):
+def knnW(data, k=2, p=2, ids=None, radius=None, distance_metric="euclidean"):
```
ljwolf, line 21: I would like to full delete this function next release.

jGaboardi:
@ljwolf want me put in PR for review to blacken the whole library?

ljwolf:
for next release, yes I definitely would

ljwolf:
Hey folks, great vote of confidence in my abilities here, but... the fix in this PR is (a) ignorant of the `ids` passed and (b) demonstrably broken when the number of duplicates gets large:

```python
import libpysal, numpy, scipy.spatial as spatial, uuid

coordinates = numpy.random.random(size=(20,2))
coordinates = numpy.row_stack([coordinates]*5)
ids = [uuid.uuid4() for _ in range(100)]
kdt = spatial.KDTree(coordinates)
distances, indices = kdt.query(kdt.data, 2, p=2)
full_indices = numpy.arange(kdt.n)
not_self_mask = indices != full_indices.reshape(-1,1)
not_self_indices = indices[not_self_mask].reshape(kdt.n, -1) # breaks! 
```

My test case was a situation where coincident points were rare.

ljwolf:
This now addresses the k+1 overdraw as well as names all in numpy.

ljwolf:
I need to rebase this and it needs another round of review!

jGaboardi (approved):

ljwolf:
An example workflow that fails immediately at the `k=2` option in the old version, but passes now:

```python
from libpysal import examples, weights
import geopandas, numpy

numpy.random.seed(2478879)

# read in the baltimore dataset (n=211)
baltim = geopandas.read_file(examples.get_path('baltim.shp'))
# bootstrap it so that we have many coincident points (on average, 10 per observation)
baltim_random = baltim.sample(n=2000, replace=True).reset_index()

# can we pick 2 neighbors?
w_dedupe2 = weights.KNN.from_dataframe(baltim_random, k=2)
# can we pick 10 neighbors?
w_dedupe10 = weights.KNN.from_dataframe(baltim_random, k=10)
# can we pick 50 neighbors?
w_dedupe50 = weights.KNN.from_dataframe(baltim_random, k=50)
```

The warning about disconnected components should be sufficient (imho) to tell users that there are coincident points in their data, but I've also included extra in the docstrings!

PR #287, opened 2020-05-03
[WIP] Handle coincident points in KNN weights

sjsrey:
This is aimed at https://github.com/pysal/pysal/pull/1060.

If the approach makes sense, it can be extended to Kernel weights and DistanceBand weights.  The latter will need slightly different handling for symmetry.

### DIFF
--- a/libpysal/weights/distance.py
+++ b/libpysal/weights/distance.py
@@ -61,16 +61,6 @@ def duplicated(array):
 
     return duplicate
 
-def _check_duplicates(data):
-    if PANDAS:
-        df = pandas.DataFrame(data)
-        return df.duplicated()
-    else:
-        print('pandas is required for _check_duplicates')
-        return None
-
-
-
 def knnW(data, k=2, p=2, ids=None, radius=None, distance_metric='euclidean'):
     """
     This is deprecated. Use the pysal.weights.KNN class instead. 
@@ -109,6 +99,22 @@ class KNN(W):
                 instance
                 Weights object with binary weights
 
+    See Also
+    --------
+    :class:`libpysal.weights.weights.W`
+
+    Notes
+    -----
+
+    Ties between neighbors of equal distance are arbitrarily broken.
+
+    In the case of coincident points, the first record in a set of duplicates
+    (i.e., points with same coordinates) is defined as the coincident seed and
+    the remaining points in the set are coincident duplicates. Initial neighbors are identified using the the set of unique+coincident seed points (i.e., the
+    coincident duplicates are not included initially). Then, each coincident
+    duplicate has its neighbors set equal to that of its coincident seed.
+
+
     Examples
     --------
     >>> import libpysal
@@ -133,15 +139,6 @@ class KNN(W):
     {1: 1.0, 4: 1.0}
     >>> 0 in wnn2.neighbors
     False
-
-    Notes
-    -----
-
-    Ties between neighbors of equal distance are arbitrarily broken.
-
-    See Also
-    --------
-    :class:`libpysal.weights.weights.W`
     """
     def __init__(self, data, k=2, p=2, ids=None, radius=None,
                  distance_metric='euclidean', **kwargs):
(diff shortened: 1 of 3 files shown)


ljwolf:
I would super prefer #285... is there something wrong with #285?

ljwolf:
OK, to be clear, this does *more* than simply ensure that co-incident points are assigned valid (co-incident) neighbors, it forces the neighbors of co-incident points to be outside of the coincident set, correct?

libpysal/weights/distance.py
```diff
@@ -94,21 +170,46 @@ def __init__(self, data, k=2, p=2, ids=None, radius=None,
         else:
             self.kdtree = KDTree(data, radius=radius, distance_metric=distance_metric)
             self.data = self.kdtree.data
+
+        duplicates = duplicated(self.data)
+        coincident = duplicates[:,1].any()
+
+        self.duplicates = duplicates
```
ljwolf, line 177: Does this mean that all distance-style weights objects gain a duplicates attribute?

ljwolf:
This method either ignores or fails on ids.
```python
import libpysal, numpy, uuid

coordinates = numpy.random.random(size=(5,2))
coordinates = numpy.row_stack([coordinates]*5)

ids = [uuid.uuid4().hex for _ in range(100)] 

#fails, since we pass ids as id_order to resolve sorting issues from the dataframe
w = libpysal.weights.KNN(coordinates, id_order=ids) 
# ignores the ids because of libpysal#284
w = libpysal.weights.KNN(coordinates, ids=ids) 

```

libpysal/weights/distance.py
```diff
@@ -94,21 +170,46 @@ def __init__(self, data, k=2, p=2, ids=None, radius=None,
...
+            for i, row in enumerate(to_weight):
+                row = row.tolist()
+                row.remove(i)
+                row = [unique_ids[j] for j in row]
+                focal = unique_ids[i]
+                neighbors[focal] = row
+            for row in duplicate_ids[0]:
+                neighbors[row] = neighbors[duplicates[row, 2]]
+            n = self.data.shape[0]
+            ids = list(range(n))
```
ljwolf, line 204: This forces ids to be `0,n-1`. The other branch does not do this.

ljwolf:
I wanted to add a `remove_coincident=True` argument that allows users to keep the coincident points but safely remove the self-neighbor. But, I can't really figure out how to get the ids/`id_order` stuff fixed without adding further changes.

pedrovma:
Just so I can understand, what happens now if you have different apartment units in a building (same coordinates)? Would they be considered neighbors of each other or will they be assigned neighbors in other buildings?

sjsrey:
> I would super prefer #285... is there something wrong with #285?
> 
> _EDIT: Yes, I (a) forgot to deal with ids and (b) tested on a scenario where th number of coincident points was usually quite small relative to `k`, so you always got the self-index in the output._

My bad was I didn't see #285 until after I did this.

libpysal/weights/distance.py
sjsrey: The thinking was to let the user know, somehow, that they have coincident points - duplicates is a bad choice in this regard as that might be taken to mean the records are identical in all attributes, whereas I think by coincident implies spatial duplicates.

sjsrey:
> OK, to be clear, this does _more_ than simply ensure that co-incident points are assigned valid (co-incident) neighbors, it forces the neighbors of co-incident points to be outside of the coincident set, correct?

Yes, that was the intention.

sjsrey:
> Just so I can understand, what happens now if you have different apartment units in a building (same coordinates)? Would they be considered neighbors of each other or will they be assigned neighbors in other buildings?

This would have them assigned neighbors with non-zero distances.

There are two (at least) cases where this happens - the one you point out where the issue is the data lacks sufficient spatial information to spatially disambiguate units in the same apartment building, and the other is with repeat sales data. In the latter, the data may have a temporal attribute that can be used to differentiate the records.

ljwolf:
> the data may have a temporal attribute that can be used to differentiate the records

Ah, brilliant! never even thought that you could just send `(n,3)` arrays into the constructor! 

That'd be very strongly sensitive to scaling, right? For data measured in UTM at a monthly timeframe, your inter-temporal neighbors are much "closer" than your spatial neighbors, right? UTM coordinates are typically in thousands of meters, whereas months will be 0...11.

ljwolf:
I think I'd want to see
- [ ] an option to force neighbors back to being drawn from the coincident set
- [ ] ids to be handled
- [ ] use the numpy-oriented fix in #285 instead of the for looping. 

I also don't see how this needs to be ported to kernel/distance band weights, save for estimating the bandwidth of a kernel?

sjsrey:
> > the data may have a temporal attribute that can be used to differentiate the records
> 
> Ah, brilliant! never even thought that you could just send `(n,3)` arrays into the constructor!
> 
> That'd be very strongly sensitive to scaling, right? For data measured in UTM at a monthly timeframe, your inter-temporal neighbors are much "closer" than your spatial neighbors, right? UTM coordinates are typically in thousands of meters, whereas months will be 0...11.

Just to be clear, I wasn't thinking that they would be passing in (n,3) as the data for the knn search, with one of the three being the temporal coordinate. That would raise all kinds of scaling issues that you point out.

The motivation for the approach in the PR was that coincident spatial points violate the law (I think it was Keith Clarke's) that no two events can occupy the same point at the same time. In data sets where at temporal attribute is missing, the coincident points imply this happens. 

Maybe the suggested approach should be an option, rather than the default. Another option would be to jigger the coordinates of the coincident points prior to the search - but that also comes with some possible side-effects.

ljwolf:
> Maybe the suggested approach should be an option, rather than the default. 
Yes, I think that's appropriate. I was trying to add a `remove_coincident=False` above and was running into the `ids`/`id_order` issues.
> Another option would be to jigger the coordinates of the coincident points prior to the search
Sure, that'd be reasonable. I think the solution implemented in #285 though is just fine, and better than reducing the precision of the data through jittering.

ljwolf:
I'd like to merge #285 to fix the problem, and then return to this as a "ignore_coincident" option for the constructor...

knaaptime:
i think this is resovled by graph and can be closed

PR #291, opened 2020-05-30
[WIP] DEV: blacken codebase

sjsrey:


requirements_dev.txt
```diff
@@ -1,6 +1,7 @@
+coverage
```
jGaboardi, line 1: Since we have migrated to `pytest`/`codecov` we should update this by removing `nose` and `coveralls` related packages and adding in `pytest`, `pytest-cov`, and `codecov`.

jGaboardi:
@slumnitz noticed that I [got a bit carried away in the spacing](https://github.com/pysal/pysal/issues/1174#issuecomment-640026700) for the standard doctoring example. I am going to update [the example](https://github.com/pysal/pysal/issues/1174#issuecomment-640026700) now.

jGaboardi:
@sjsrey Shall I work on resolving these conflicts then merge? After this we can focus on #319 and #342

jGaboardi:
Windows failures for Python 3.6 only due to the current issue upstream with [`fiona`](https://github.com/pysal/libpysal/pull/291/checks?check_run_id=1305557717#step:9:2726). Going ahead and merging, as this PR is independent of that issue.

Issue #284, opened 2020-04-29
Ids option is ignored by W.__init__


ljwolf:
You can send whatever you like to `W(neighbors, ids='garbage')`, because it's never used. This is the genesis of many bugs in the past & present. 

```python
W({1:[2,3], 2:[3], 3:[2]}, ids="If you're reading this, something's up!")

```

It looks like it's never actually been plugged in anywhere. Back in [3ef349](https://github.com/pysal/libpysal/commit/3ef349d6241b5909fa80fe4026a7cd57140c0bb5), it was added, but never used anywhere afaict? 

This really needs to be cleaned up once and for all, I think.

MgeeeeK:
Will `ids` become a property/attribute of `W` object in future? 
I've not seen many snippets where it's used, maybe only in `lat2W` ([line](https://github.com/pysal/libpysal/blob/bb30361a78324c1628b2a908272deeb15d516a98/libpysal/weights/util.py#L211))

PR #299, opened 2020-06-29
[WIP] remove Ids attribute

sjsrey:
This is addressing #284 and would remove the ids attribute.

Remaining to be done:
- [ ]  3 locally failing tests in test_contiguity.py
- [ ]  testing downstream

### DIFF
--- a/libpysal/weights/contiguity.py
+++ b/libpysal/weights/contiguity.py
@@ -11,15 +11,16 @@ from .weights import WSP, W
 try:
     from shapely.geometry import Point as shapely_point
     from ..cg.shapes import Point as pysal_point
+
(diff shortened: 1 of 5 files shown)


sjsrey:
The remaining tests that are failing involve the [idVariable](https://github.com/pysal/libpysal/pull/299/checks?check_run_id=817185936#step:9:527).

In exploring those tests, I think they are covering behaviour that is not longer supported when `ids` are removed. At the same time, we might also want to remove  `idVariable` support as it was the combination of `ids` and `idVariable` that have causes a lot of complications.

knaaptime:
resolved by Graph

PR #318, opened 2020-07-13
[WIP] : Base Raster Interface

MgeeeeK:
### Raster Interface

This pull request adds the base raster interface with finalized API. Newly added transformation methods are stored in a new file raster.py. Overview of finalized API for the interface is listed below:
- Using direct functions e.g. `da2W`, `da2WSP`, `w2da`, `wsp2da`
- Using class methods e.g. `Rook.from_xarray`...

**The interface can only work with 2D`(height, width)` and 3D`(layer, height, width)` DataArray.**

Short outline of the interface:

```python
# Creating test DataArray
>>> da = testDataArray(shape=(10, 56, 49), time=False, rand=False, missing_vals=True)
>>> da.dims
('band', 'y', 'x')
# In this case our dataarray is layered using multiple bands
# Method 1
>>> wsp = da2WSP(da, criterion="queen", layer=5, dims=None)
# Method 2
>>> wsp = Queen.from_xarray(da, layer=5, dims=None, sparse=True)
# Accessing data from da using index attribute
>>> data = da.to_series()[wsp.index]
>>> data.head()
band  y     x   
5     71.6  92.4    146
            92.5    121
            92.7    249
            93.0     12
            93.3    210
dtype: int64
# Converting back to the DataArray
>>> da_back = wsp2da(data, wsp)
<xarray.DataArray (band: 1, y: 56, x: 49)>
-1 211 -1 -1 -1 41 37 -1 62 -1 ... -1 171 118 135 28 151 -1 -1 -1 146
Coordinates:
  * band     (band) int64 5
  * y        (y) float64 66.1 66.2 66.3 66.4 66.5 ... 71.2 71.3 71.4 71.5 71.6
  * x        (x) float64 92.1 92.2 92.3 92.4 92.5 ... 96.5 96.6 96.7 96.8 96.9
Attributes:
    nodatavals:  (-1,)
```
_You can checkout example [notebook](https://github.com/MgeeeeK/libpysal/blob/rast_interface/notebooks/Raster%20Interface.ipynb) for more in-depth overview of the interface_

This is work in progress and mainly focuses on introducing the API. Any suggestions/feedback are most welcome. Once the `W` attributes are finalized, I'll:
- [x] Add tests
- [x] Add example on the docstring


MgeeeeK:
@darribas @slumnitz

libpysal/weights/distance.py
```diff
@@ -260,6 +264,39 @@ def from_dataframe(cls, df, geom_col='geometry', ids=None, *args, **kwargs):
...
+        """
+        Construct a weights object from a xarray dataarray.
+
+        Parameters
+        ----------
+        da         : xarray.DataArray
+                    raster file accessed using xarray.open_rasterio method
+        band       : int
+                    select band for raster with multiple bands
+        **kwargs   : keyword arguments
```
slumnitz, line 278: `*args` description missing

libpysal/weights/contiguity.py
```diff
@@ -183,6 +185,31 @@ def from_dataframe(cls,
...
                                  **kwargs)
 
+    @classmethod
+    def from_xarray(cls, da, band=None, sparse=False, **kwargs):
+        """
+        Construct a W object from a xarray.DataArray object.
+
+        Parameters
+        ----------
+        da         : xarray.DataArray
```
slumnitz, line 195: Apply Numpy-docs convention:

param[one space]:[one space]dtype
[four spaces]description

libpysal/weights/contiguity.py
```diff
@@ -183,6 +185,31 @@ def from_dataframe(cls,
...
+                    raster file accessed using xarray.open_rasterio method
+        band       : int
```
slumnitz, line 197: Default is None with automatically takes first band. Maybe worth being explicit and writing 1?

libpysal/weights/contiguity.py
```diff
@@ -183,6 +185,31 @@ def from_dataframe(cls,
...
+                    select band for raster with multiple bands
+        sparse     : boolean
+                    type of weight object. Default is dense. For sparse, sparse = True
```
slumnitz, line 200: why is this a boolean? could be potentially make this a str with sparse or dense? What do you think? :)

libpysal/weights/contiguity.py
```diff
@@ -183,6 +185,31 @@ def from_dataframe(cls,
...
+        **kwargs   : keyword arguments
+                    optional arguments for :class:`pysal.weights.W`
```
slumnitz, line 202: I'd recommend to add that this is only working for dense weights objects, or alternatively make it work for sparse objects too?

libpysal/weights/contiguity.py
```diff
@@ -348,6 +376,31 @@ def from_dataframe(cls, df, geom_col='geometry', **kwargs):
                               **kwargs)
         return w
 
+    @classmethod
+    def from_xarray(cls, da, band=None, sparse=False, **kwargs):
+        """
+        Construct a W object from a xarray.DataArray object.
```
slumnitz, line 382: Same comments for documentation and API as above

libpysal/weights/distance.py
```diff
@@ -260,6 +264,39 @@ def from_dataframe(cls, df, geom_col='geometry', ids=None, *args, **kwargs):
             ids = df[ids].tolist()
         return cls(pts, *args, ids=ids, **kwargs)
 
+    @classmethod
+    def from_xarray(cls, da, band=None, *args, **kwargs):
+        """
+        Construct a weights object from a xarray dataarray.
+
+        Parameters
```
slumnitz, line 272: Same formatting comments for docs throughout the document.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    Create a W object from rasters(xarray.DataArray)
+
+    Parameters
+    ----------
+    da         : xarray.DataArray
+                 raster file accessed using xarray.open_rasterio method
+    criterion  : {"rook", "queen"}
+                 type of contiguity. Default is rook.
+    band       : int
+                 select band for da with multiple bands. Default is 1
```
slumnitz, line 28: **kwargs description missing

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    attrs = da.attrs
+    attrs["dims"] = da.dims
+    attrs["shape"] = da.shape
+    sw.attrs = attrs
+    return sw
+
+
+def w2da(data, w, coords=None, attrs=None):
+    """
+    converts calculated results to a DataArray
```
slumnitz, line 92: specify what `calculated results` are in description

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Arguments
+    ---------
+    data    :   array/list
+                data values stored in 1d array or list
```
slumnitz, line 97: Maybe rephrase like original data values stored in 1d array used to calculate xyz? I think I would not be sure on first glance what data is meant here.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    w       :   W
```
slumnitz, line 98: libpysal.weights.W

libpysal/weights/raster.py
slumnitz: or similar to your W description, data aligned with weights?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+                Spatial weights object aligned with data
+    coords  :   Dictionary/xarray.core.coordinates.DataArrayCoordinates
+                coordinates corresponding to DataArray
```
slumnitz, line 101: does DataArray point toward data? I'd recommend to use a consistent naming scheme for all variables, so that the user can connect them in your description

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    attrs   :   Dictionary
+                Attributes stored in dict related to DataArray
```
slumnitz, line 103: space here

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    Returns
+    -------
+
```
slumnitz, line 106: no space

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    if not isinstance(da, DataArray):
+        raise TypeError("`da` must be an instance of xarray.DataArray")
+    if da.ndim != 2:
+        raise ValueError("`raster` must be 2D")
+    if not (issubclass(da.values.dtype.type, np.integer) or
+            issubclass(da.values.dtype.type, np.float)):
+        raise ValueError("`da` must be an array of integers or float")
+    if band is None:
+        if da.sizes['band'] != 1:
+            warn('Multiple bands detected in da. Using band 1 as default')
```
slumnitz, line 157: Ok I see why None is used, ignore my previous comment :)

libpysal/weights/weights.py
```diff
@@ -329,6 +329,53 @@ def to_networkx(self):
...
+        converts calculated results to ``xarray.DataArray`` object
+
+        Arguments
+        ---------
+        data    :   array
+                    data values stored in 1d array
+        coords  :   Dictionary/xarray.core.coordinates.DataArrayCoordinates
+                    coordinates corresponding to DataArray
+        attrs   :   Dictionary
+                    Attributes stored in dict related to DataArray
```
slumnitz, line 343: Is there a way to call this dict with Xarray? if yes maybe add the code in our description?

libpysal/weights/contiguity.py
MgeeeeK: i followed layout of other methods for sparse argument, they all use sparse as boolean dtype.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+    -------
+    w    : libpysal.weights.W
+           instance of spatial weights class W
+    """
+    band = da_checker(da, band)
+    da = da[band-1:band]
+    if criterion != 'rook':
+        rook = False
+    else:
+        rook = True
```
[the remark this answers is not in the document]
MgeeeeK: i've corrected the default argument now, idk why i missed that :?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,159 @@
...
+__all__ = ['da_checker', 'da2W', 'da2WSP', 'w2da']
+
+
+def da2W(da, criterion, band=None, **kwargs):
+    """
+    Create a W object from rasters(xarray.DataArray)
+
+    Parameters
+    ----------
+    da         : xarray.DataArray
```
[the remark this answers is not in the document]
MgeeeeK: thnx, i was looking at other methods and then decided to follow this syntax. I've corrected all the docstrings now

libpysal/weights/raster.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    Creates DataArray object from passed data
+
+    Arguments
+    ---------
+    data : array/list
+       numpy 1d array or list with dimensionality conforming to w
+    w : libpysal.weights.W
+       Spatial weights object aligned with passed data
+    attrs : Dictionary
```
darribas, line 88: We need to think whether this is strictly required or not. My sense is that it's not so we could do `attrs=None` and, in that case, build a `DataArray` from `w.index` by:

- Selecting the units `u` by looking at the minimum difference between two consecutive dimension labels
- Selecting the width `w` by picking the longest row and using its length
- Selecting the height `h` by picking the longest column and using its length
- Building the full array by creating a mesh of `w` by `h` expressed in `u`

This should be a separate method outside `w2da` and it's called when the user passess `attrs=None`.

Also, this method could probably benefit from `numba` for heavy loops.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,147 @@
...
+        Attributes stored in dict related to DataArray, e.g. da.attrs
+    coords : Dictionary/xarray.core.coordinates.DataArrayCoordinates
```
darribas, line 90: That's great as optional. If not passed, we build them based on what we derive from the `attrs` comment above

libpysal/weights/weights.py
```diff
@@ -1353,7 +1399,7 @@ class WSP(object):
 
     """
 
-    def __init__(self, sparse, id_order=None):
+    def __init__(self, sparse, id_order=None, index=None):
```
MgeeeeK, line 1402: i've temporarily added index parameter here,

libpysal/weights/raster.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def w2da(data, w, attrs, coords=None):
+    """
+    Creates DataArray object from passed data
+
+    Arguments
+    ---------
+    data : array/list
+       numpy 1d array or list with dimensionality conforming to w
```
darribas, line 85: I'd permit this to be a `pandas.Series` too, ideally indexed on `w` and we could check for alignment if it's a series.

libpysal/weights/weights.py
```diff
@@ -8,6 +8,7 @@
 import math
 import warnings
 import numpy as np
+import pandas as pd
```
darribas, line 11: This will eventually be like this but, for now, we might want to import only for the `xarray`-related parts of the module

libpysal/weights/weights.py
```diff
@@ -329,6 +330,51 @@ def to_networkx(self):
...
+            _, y = transform * (np.zeros(ny) + 0.5, np.arange(ny) + 0.5)
+            coords["band"] = np.ones(1)
+            coords["y"] = y
+            coords["x"] = x
+        og_index = pd.MultiIndex.from_product([i for i in coords.values()], names=dims)
+        ser = pd.Series(attrs["nodatavals"][0], index=og_index)
+        ser[self.index] = data
+        data = ser.to_numpy().reshape(shape)
+        da = DataArray(data, coords=coords, dims=dims, attrs=attrs)
+        return da
```
darribas, line 376: This seems repetitive code from `w2da`? I'd not have heavy-lifting code in `weights.py` only dispatchers to `raster.py`

libpysal/weights/weights.py
```diff
@@ -329,6 +330,51 @@ def to_networkx(self):
...
+        -------
+        da : xarray.DataArray
+            instance of xarray.DataArray
+        """
+        try:
+            from xarray import DataArray
+            from affine import Affine
+        except ImportError:
+            raise ImportError(
+                "xarray and affine must be installed to use this functionality")
```
darribas, line 356: Very cool!

libpysal/weights/weights.py
```diff
@@ -1353,7 +1399,7 @@ class WSP(object):
 
     """
 
-    def __init__(self, sparse, id_order=None):
+    def __init__(self, sparse, id_order=None, index=None):
         if not scipy.sparse.issparse(sparse):
```
darribas, line 1403: I'd `import pandas as pd` here probably, at least for now.

libpysal/weights/distance.py
```diff
@@ -260,6 +264,35 @@ def from_dataframe(cls, df, geom_col='geometry', ids=None, *args, **kwargs):
...
+            select band for raster with multiple bands
+
+        See Also
+        --------
+        :class:`libpysal.weights.weights.W`   
+        """
+        band = da_checker(da, band)
+        da = da[band-1:band]
+        ser = da.to_series()
+        id_order = np.where(ser != da.nodatavals[0])[0]
```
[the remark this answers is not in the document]
MgeeeeK: this is for indexing the array (to remove the x-y coords of missing values)

libpysal/weights/raster.py
MgeeeeK: sure! will test and correct the description

libpysal/weights/weights.py
MgeeeeK: my bad, will do as you suggested below

libpysal/weights/distance.py
```diff
@@ -743,16 +779,16 @@ class DistanceBand(W):
     """
 
     def __init__(self, data, threshold, p=2, alpha=-1.0, binary=True, ids=None,
```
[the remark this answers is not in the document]
MgeeeeK: okay i'll add a separate issue for distance band, will also give a heads up on 293

libpysal/weights/raster.py
```diff
@@ -0,0 +1,147 @@
...
+    -------
+    wsp : libpysal.weights.WSP
+        instance of spatial weights class WSP
+    """
+    band = da_checker(da, band)
+    da = da[band-1:band]
+    sw = lat2SW(*da[0].shape, criterion)
+    ser = da.to_series()
+    nodata = np.where(ser == da.nodatavals[0])[0]
+    mask = np.ones((sw.shape[0],), dtype=np.bool)
```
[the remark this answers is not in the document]
MgeeeeK: right now this method is using boolean masking/indexing to remove rows and cols related to missing from csr matrix, i'm working on revamping this method by building the sparse matrix from scratch from only non-missing values.

libpysal/weights/distance.py
```diff
@@ -260,6 +264,35 @@ def from_dataframe(cls, df, geom_col='geometry', ids=None, *args, **kwargs):
...
+            raster file accessed using xarray.open_rasterio method
+        band : int
+            select band for raster with multiple bands
+
+        See Also
+        --------
+        :class:`libpysal.weights.weights.W`   
+        """
+        band = da_checker(da, band)
+        da = da[band-1:band]
```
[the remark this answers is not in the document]
MgeeeeK: if user passes non-numeric value? i've already specified on docstring that the band should be an integer

libpysal/weights/distance.py
```diff
@@ -260,6 +264,35 @@ def from_dataframe(cls, df, geom_col='geometry', ids=None, *args, **kwargs):
...
+        ser = da.to_series()
+        id_order = np.where(ser != da.nodatavals[0])[0]
+        array = np.dstack(
+            np.meshgrid(da.y, da.x, indexing='ij')).reshape(-1, 2)[id_order]
```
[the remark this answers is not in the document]
MgeeeeK: ohk, right now the interface assumes that user will use `xarray.open_rasterio` method to access raster data and convert it to `DataArray`.
`open_rasterio` method will add fixed dimensions (`band`, `x` and `y`) to the `DataArray`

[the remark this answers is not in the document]
MgeeeeK: this [link](https://github.com/pydata/xarray/blob/1be777fe725a85b8cc0f65a2bc41f4bc2ba18043/xarray/backends/rasterio_.py#L348) points to the specific code of the `open_rasterio` i'm talked about.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,147 @@
...
+    band : int
+        select band for da with multiple bands. Default is 1
+
+    Returns
+    -------
+    wsp : libpysal.weights.WSP
+        instance of spatial weights class WSP
+    """
+    band = da_checker(da, band)
+    da = da[band-1:band]
```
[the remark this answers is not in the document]
MgeeeeK: value of the `band` argument will always be an integer

libpysal/weights/raster.py
MgeeeeK: only problem with this method was that it will trim down the rows and columns with missing values, e.g. lets say the initial 500x3 part out of 500x500 raster contains missing values, the resultant `DataArray` will be of shape 500x497 which wont contain the initial 500x3 part (though this is not usefull).
thats why i was storing the original shape of the `DataArray`.

MgeeeeK: Also, if this is acceptable then we can use `from_series` method inside xarray which can build the missing coordinates using tensor product of the available coordinates and it will fill the missing data with `np.nan`

libpysal/weights/distance.py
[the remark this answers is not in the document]
darribas: What is the rationale for forcing it to be numeric? If possible, I'd generalise it to any type of value. I can imagine users having `("red", "green", "blue")` for example or other strings and it'd be very inconvenient to have to rename.

I might be missing something obvious so feel free to push back if there's a better reason :-)

libpysal/weights/distance.py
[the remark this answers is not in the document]
darribas: I see your point now. My sense is that it'd be preferable not to assume this and stick to the more general structure of a standard `DataArray`. Two reasons:

1. This widens the set of use cases your code will be useful for (e.g. anything coming from a NetCDF such as temperature, pollution, etc.)
1. `open_rasterio` _might_ change that convention in the future and then we'd have to change this to align it further

A more general solution would get around these two cases and would not affect the functionality already provided.

libpysal/weights/distance.py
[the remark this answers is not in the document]
darribas: When that is open, please link it here so we can trace back its origin.

libpysal/weights/raster.py
darribas: Yep, that's correct. If the original array has missing values that are fully "trimmable" (as you describe in your 500x500 --> 500x497), the method will not be able to recover. I think this is fine for a "best-guess" approach _as long as_ we document exactly what it does. If the user has the case above, they'll just need to pass their own `attrs` to specify it and that's fine. But my sense is that the use case where the array is not trimmable (ie. missing values do not crop full rows or columns) is relevant enough that a convenience method would be very handy.

I'm not aware of the `from_series` method, but it sounds exactly right so I'd be all in to use it in this case :-)

libpysal/weights/weights.py
darribas: No worries. When `index` is officially built-in on any `W`, `pandas` will be much more widespread on `libpysal`. It's just until then, it might be good to keep its use confined to where we strictly need it :-)

libpysal/weights/raster.py
MgeeeeK: No worries, I had to dig into `from_series` method and documentation for `pandas.MultiIndex`... now the method can completely build `DataArray` without altering original shape and requiring other information

libpysal/weights/distance.py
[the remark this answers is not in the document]
MgeeeeK: This is similar to the case you pointed below, since `xarray.open_rasterio()` store the bands as int that is why I had to force it to be an integer. 

Will push an update which will make the band argument accept strings/floats/ints as long as the passed band belongs to the `DataArray`. I can add a method to check if the passed band is valid or not inside `da_checker`.

libpysal/weights/distance.py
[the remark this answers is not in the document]
MgeeeeK: Sure, will have to think about more generalized solution... mostly I've seen that the norm is to use either `lon`/`lat` or `x`/`y` to represent cell's position in terms of coordinates.

Here KNN is a special case where I needed these dimensions to build the KDTree, this will be also used in `Distance band` functionality for raster.

libpysal/weights/distance.py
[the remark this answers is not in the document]
MgeeeeK: [Distance Band functionality for Raster Interface #328](https://github.com/pysal/libpysal/issues/328)

libpysal/weights/raster.py
```diff
@@ -0,0 +1,148 @@
...
+        dims = tuple(key for key in coords.keys())
+        data_complete = np.array(data).reshape(shape)
+    da = DataArray(data_complete, coords=coords, dims=dims, attrs=attrs)
+    return da
+
+
+def da_checker(da, band):
+    """
+    xarray dataarray checker
+    Parameters
```
slumnitz, line 125: double check formatting here :)

libpysal/weights/distance.py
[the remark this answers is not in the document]
slumnitz: Hi all, to chip in, I think as @MgeeeeK explained, loading a tiff via `rasterio` into an `xarray.DataArray` there is no choice in naming the band `("red", "green", "blue")`, `rasterio` requires the [Band to be an int](https://rasterio.readthedocs.io/en/latest/api/rasterio.html#rasterio.band). At least that is how I understand it. 

There is a [`band_description()`](https://rasterio.readthedocs.io/en/latest/api/rasterio.io.html#rasterio.io.BufferedDatasetWriter.set_band_description) that could potentially be used to load bands via str, though I do not know if this is accessible in an `xarray.DataArray`. Maybe one could use this or a similar definition, if accessible in xarray (?), to print out what  band 1 represents, additionally to that band 1 is taken?

I would recommend against hard coding band1, 2, 3 in "red", "green", "blue", as many different satellites have different band numbers and values for their spectral bands. I think this could lead to a lot of confusion and errors on the users side. Users should be aware of which band they feed in and what this band represents in their data.

Though maybe someone will have added a new dimension that is not called band, and that is not an integer but something else? See additional comments in next thread. :)

libpysal/weights/distance.py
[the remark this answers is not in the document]
slumnitz: I see @darribas point now, currently we are requiring that our `xarray.DataArray()` has `band, x, y` as a dimension? Maybe we could add an option for the user to choose the dimensions used? I.e. if someone is not using airborne or satellite imagery rasters, but interpolated temperate measurements, that do not have a `band` dimension but a `time` dimension? Like sensible defaults and, i.e. a dict parameter that defines the names for customization?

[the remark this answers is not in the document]
MgeeeeK: I gave it a thought... mainly following names are being used `band, x, y, lon, lat, time` to define the dimensions in the netCDF dataset, I can modify `da_checker` to look for these specific dimensions and then proceed with the weight creation and if these dims does not exist then it'll show an error. I'll also add an optional argument for users to pass custom dimensions if the default ones does not exist?

So the modified function will look like this:
```python

KNN.from_xarray(da, criterion="rook", band=None, time=None, x=None, y=None, *args, **kwargs):
    band, time = da_checker(da, band, time, x, y)

```
Through `x` and `y` users can pass the custom names of the dimensions if they are different from default (`x/y` or `lon/lat`).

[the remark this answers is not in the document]
slumnitz: yes, I think it's a good idea to check for defaults and let the user alter these if necessary. Though, I don't like the looks of 4 extra parameters (band, time, x, y) that basically just clarify naming. Maybe there is another way of letting the user choose these, i.e. through a dict like structure dims={"band":None, "time":time, "lat":x, "Lon":y} or something along these lines? Maybe we can discuss different solutions in the call later?

libpysal/weights/tests/test_raster.py
```diff
@@ -0,0 +1,86 @@
...
+from .. import raster
+from ..util import lat2SW
+import unittest
+import numpy as np
+import pandas as pd
+from xarray import DataArray
+
+
+class Testraster(unittest.TestCase):
+    def setUp(self):
```
MgeeeeK, line 11: I've added tests for `raster.py`, do you guys approve of this? I don't wanted to pollute `contiguity`, `distance` and `weights` by hardcoding the `DataArray` but anyways i can add tests to those as well if necessary.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+       e.g. dims = {"lat": "latitude", "lon": "longitude", "layer": "year"}
+    **kwargs : keyword arguments
+        Optional arguments for :class:`libpysal.weights.W`
+
+    Returns
+    -------
+    w : libpysal.weights.W
+       instance of spatial weights class W
+
+    Examples
```
slumnitz, line 40: Very cool!! @darribas are outputs rendered in libpysal documentation, I think we might need to remove everything without a ```>>>```?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Parameters
+    ---------
+    data : array/list/pd.Series
+        1d array-like data with dimensionality conforming to w
+    w : libpysal.weights.W
+        Spatial weights object aligned with passed data
+    attrs : Dictionary
+        Attributes stored in dict related to DataArray, e.g. da.attrs
+    coords : Dictionary/xarray.core.coordinates.DataArrayCoordinates
```
slumnitz, line 138: I'd recommend adding a `Default=` for all arguments looking like `param=`
What is the default of `attrs` just an empty dict?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+        Coordinates corresponding to DataArray, e.g. da.coords
+
+    Returns
+    -------
+    da : xarray.DataArray
+        instance of xarray.DataArray
```
slumnitz, line 144: I would also recommend adding an Example here.

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+    if hasattr(w, 'index'):
+        da = _index2da(data, w.index, attrs, coords)
+    else:
+        raise AttributeError("Cannot convert deprecated W with no index attribute")
+    return da
+
+
+def wsp2da(data, wsp, attrs={}, coords=None):
+    """
+    Creates DataArray object from passed data aligned with WSP object.
```
slumnitz, line 157: Great description! :)

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Parameters
+    ---------
+    data : array/list/pd.Series
+        1d array-like data with dimensionality conforming to w
```
slumnitz, line 162: to wsp?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+    if hasattr(wsp, 'index'):
+        da = _index2da(data, wsp.index, attrs, coords)
+    else:
+        raise AttributeError("Cannot convert deprecated wsp object with no index attribute")
+    return da
+
+
+def testDataArray(shape=(3, 4, 4), time=False, rand=False, missing_vals=True):
+    """
+    Creates 3D test DataArray object
```
slumnitz, line 186: I'd add `xarray.DataArray` object everywhere. PySAL users have not seen xarray used in PySAL before and might not know at first glance which DataArray object we are talking about.

Very cool though! :)

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Parameters
+    ---------
+    shape : tuple
+        Tuple containing shape of the DataArray aligned with
+        following dimension= (layer, y, x)
```
slumnitz, line 192: I think we called it `shape=(layer, height, width)/(layer,lat,lon)` before? let's try to keep it consistent.

libpysal/weights/contiguity.py
```diff
@@ -183,6 +185,42 @@ def from_dataframe(cls,
                                  id_order=id_order,
                                  **kwargs)
 
+    @classmethod
+    def from_xarray(cls, da, layer=None, dims=None, sparse=False, **kwargs):
+        """
+        Construct a weights object from a xarray DataArray.
+
+        Parameters
+            ----------
```
slumnitz, line 194: whoops :)

libpysal/weights/tests/test_raster.py
```diff
@@ -0,0 +1,73 @@
...
+        self.da1 = raster.testDataArray()
+        self.da2 = raster.testDataArray((1, 4, 4), missing_vals=False)
+        self.data1 = pd.Series(np.ones(10))
+
+    def test_da2W(self):
+        w1 = raster.da2W(self.da1, "queen")
```
slumnitz, line 17: I'd suggest to add a couple lines that test if dims and layer parameters are working as expected. What do you think @darribas?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+def da2W(da, criterion="queen", layer=None, dims=None, **kwargs):
+    """
+    Create a W object from xarray.DataArray
+
+    Parameters
+    ----------
+    da : xarray.DataArray
+       Input 2D or 3D DataArray with shape=(layer, height, width)
```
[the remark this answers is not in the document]
MgeeeeK: > Is this really only working for 2D or 3D data? 

Yes this method will only work with 2D and 3D DataArray (1 layer dimension and 2 coordinates dimension), last time dani said that this will satisfy majority of cases... I can look into this if it's required

Sure! i'll change dimensions names... i forgot to change it there

libpysal/weights/raster.py
MgeeeeK: No problem, I'll mention the defaults in the docstrings.

 > What is the default of attrs just an empty dict?
yes

libpysal/weights/raster.py
MgeeeeK: sure! will do

libpysal/weights/tests/test_raster.py
MgeeeeK: ohk

libpysal/weights/contiguity.py
```diff
@@ -348,6 +388,43 @@ def from_dataframe(cls, df, geom_col='geometry', **kwargs):
...
+        """
+        Construct a weights object from a xarray.DataArray.
+
+        Parameters
+        ----------
+        da : xarray.DataArray
+            Input 2D or 3D DataArray with shape=(layer, lat, lon)
+        layer : int/string/float
+            Select the layer of 3D DataArray with multiple layers
+        dims : dictionary
```
ljwolf, line 402: So this is *always* filled with `lat`, `lon`, and `layer`? I think this needs to be generalized to `row`,`col`,`layer`, and I'd use them as separate arguments, not in this dict-style input.

libpysal/weights/distance.py
```diff
@@ -260,6 +263,54 @@ def from_dataframe(cls, df, geom_col='geometry', ids=None, *args, **kwargs):
...
+        da : xarray.DataArray
+            Input 2D or 3D DataArray with shape=(layer, height, width)
+        layer : int/string/float
+            Select the layer of 3D DataArray with multiple layers
+        dims : dictionary
+            Pass dimensions for coordinates and layers if they do not
+            belong to default dimensions, which are (band/time, y/lat, x/lon)
+            e.g. dims= {"lat": "latitude", "lon": "longitude", "layer": "year"}
+            Default is {} empty dictionary.
+        *args : KNN class arguments
```
ljwolf, line 282: I don't think there are any mandatory arguments for `knn` beyond the input data you derive from `da`. I don't think `*args` is necessary?

libpysal/weights/raster.py
```diff
@@ -0,0 +1,369 @@
...
+    >>> len(w.index)
+    7
+
+    See Also
+    --------
+    :class:`libpysal.weights.weights.W`
+    """
+    wsp = da2WSP(da, criterion, layer, dims)
+    w = wsp.to_W(**kwargs)
+    w.index = wsp.index
```
ljwolf, line 67: This is still adding an `.index` attribute to a `W` that makes this incompatible with other `W`, and will not round-trip through something like

`W.from_adjlist(w.to_adjlist())`

I think we need to make an issue for "refactoring the Weights class" and add this addition to that, to make sure we hold to our new ideas from the call on 21 July

libpysal/weights/contiguity.py
MgeeeeK: Sure I can look into making this more generalized. 

just to be clear `layer` argument only takes the value of single layer id e.g. if `DataArray` consists of multiple bands then user will pass the single band id using `layer` argument (same for time series DataArray, user can pass specific time value using `layer` argument). and the `dims` dictionary accepts the name of the corresponding dimensions.

MgeeeeK: Little update from last project call, mainly these 3 solutions sprung up :
  - First was that we can simply fix the names of the arguments and the dictionary keys. Naming of these 2 arguments `layer` and `dims` will be corrected to ~layer~ -> `z` and ~dims~ -> `coords_label`. dictionary keys will be named as  `x_label`, `y_label`, `z_label`. This style of passing data through dictionary is common with `xarray` users.
  - As you pointed out, we can add 3 label arguments straight to the function and remove the dictionary argument. This is definitely simpler and intuitive for user than other 2.
  - Another suggestion was (@slumnitz @darribas please correct me if I'm wrong) that we can remove label arguments and assume that user will pass the data in a specific structure (`z`, `y`, `x`) and if the data does not follow this shape then user should restructure and map the data to following order of dimensions.

Which one do you all think will be better suited for pysal?

MgeeeeK:
@darribas @ljwolf @slumnitz @sjsrey I think this pr is ready to be merged (once I added API fixes that levi suggested). 
I wanted to add optimized weights builder to this pr but that will make it bit lengthy (this already contains lot of stuff) therefore I'll add that functionality in another pr once this pr is merged. 
One more thing, rIght now `KNN.from_xarray` is failing due to different order of neighbors in different os environment. I will remove this functionality right now due to _ties_ conflict.

libpysal/weights/contiguity.py
slumnitz: @ljwolf, @darribas and @MgeeeeK to add on a couple thoughts here:
* Building a `Xarray.DataArray` (see [short explanations](http://xarray.pydata.org/en/stable/data-structures.html#dataarray)) the user needs to passes `data_vars`, and define `coords` (coordinates) with according `dims` (dimensions) using a dictionary. Working with `xarray` I think one get's used to the packages style of working with dims and coords with dictionaries when'd defining input data. I think what happens here and is demonstrated quite nicely in cell 13 in @Mgeeeek 's [example notebook](https://github.com/MgeeeeK/libpysal/blob/rast_interface/notebooks/Raster_awareness_API.ipynb), is that in order for us to process the passed `xarray.DataArray` properly, we need to pass the correct "wording" for the previously defined `coords` (i.e. the user needs to pass coords namings) which currently happens in the parameter called dims. I would here follow `xarray.DataArray` conventions defining coords, keep the dictionary style.
* we then require a Parameter to select which slice of the z or layer-coordinate to work with, here the `layer` parameter.

I generally think the way we are passing coordinate values (dims) if necessary and the actual slice that should be used (layer) is the way to go. As most people would only have to select a slice, and not touch the dims. If they do have to touch dims, they will be familiar with assigning coords similar to how you would assign coords in xarray. I think making separate Parameters for the currently named `dims` argument may add confusion. I certainly would find it easier to just stick to the way xarray is laying out their parameters. 

*Hence I would suggest to keep the given structure, but definitely change the naming.*

However, maybe there is a flaw in my logic right now and after all (please double check) and different people prefer different ways of doing things ;P! Generally I'd highly recommend having a look at the notebook, as it showcases how to work with arrays nicely!

darribas: I agree as much alignment as possible with `xarray` idioms is the way to go. +1 on this from my end!

libpysal/weights/raster.py
darribas: Has this been acted upon (ie. issue open)? Irrespective, I thins PR can go ahead? Or @ljwolf do you think we should add `id_order` for completion now, even if in reality we're relying on `index` anyway? Another alternative is to have legacy attributes built on-demand if they're queried (a bit like we do with other properties in `W` objects)? That'd "past-proof" it while keeping them slim and efficient?

darribas:
I'd be game to merge (we can add my suggestion of adding `id_order`, etc. for roundtrip as suggested above later as an additional feature?), assuming tests pass. I'd say for this one, KNN out is a good way to go. Also, KNN will not be that relevant with the new contiguity builder @MgeeeeK is cooking up on [#10](https://github.com/MgeeeeK/libpysal/pull/10), which allows you to set the order.

What do you think @ljwolf / @sjsrey ?

libpysal/weights/raster.py
ljwolf: [Yes](https://github.com/pysal/libpysal/projects/4), albeit only currently in the "exploration" phase to suss out which backend to pick.

darribas:
OK this looks great to me! @sjsrey, @slumnitz if you OK this (or go ahead to merge) this is good to go! 🎉

slumnitz (approved):

slumnitz:
Happy to see it merged too! Thank you for all your hard work @MgeeeeK! This looks amazing! Maybe we should all merge it together in a celebration tonight :)

PR #321, opened 2020-07-14
represent islands as self-joins with zero weight in adjacency lists

ljwolf:
justification in point 5. 

4. [x] You have [assigned a
Adjacency lists are one of the three major representations of graphs. We need to figure out how to accurately represent islands in their structure, rather than simply warning that islands will be dropped. I propose using a simple syntax, where the `neighbor_id` of an island is set to ~`None`~ *the same as the `focal_id`* and the weight to ~`numpy.nan`~ `0`. This PR provides this in the `to_adjlist()` method, and will examine (and fix) the `from_adjlist()` method if necessary.


ljwolf:
Actually, there's a type consistency issue with using the neighbor_id` as `None`, since that changes the `dtype` of `adjlist.neighbor`. 

So, @sjsrey's initial proposal is probably the right way to deal with this:
> what about adding a self-loop with zero weight?

ljwolf:
seems to be an issue with round-tripping the weights when IDs are used. I am not sure, but I have seen this keyerror before when building weights from a dictionary with non-integer ids.

ljwolf:
The issue ended up being #322 
I have put a workaround in the tests to avoid this issue. 

I can include the suggested fix to #322 if approved?

ljwolf:
The test failures appear to be on code that I've changed. But, in the traceback, the code is *not* changed. This is odd! Any advice?

jGaboardi:
> The test failures appear to be on code that I've changed. But, in the traceback, the code is _not_ changed. This is odd! Any advice?

Is odd... GitHub seems to be a bit funky today. Maybe try another empty commit to the PR to see if that triggers the correct code for the GHA? Also, looks like I was added and removed as a reviewer here, but still listed as a reviewer along with Wei. Review desired or not?

ljwolf:
Not sure why folks were removed from the request? I had intended to get @weikang9009 @jGaboardi and @sjsrey

jGaboardi:
> The test failures appear to be on code that I've changed. But, in the traceback, the code is _not_ changed. This is odd! Any advice?

Can confirm that your changes pass locally.

jGaboardi:
> The issue ended up being #322
> I have put a workaround in the tests to avoid this issue.
> 
> I can include the suggested fix to #322 if approved?

I think the inclusion of #322 is a good idea, pending an explanation of the set name (`out_islands`).

libpysal/weights/weights.py
```diff
@@ -301,18 +302,20 @@ def to_adjlist(
...
-                "adjacency list will drop these observations!"
+        except (ImportError, ModuleNotFoundError):
+            raise ImportError(
+                "pandas must be installed & importable to use this method"
             )
-        adjlist = pd.DataFrame(
-            ((idx, n, w) for idx, neighb in self for n, w in list(neighb.items())),
-            columns=("focal", "neighbor", "weight"),
-        )
+        n_islands = len(self.islands)
```
weikang9009, line 309: `n_islands` is not further used and could be dropped.

ljwolf:
Fixing #322 is not as simple, since this affects most of the rest of the weights id machinery. 

I'd suggest we merge this for now, and address #322 in the project about refactoring the weights.

martinfleis (changes requested):
I have a few points here.

1) This is a breaking change. If you do not care about weights but want to know neighbours to do other things, this adds a new row which may affect the result. For example, if I want to know all pairs and measure mean distance between them, this adds 0 and affects the value (this would in fact break one function in momepy).

2) Can we add a keyword like `include_islands` to control this behaviour? In the example above, I just do not want them at all since they do not have neighbours. This could start with `False` as a default with a warning that in the future it will be `True` to avoid a hard breaking change here.

sjsrey (approved):

ljwolf:
I've addedthe option outlined by @martinfleis, and I've set it so that, by default, islands are dropped. I would like to transition to *not* dropping islands quite quickly, as formats should not be lossy.... to_adjlist()/from_adjlist() should be inverses, and they're not at the moment. 

This is now needed for bugfixes over in [pysal/esda#190](https://github.com/pysal/esda/pull/190), so I hope to merge & cut a point release shortly.

martinfleis:
We can do what we often do in geopandas. Set `drop_islands=None` and warn that this now defaults to True but it will default to False in the future. If you set it manually to True/False, no warning. Then, in the next release, we can change the default to False and remove the warning.

ljwolf:
Great, have done this!

martinfleis:
The default value on line 278 should be `None`, not `True`.

ljwolf:
Yep, sorry again; I had a local issue with ce1b1eb, so I rolled back & redid the changes manually. Missed that. Thanks!

martinfleis (approved):
Thanks!

PR #335, opened 2020-08-11
ENH: use query_bulk in fuzzy_contiguity

### LINKED ISSUE #334: BUG: fuzzy_contiguity picks more neighbours than it should
I am working on `fuzzy_contiguity` refactoring based on geopandas `query_bulk` and I've realized that current implementation has two bugs in it. Both in the sindex query.

https://github.com/pysal/libpysal/blob/9e6595171013296fd12777f767fbc2e7c5a1a355/libpysal/weights/util.py#L1565-L1571

1) `ids = possible.intersects(geom).index.tolist()` lists all indices, not only those with True value, which means that all geometries which bounds intersect with tree bounds are marked as neighbours.

2) `ids.remove(i)` expects `RangeIndex`, and raises ValueError on anything else.

This is how it looks in reality (this is actually from tests):

download

All blue polygons are marked (and tested) as neighbours, whilst the top left is clearly intersecting only bounding box.

The fix is on the way, I just have a question - what should be used as an index here (also considering https://github.com/pysal/libpysal/projects/4)? From the code, I feel the intention to use the actual index, but `Queen` uses integer. Due to the bug above, `fuzzy_contiguity` currently tries to do index, but in reality does only integer.

martinfleis:
Reimplements core part of `fuzzy_contiguity` using new `sindex.query_bulk` introduced in GeoPandas 0.8.0. It brings cleaner code (once compatibility layer is removed), performance enhancements (`438 ms` vs `2.58 s` for `map_RS_BR.shp`) and allows specification of `predicate`, which means that we can generate weights not only based on `intersects` but also `'within', 'contains', 'overlaps', 'crosses', 'touches'` (plus more once pygeos/shapely 2.0 will support them).

At the same time resolves #334.

Note that it now uses the actual index of the GeoDataFrame, not an integer (as per @sjsrey's suggestion).


libpysal/weights/util.py
```diff
@@ -13,9 +13,12 @@
 from collections import defaultdict
 from itertools import tee
 from ..common import requires
+from distutils.version import LooseVersion
```
jGaboardi, line 16: Would there be any drawback to using [`packaging.version.parse`](https://packaging.pypa.io/en/latest/version/#packaging.version.parse) here?

```python
In [1]: import geopandas
In [2]: from packaging import version
In [3]: version.parse(geopandas.__version__) >= version.parse("0.8.0")
Out[3]: True

```

libpysal/weights/util.py
```diff
@@ -1552,23 +1566,36 @@ def fuzzy_contiguity(gdf, tolerance=0.005, buffering=False, drop=True, buffer=No
...
             # buffer each shape
             minx, miny, maxx, maxy = gdf.total_bounds
-            buffer = tolerance * 0.5 * abs(min(maxx-minx, maxy-miny))
+            buffer = tolerance * 0.5 * abs(min(maxx - minx, maxy - miny))
         # create new geometry column
         new_geometry = gdf.geometry.buffer(buffer)
         gdf['_buffer'] = new_geometry
         old_geometry_name = gdf.geometry.name
         gdf.set_geometry('_buffer', inplace=True)
-    assert gdf.sindex, 'GeoDataFrame must have a spatial index. Please make sure you have `libspatialindex` installed'
```
jGaboardi, line 1561: Is there a reason to remove the `assert gdf.sindex` statement?

libpysal/weights/util.py
martinfleis: Do we already require `packaging` as a dependency? Because `distutils` is built-in module, `packaging` needs to be installed. If we already have it, then it is probably fine to use it. If not, I don't see the benefit of adding another dependency.

libpysal/weights/util.py
martinfleis: It is duplicating the same [check from GeoPandas](https://github.com/geopandas/geopandas/blob/b366d890c35d68a52f7f608e47f00b89eaa85850/geopandas/sindex.py#L42-L44). Also, currently you can use either `rtree` (i.e.`libspatialindex`) or `PyGEOS` and there are plans to have pluggable `sindex` in GeoPandas, so it is ideal to keep this, including up-to-date error messages, to be handled at the source.

libpysal/weights/util.py
jGaboardi: > Do we already require `packaging` as a dependency? Because `distutils` is built-in module, `packaging` needs to be installed. If we already have it, then it is probably fine to use it. If not, I don't see the benefit of adding another dependency.

Great point. I somehow missed that `packaging` isn't included as a built-in.

libpysal/weights/util.py
```diff
@@ -1552,23 +1566,36 @@ def fuzzy_contiguity(gdf, tolerance=0.005, buffering=False, drop=True, buffer=No
...
+        # query tree based on set predicate
+        inp, res = gdf.sindex.query_bulk(gdf.geometry, predicate=predicate)
+        # remove self hits
+        itself = inp == res
+        inp = inp[~itself]
+        res = res[~itself]
+
+        # extract index values of neighbors
+        for i, ix in enumerate(gdf.index):
+            ids = gdf.iloc[res[inp == i]].index.tolist()
```
jorisvandenbossche, line 1587:
```suggestion
            ids = gdf.index[res[inp == i]].tolist()
```

If you only need the indices, instead of indexing the dataframe, you can directly index the index, which will be faster (and avoiding an unnecessary copy of the subset of the data)

Issue #328, opened 2020-07-23
Distance Band functionality for Raster Interface


MgeeeeK:
Following along [libpysal/#293](https://github.com/pysal/libpysal/issues/293) and weights object meeting, this issue documents the discussion for the addition of `DistanceBand` functionality inside raster interface.

Background
- In the case of raster the neighborhood obtained in queen or rook contiguity can be very small for further analysis. 
- Using KNN we can grow the number of neighbors (current implementation of `KNN` converts cell centroids into points and then ship directly into `KNN` builder) but this would not resolve the _ties_ issue particularly for rasters.
- Can also use `higher_order`, but right now it removes lower order contiguities (which is also raised here [libpysal/#313](https://github.com/pysal/libpysal/issues/313))

Potential Solution
- As suggested in the weights meeting, `DistanceBand` can be a good alternative approach for obtaining larger neighbors of each cell.
- One potential idea that was floated in the meeting was to first implement `DistanceBand` functionality and then map `higher_order` operation to the `DistanceBand`.

MgeeeeK:
@darribas @slumnitz

martinfleis:
> One potential idea that was floated in the meeting was to first implement DistanceBand functionality and then map higher_order operation to the DistanceBand.

Can you elaborate on this? In my eyes, these two are based on different concepts. `DistanceBand` captures geographical position of elements, while `higher_order` topological. I am trying to understand how do you envisage one to be used within another.

I wasn't part of the meeting, but in my experience, using `higher_order` adapted to `k<=n` instead of `k==n` as mentioned in #313 works very well if you need larger neighbourhood and may work equally well for raster data. We just need an efficient algorithm for that (both computationally and memory efficient - the one proposed in the issue may eat a lot of ram for higher `k`).

ljwolf:
The suggestion in the meeting was that inclusive higher-order Rook/Queen weights are equivalent to distance-band weights for a given radius. We might be able to make that fast using a `query_ball_tree` strategy. 

It was only a suggestion for exploration if performance issues persisted with a performant sparse adjacency matrix-based algo.

ljwolf:
Although, now thinking about it, Distance Bands are *not* reducible to higher-order rook/queen in the case of nulls:
```
- - - + - - - - 
- - + + + - - -
. . + + + + - - 
o . . X + + + -
- o . . + + - - 
- - o . . - - - 
- - - o . . - - 
```
[...]

Issue #313, opened 2020-07-05
higher_order weights of <= k


martinfleis:
Would it be of interest to have an option in `higher_order` and consequently `higher_order_sp`, which returns not only contiguity weights of order `k`, but also of order `<=k`? `higher_order_sp(w, shortest_path=False)` is close, but misses some features which are accessible on `k-n` but not `k`.

Implementation-wise it would be trivial, just using the following:

```python
wk = sum(map(lambda x: w ** x, range(2, k + 1)))
```

instead of this, in that case, and setting `shortest_path` to `False`:
https://github.com/pysal/libpysal/blob/10c4089b05f418b11ffdb36321d2455afa3b60fa/libpysal/weights/util.py#L499

In `momepy` I have a function simulating this as

```py
joined = first_order
for i in list(range(2, k + 1)):
    i_order = libpysal.weights.higher_order(first_order, k=i)
    joined = libpysal.weights.w_union(joined, i_order)
```

but the `map` approach above is of an order of magnitude faster.

I can make a PR if there's interest.

sjsrey:
Linking to #293 as I think there are some synergies here.

sjsrey:
Having the option to support both the "pure" higher order neighbor weights (that do not include lower order contiguities), and the "encompassing" higher order weights (that do include the lower order contiguities) would be a good thing, as different analytical methods make use of these different cases.

Mirroring the treatment in [GeoDa](https://geodacenter.github.io/workbook/4a_contig_weights/lab4a.html#higher-order-contiguity) would be a good idea (imo).

PR #343, opened 2020-08-28
[WIP]: Optimized raster-based weights builder

MgeeeeK:
This pull request adds a new optimized weight builder for the raster interface.

Main takeaway:
- Sparse matrix is build using rows and cols values which are generated iteratively. 
- The parallel implementation is based on the crand engine from [esda/#116](https://github.com/pysal/esda/pull/116), though this work uses `threading` backend.
- This pr also adds functionality to add `higher_order` neighbors, this is done using 2 ways:
  - Firstly, 2 arguments are added to the builder method (`k=int` and `include_nas=boolean`). `k` is used for specifying the kth order and `include_nas=True` is used for including higher-order neighbors even if lower-order neighbors are missing.
  - `higher_order` functionality is based on the suggestion provided by Martin in this issue [libpysal/#issue313](https://github.com/pysal/libpysal/issues/313). This also adds diagonal elements that are first converted to zero and then these zero elements are removed in the last step. But this changes the sparsity of the matrix!.
  - If `include_nas` is `True` then the method simply uses modular arithmetic to add higher-order neighbors. (with one caveat explained earlier and in this thread [libpysal/#328](https://github.com/pysal/libpysal/issues/328#issuecomment-664948667))

Here is the plot showing difference between two approaches.
Fig. 1
Fig. 2


sjsrey:
I can see the logic of offering the option for `include_nas`. So I'm +1 on that. I'm wondering if there is a more informative name for the argument, possibly? I'm not clear on what `nas` stands for?

MgeeeeK:
Sure! we can rename that argument. What do you guys suggest @darribas @slumnitz?
I think we assigned that as the temporary name because the functionality was not fully finalized.

libpysal/weights/tests/test_raster.py
```diff
@@ -27,25 +26,22 @@ def test_da2W(self):
...
+        self.assertEqual(sorted(w3.neighbors[2]), [1, 3, 5, 6, 7])
         self.assertEqual(w3.n, 16)
         self.assertEqual(w3.index.names, self.da3.to_series().index.names)
         self.assertEqual(w3.index.tolist(), self.da3.to_series().index.tolist())
 
-
-        
-
     def test_da2WSP(self):
         w1 = raster.da2WSP(self.da1, "queen")
```
darribas, line 46: I _think_ this file misses testing building a rook matrix on multi-core? Check, it might be an easy one to add to increase coverage.

libpysal/weights/contiguity.py
```diff
@@ -402,10 +429,19 @@ def from_xarray(cls, da, z_value=None, coords_labels={}, sparse=False, **kwargs)
...
-            e.g. dims = {"y_label": "latitude", "x_label": "longitude", "z_label": "year"}
+            e.g. coords_labels = {"y_label": "latitude", "x_label": "longitude", "z_label": "year"}
             Default is {} empty dictionary.
         sparse : boolean
-            type of weight object. Default is False. For sparse, sparse = True
+            type of weight object. Default is True. For libpysal.weights.W, sparse = False
+        k : int
+            Order of contiguity, this will select all neighbors upto kth order.
+            Default is 1.
+        include_nas : boolean
```
darribas, line 439: Same as above, make sure it's replaced throughout

libpysal/weights/raster.py
```diff
@@ -23,8 +35,17 @@ def da2W(da, criterion="queen", z_value=None, coords_labels={}, **kwargs):
...
-        e.g. dims = {"y_label": "latitude", "x_label": "longitude", "z_label": "year"}
+        e.g. coords_labels = {"y_label": "latitude", "x_label": "longitude", "z_label": "year"}
         Default is {} empty dictionary.
+    k : int
+        Order of contiguity, this will select all neighbors upto kth order.
+        Default is 1.
+    include_nas : boolean
+        If True, missing values will be assumed as non-missing when
+        selecting higher_order neighbors, Default is False
+    n_jobs : int
```
darribas, line 46: Add what's the default

libpysal/weights/raster.py
```diff
@@ -82,14 +117,27 @@ def da2WSP(da, criterion="queen", z_value=None, coords_labels={}):
...
+        Number of cores to be used in the sparse weight construction. If -1,
+        all available cores are used.
 
     Returns
     -------
-    w : libpysal.weights.WSP
+    wsp : libpysal.weights.WSP
        instance of spatial weights class WSP
```
darribas, line 135: Here it might be good also to add more information about the index built

libpysal/weights/raster.py
```diff
@@ -122,24 +170,89 @@ def da2WSP(da, criterion="queen", z_value=None, coords_labels={}):
...
-            id_order = np.where(mask)[0].tolist()
+    dtype = np.int32 if (shape[0] * shape[1]) < 46340 ** 2 else np.int64
+    if "nodatavals" in da.attrs and da.attrs["nodatavals"]:
+        mask = (ser != da.attrs["nodatavals"][0]).to_numpy()
+        ids = np.where(mask)[0]
+        id_map = _idmap(ids, mask, dtype)
+        ser = ser[ser != da.attrs["nodatavals"][0]]
+    else:
+        ids = np.arange(len(ser), dtype=dtype)
+        id_map = ids.copy()
```
darribas, line 188: Make sure this is tested

libpysal/weights/raster.py
```diff
@@ -122,24 +170,89 @@ def da2WSP(da, criterion="queen", z_value=None, coords_labels={}):
...
+
+    n = len(ids)
+
+    try:
+        import numba
+    except (ModuleNotFoundError, ImportError):
+        warn(
+            "numba cannot be imported, parallel processing "
+            "and include_nas functionality will be disabled. "
+            "falling back to inefficient method"
```
darribas, line 198: Maybe replace "inefficient method" by "slower version"

libpysal/weights/raster.py
```diff
@@ -177,10 +290,11 @@ def w2da(data, w, attrs={}, coords=None):
     """
     if not isinstance(w, W):
         raise TypeError("w must be an instance of weights.W")
-    if hasattr(w, 'index'):
+    if hasattr(w, "index"):
         da = _index2da(data, w.index, attrs, coords)
     else:
-        raise AttributeError("Cannot convert deprecated W with no index attribute")
+        raise AttributeError(
+            "Cannot convert deprecated W with no index attribute")
```
darribas, line 297: I'd replace the message by on that is more informative about what is required. Maybe something like:

```
`w` requires to include a `index` attribute that is built as a `pandas.MultiIndex` object. This will be used to build the coords in the resulting `da`
```

Something like that, you might have to tweak the message a bit with some of the information you include in the description of the index built by `da2WSP`

The future structure of weights object is still a bit in flux, so I'd not give the impression that is clear and the current one will be phased out.

Replace in all instance (it's on a couple other places)

libpysal/weights/raster.py
```diff
@@ -308,29 +423,34 @@ def _da_checker(da, z_value, coords_labels):
...
         raise TypeError("da must be an instance of xarray.DataArray")
     if da.ndim not in [2, 3]:
         raise ValueError("da must be 2D or 3D")
-    if not (np.issubdtype(da.values.dtype, np.integer) or
-            np.issubdtype(da.values.dtype, np.floating)):
-        raise ValueError(
-            "da must be an array of integers or float")
+    if not (np.issubdtype(da.values.dtype, np.integer)
+            or np.issubdtype(da.values.dtype, np.floating)):
+        raise ValueError("da must be an array of integers or float")
```
darribas, line 433: This should be another easy one to test to increase coverage

libpysal/weights/raster.py
```diff
@@ -122,24 +170,89 @@ def da2WSP(da, criterion="queen", z_value=None, coords_labels={}):
...
+        sw.setdiag(0)
+        sw.eliminate_zeros()
+        sw.data[:] = np.ones_like(sw.data, dtype=np.int8)
+
+    # Since a raster can contain missing data, and some
+    # of the functionality in pysal uses id_order property.
+    # Therefore, ids array is shipped to the WSP builder
+    # as this contains ids of only non-missing values which
+    # differ from range(len(ids)).
+    wsp = WSP(sw, index=ser.index, id_order=ids.tolist())
```
darribas, line 255: I'd vote here to use `sw.index.tolist()` as the values for `id_order`. It's repetition but it's better for consistency and allows us to be more backwards compatible with the traditional model, but also to support the idea of an `index` approach to weights

darribas (changes requested):
Almost there!!! Mostly issues of documentation and style, the only more substantive bit is to ensure the `id_order` of the weights we build is relatable to the `DataArray`

libpysal/weights/raster.py
darribas: Please add a short description about the `index` attribute that `wsp` is returned with, since this will not be the same across the library. Something like: "instance of spatial weights class WSP with an `index` attribute that includes a MultiIndex object from the DataArray"

libpysal/weights/tests/test_raster.py
darribas: Has this been added?

darribas (changes requested):
Just a couple minor bits to add, only docs and tests and I think this is good to ship!

libpysal/weights/raster.py
slumnitz: This is one of the last follow ups, I still detect.

libpysal/weights/tests/test_raster.py
slumnitz: This still seems outstanding @MgeeeeK

libpysal/weights/raster.py
slumnitz: @MgeeeeK you can test if a ValueError is raised here, this is the only line not covered, the rest seems good to go now.

libpysal/weights/raster.py
slumnitz: Codecov indicates that it is currently not tested.

libpysal/weights/raster.py
slumnitz: Looks good to me, what do you think @darribas?

libpysal/weights/contiguity.py
```diff
@@ -389,7 +407,16 @@ def from_dataframe(cls, df, geom_col='geometry', **kwargs):
...
+        da,
+        z_value=None,
+        coords_labels={},
+        k=1,
+        include_nas=False,
+        sparse=True,
+        **kwargs,
+    ):
         """
         Construct a weights object from a xarray.DataArray.
```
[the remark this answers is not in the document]
slumnitz: Documentation looks great! This is resolved? What do you think @darribas?

slumnitz (changes requested):
@MgeeeeK this looks great and I think we are super close to merging. The only thing remaining is to test a couple more lines. I resolved all of @darribas comments that where addressed and looked great. There are a couple minor comments still standing that slipped through the radar and are still open (even if marked outdated). Once these are added I am happy to get this merged. Thank you so much for your hard work this and last year!

libpysal/weights/tests/test_raster.py
MgeeeeK: oops my bad...

libpysal/weights/contiguity.py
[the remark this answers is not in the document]
darribas: This looks good to go, resolved

darribas (commented):
I've resolved a couple of the standing threads, they look good. I'm not sure `test_da2WSP` is being fully tested. I can see `rook` and `queen` are now both tested, but it's always in the multi-core context. Is it a different bit of code that runs it on single core? If not, it's fine, but if it is, we should test that too.

MgeeeeK:
> Is it a different bit of code that runs it on single core?
No, multi-thread is extension of the base constructor but still I'll add another test comparing results of multi-threaded and single-threaded builder.

libpysal/weights/tests/test_raster.py
```diff
@@ -12,78 +11,91 @@ def setUp(self):
...
         self.assertEqual(w1.index.names, self.da1.to_series().index.names)
         self.assertEqual(w1.index.tolist()[0], (1, 90.0, 180.0))
         self.assertEqual(w1.index.tolist()[1], (1, -30.0, -180.0))
         self.assertEqual(w1.index.tolist()[2], (1, -30.0, 180.0))
         self.assertEqual(w1.index.tolist()[3], (1, -90.0, -60.0))
-        w2 = raster.da2W(self.da2, "queen")
-        sw = lat2SW(4, 4, "queen")
-        self.assertEqual(w2.sparse.nnz, sw.nnz)
-        self.assertEqual(w2.sparse.todense().tolist(), sw.todense().tolist())
+        w2 = raster.da2WSP(self.da2, "queen", k=2, include_nodata=True)
```
MgeeeeK, line 67: Test for single and multi-threaded results, this also compares `include_nodata = True` and `include_nodata = False` for `k=2` (Since there is no missing value, output in both the case will be same)

darribas:
I think this is good to go as far as I'm concerned. @slumnitz ? If both give it green light, I think @sjsrey or us can merge?

darribas (approved):

slumnitz (approved):
Let's get this merged then! :)

PR #349, opened 2020-10-05
Format UserWarning string in W.to_adjlist

martinfleis:
Warning expects a number of islands, but the formatting is not there.

### DIFF
--- a/libpysal/weights/weights.py
+++ b/libpysal/weights/weights.py
@@ -306,7 +306,7 @@ class W(object):
         n_islands = len(self.islands)
         if n_islands > 0 and (not self.silence_warnings):
             warnings.warn(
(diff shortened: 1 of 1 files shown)


jGaboardi (approved):

ljwolf:
Hi, thanks for this! I'd rather merge #321 though, which has stalled after approval. 

That would remove this warning entirely, and address the source of the warning...

jGaboardi:
> Hi, thanks for this! I'd rather merge #321 though, which has stalled after approval.
> 
> That would remove this warning entirely, and address the source of the warning...

@ljwolf Are we waiting for other's approval for #321?

martinfleis:
I wasn't aware of #321. That would be a better solution (although I left some comments there).

ljwolf:
closed by #321

PR #356, opened 2020-10-25
[WIP] initial draft of network weights 

knaaptime:
this starts a new weights module that creates a neighbor relationship based on the shortest path through a pandana network.

Currently the weights are set to distance, so need to include some inverse distance weighting functions, but this is a start for generating feedback


stuartlynn:
Some thoughts about handling situations where the origin or destination dataframes are polygons rather than points. 

Currently, I think the approach is to use the geometry centroid as the point geometry and snap that to the closest node of the network. 

One can imagine other strategies that might be more appropriate in other cases. For example 

1) Using a representative point on the polygon instead of a centroid 
2) If the origin polygons have population like data and we can assume that data is uniform across the polygon, assigning some weighted fraction of the orgin polygon data to each network node that the polygon intersects.

It would be interesting to include a mechanism to specify different strategies for assigning polygon data to the network. Perhaps a function that can be passed into the from_dataframe function with a signature similar to 

```python
"""
parameters
--------------
polygon:  The origin or destination polygon  + properties as a Data Frame
intersecting_nodes: A list of nodes that intersect with the polygon 

returns
---------
A DataFrame of the nodes to be used with an associated weight value
"""

def assign_polygon_data_to_network(polygon,intersecting_nodes):
   return nodes, weights
```

An implementation of this function for centroid assignment would simply find the closest intersecting_node to the polygon centroid and return that with a weight of 1.

While an implementation for assigning equal weights for each of the intersecting nodes would return the list of intersecting_nodes with `1/len(intersecting_nodes) `as the weights for each. 

This probably isn't quite the right way to do this but hopefully gets across the idea and can spark some discussion.

knaaptime:
as @stuartlynn raises above, [centroids are sketchy](https://onlinelibrary.wiley.com/doi/10.1111/gean.12254) so will be cool to consider other options for attaching polys to the network

jGaboardi:
> as @stuartlynn raises above, [centroids are sketchy](https://onlinelibrary.wiley.com/doi/10.1111/gean.12254) so will be cool to consider other options for attaching polys to the network

This is what I did my dissertation on [`pp2n`](https://github.com/jGaboardi/pp2n). I am planning on making it part of [`spaghetti`](https://github.com/pysal/spaghetti) eventually. Definitely here, too, if desired.

stuartlynn:
@jGaboardi that looks great!

knaaptime:
the error looks to be specific to windows/fiona/py3.6, but otherwise i think this is basically passing

jGaboardi:
> the error looks to be specific to windows/fiona/py3.6, but otherwise i think this is basically passing

Yes, and I think specific to the feedstock. I opened an issue over there last week.

https://github.com/conda-forge/fiona-feedstock/issues/171

libpysal/weights/network.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    origins = origins.copy()
+    destinations = destinations.copy()
+
+    origins["osm_ids"] = network.get_node_ids(
+        origins.centroid.x, origins.centroid.y
+    ).astype(int)
+    destinations["osm_ids"] = network.get_node_ids(
+        destinations.centroid.x, destinations.centroid.y
+    ).astype(int)
```
martinfleis, line 116: We could check if `origins == destinations` and copy `osm_ids` in that case. I guess that this can be costly operation, so it is worth skipping if we can.

libpysal/weights/network.py
```diff
@@ -0,0 +1,270 @@
...
+    ods = []
+
+    if not index_orig:
+        origins['idx'] = origins.index.values
+        index_orig = 'idx'
+    if not index_dest:
+        destinations['idx'] = destinations.index.values
+        index_dest = 'idx'
+
+    # I dont think there's a way to do this in parallel, so we can at least show a progress bar
```
martinfleis, line 127: You could have `origins["osm_ids"]` as `dask.Series` and apply over it in parallel, but that may be overkill. Isn't `network.shortest_path_lengths` parallelised in pandana?

martinfleis (commented):
I feel that the most important part here will be the integration of `NetworkW` with the rest of the ecosystem, i.e. the simple way of creating `pandana.Network` from LineString GeoDataFrame, OSMnx-like networkx.Graph, momepy-like networkx.Graph, spaghetti.Network and likely some other options.

Btw, do we know what is the performance of pandana's shortest paths vs networkx? I guess it is much faster since it is vectorised C code, but I haven't found it.

libpysal/weights/network.py
```diff
@@ -1,6 +1,5 @@
 import pandas as pd
```
jGaboardi, line 1: I'm assuming having `pandas` imported twice is simply a mistake?

libpysal/weights/network.py
```diff
@@ -0,0 +1,270 @@
...
+        .apply(lambda x: x[0])
+    )
+    feeds.name = feeds.name.fillna(feeds.onestop_id)
+    feeds = (
+        feeds[["name", "url"]].dropna(subset=["url"]).set_index("name").to_dict()["url"]
+    )
+    return feeds
+
+
+class NetworkW(W):
```
jGaboardi, line 42: I know this PR is a WIP, but would it be possible to clean up the docstrings to conform with pysal/pysal#1174 once it is nearing non-WIP status?

libpysal/weights/network.py
knaaptime: yup. these functions were spread across a few different files, then i copied them here without scouring closely enough

libpysal/weights/network.py
knaaptime: definiteky

libpysal/weights/network.py
knaaptime: > do we know what is the performance of pandana's shortest paths vs networkx? I guess it is much faster since it is vectorised C code, but I haven't found it. 

i think i can find it somewhere. Its not only that pandana is vectorized c, but it also uses contraction hierarchies instead of e.g. dijkstra

libpysal/weights/network.py
knaaptime: i've been wondering about whether this is dask-able for massive datasets. Still very curious to see if it works/how performant it could be, so i may try out some explorations

libpysal/weights/network.py
jGaboardi: I can also work on this once thing are more concrete.

libpysal/weights/network.py
martinfleis: Partially. You will always need to fit the whole `Network` object in memory. But you can use dask to parallelize that for loop. I don't think it is possible to actually distribute this computation to a cluster, but on a single machine, the parallel loop may help. One way how to do that is here https://urbangrammarai.github.io/spatial_signatures/measuring/morphometrics.html#generate-spatial-weights-w

libpysal/weights/network.py
knaaptime: >We could check if origins == destinations and copy osm_ids in that case. I guess that this can be costly operation, so it is worth skipping if we can. 

thats a really good point. I originally wrote this for use [with access](https://github.com/knaaptime/healthacc), where origins and destinations aren't always the same, but in think in the context of a `W`, they *are*, so we can skip the second call to `get_node_ids`
```suggestion
    origins["osm_ids"] = network.get_node_ids(
        origins.centroid.x, origins.centroid.y
    ).astype(int)
    destinations["osm_ids"] = network.get_node_ids(
        destinations.centroid.x, destinations.centroid.y
    ).astype(int)
```

libpysal/weights/network.py
```diff
@@ -0,0 +1,269 @@
...
+            except ImportError:
+                raise ImportError('You must have pandana installed to generate a Network')
+            assert 'epsg:4326' in df.crs.to_string(), "You must pass in an explicit pandana.Network object or provide a geodataframe in geographic (epsg:4326) coordinates"
+            network = osm.pdna_network_from_bbox(bbox=tuple(df.total_bounds))
+
+        if not geom_col == 'geometry':
+            df = df.set_geometry(geom_col)
+        adj = compute_travel_cost_adjlist(df, df, network, index_orig=ids, index_dest=ids)
+        if max_dist:
+            adj = adj[adj['cost'] <= max_dist]
```
martinfleis, line 81: One thing I just realised about this. `adj` is a dense matrix. If you set `max_dist`, you still have to create that dense matrix before, which means that you can quickly run out of memory. Would it make sense to do this check within the loop in `compute_travel_cost_adjlist`? It may be less efficient performance-wise, but it would not blow up on memory.

I was just thinking about my data with hundreds of thousands of buildings for which I'd like to get network W, realising this would probably kill the machine.

On a related note, it may be worth checking the option to generate W on demand (for all constructors) to avoid these memory issues altogether.

libpysal/weights/network.py
```diff
@@ -0,0 +1,269 @@
...
+        if not network:
+            try:
+                from pandana.loaders import osm
+            except ImportError:
+                raise ImportError('You must have pandana installed to generate a Network')
+            assert 'epsg:4326' in df.crs.to_string(), "You must pass in an explicit pandana.Network object or provide a geodataframe in geographic (epsg:4326) coordinates"
+            network = osm.pdna_network_from_bbox(bbox=tuple(df.total_bounds))
+
+        if not geom_col == 'geometry':
+            df = df.set_geometry(geom_col)
```
ljwolf, line 78: Here, you're looking to ensure that `df.geometry.name == geom_col`, right? 

I think it's clearer to have `geom_col=None` by default. Then, if `geom_col is None`, we take the existing geometry column name. Otherwise, set the geometry column to the provided `geom_col`. This would look something like:

```python
@classmethod
def from_dataframe(cls, df=None, network=None, geom_col=None, ids=None, max_dist=None, **kwargs):
... 
if geom_col is None:
    geom_col = df.geometry.name # in case it's needed later for any reason
else:
    df = df.set_geometry(geom_col)

libpysal/weights/network.py
ljwolf: Yes, definitely. This strategy works for sparse graphs, but any dense graph will have approximately n^2 entries. No fun.... 

To make these things built on the fly for a NetworkW, we'd need to make many attributes be cached properties. it's definitely doable in the same pattern as the rasterW GSOC project in #343

libpysal/weights/network.py
knaaptime: I think this is probably a fragment leftover from building this function up in a notebook where i set the geom to 'centroid' at some point but left the original geometry column (still named geometry) on the gdf. I dont think the function refers to the geom col by name elsewhere so i can probably just remove this section

libpysal/weights/network.py
knaaptime: ah, thanks. good point. I think moving that check up into the loop is probably the best way to go, but its worth noting that I sketched out a [different way](https://gist.github.com/knaaptime/0bb23a49ff523983629f5436c9cb6c72) of doing this that used pandana's `nearest_pois` functionality before they added the `shortest_path` method. 

The difference between the two approaches has to do with how/when the data get filtered. When using `nearest_pois`, pandana searches out the specified distance and returns all the nodes and their distances that fall within.  That means we have to set a parameter for the max total neighbors (which isnt terribly easy to estimate beforehand) *and* its calculating pairwise distances for every single node within the distance threshold

with `shortest_paths`, we're first filtering out nodes we don't care about, only calculating pairwise distances between the objects we've snapped to the OSM network (so e.g. going up to census tracts, that changes the resolution dramatically), but then we have to filter by distance afterward. So in the first case, we're including irrelevant nodes inside the distance threshold, in the second case we're including irrelevant nodes outside the threshold. I'm also not sure how [precomputation](http://udst.github.io/pandana/network.html#pandana.network.Network.precompute) may potentially factor into the difference

the raison d'etre for pandana is, more or less, to create spatial lag variables that can be used in urbansim's hedonic and location choice models, and it's designed explicitly to work at the types of scales you describe @martinfleis --only now we have to figure out how to translate that same efficiency to generate the full `W` object instead of just the lag. I wonder if @fscottfoti or @smmaurer have any thoughts about how best to do this? I think it could be useful for pandana users generally, because once you have the `W` object generated from the `pandana.Network` it lets you calculate different accessibility metrics or change the decay function/constant (e.g. using the generated travel matrix as [input to access](https://github.com/knaaptime/healthacc/blob/master/notebooks/walkthrough.ipynb) which would address issues [like this](https://github.com/UDST/pandana/issues/128)) and those could be incorporated into the UDST workflow as well

knaaptime:
thanks to some fantastic help from the UDST folks, the next version of pandana will [expose](https://github.com/UDST/pandana/issues/56) the vectorized range query, so this should reduce to a few lines of really fast code :). I'll take another pass once pandana 0.7 is out!

knaaptime:
resolved by #755

PR #363, opened 2020-11-22
Testing matrix

jGaboardi:
This PR provides a similar testing schema to that found in [`spaghetti`](https://github.com/pysal/spaghetti/blob/main/.github/workflows/unittests.yml#L19) where testing for the Python 3.6 environment is excluded on Windows. This exclusion gets tests back to green and is an intermediary solution to dropping support for Python 3.6 altogether.

xref: pysal/spaghetti#545, pysal/spaghetti#552, pysal/esda#152, and [gitter discussion](https://gitter.im/pysal/pysal?at=5fb448ecc10273610a106a51)


martinfleis:
Have you considered/tried adding py3.9? May be worth including it.

jGaboardi:
> Have you considered/tried adding py3.9? May be worth including it.

@martinfleis Great point. Let's make sure these all pass first then push forward with that. Also, feel free to push directly to PR if you have ideas, etc.

jGaboardi:
@martinfleis Looks like the environment solver [fails for 3.9 even on Ubuntu](https://github.com/pysal/libpysal/pull/363/checks?check_run_id=1439074514#step:3:1482). Might we have to leave off 3.9 for the time being?

martinfleis:
@jGaboardi Yep, it is apparently not time yet. I was adding it to geopandas where it works, but here we have a bit more complex environment.

martinfleis (approved):

PR #365, opened 2020-12-03
start on perimeter weighting with just pygeos

ljwolf:
This is to incorporate similar functionality from `segregation` into `libpysal` for perimeter-based areal weighting for `Rook` weights.


ljwolf:
to do:
1. handle optional standardization by allowing `perimeter_weighted='raw'` as a value. 
2. add tests
3. migrate segregation to this

knaaptime:
resolved by Graph

PR #367, opened 2020-12-09
ENH: use active geometry in from_dataframe

martinfleis:
`from_dataframe` methods of contiguity and distance weights now defaults to hard-coded `"geometry"` column. This PR uses active geometry column instead. In most of the cases, it will be the same. I encountered this behaviour with gdf with two named geometry columns (none of the called "geometry"). In such a situation it makes sense to default to active geometry.

Some `black` linting also happened  - can remove those if you want.

### DIFF
--- a/libpysal/weights/contiguity.py
+++ b/libpysal/weights/contiguity.py
@@ -133,7 +133,7 @@ class Rook(W):
 
     @classmethod
     def from_dataframe(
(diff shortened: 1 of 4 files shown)


jGaboardi (approved):
Good catch on this. Also, `black` is the way to go, so +1 on that, too.

sjsrey (approved):

sjsrey:
Nice change in the defaults.

Issue #364, opened 2020-11-23
`examples.explain` returning 404


darribas:
I've had several reports of this behaviour (including my own) on `pysal2.3.0`/`libpysal4.3.0`:

![Screenshot from 2020-11-23 10-33-27](https://user-images.githubusercontent.com/417363/99952500-aa82f780-2d77-11eb-88f5-80692c524268.png)

I'd imagine this has to do with a change in the API upstream at the @GeoDaCenter? Just guessing, might be something else.

sjsrey:
Thanks for reporting. My guess is something moved. WIll investigate.

martinfleis:
Closed by #368

PR #368, opened 2020-12-10
This address a move in remote datasets 

sjsrey:
Address #364


darribas (approved):
This seems to address a case and the changes look sensible, happy to merge @sjsrey if you think it's good to go.

PR #383, opened 2021-01-25
[WIP]: id_order as a property of WSP class

MgeeeeK:
This pull request adds the `id_order` as a property of the `WSP` class. It is complementary to the work added in [libpysal#343](https://github.com/pysal/libpysal/pull/343).

Background details of raster interface:

1. Earlier, `id_order` was used to map coordinate values (stored in `index` attribute) and the neighbors. It contained the list of indices (basically, `list(range(len(index)))`) in the int format to make the `W` object more memory efficient.
2. After some discussion, we decided to make it more backward compatible with the traditional model used in the libpysal. Therefore, `id_order` now holds the values of the `index` attribute in the form of a list (`index.tolist()`).
3. But after assigning `index.tolist()` to `id_order`, we observed that there was a significant performance loss when creating weights object (memory also takes a hit especially, when building `W` object) since the coordinates are stored as a list of tuples.

Making `id_order` as a property of the `WSP` class solved this problem to some extent since sparse weights only require `id_order` when building the `W` object. Though, it did not bring any improvements when creating a `W` object, which will be unusable when dealing with a large dataset.

Currently, I've added a warning when building the `W` object from raster. Until the refactoring is completed, maybe we can add a condition that allows `da2W` to continue its computation if the estimated memory usage of the function fits within the free memory.

### DIFF
--- a/libpysal/weights/raster.py
+++ b/libpysal/weights/raster.py
@@ -268,7 +268,7 @@ def da2WSP(
         sw.eliminate_zeros()
         sw.data[:] = np.ones_like(sw.data, dtype=np.int8)
     index = ser.index
-    wsp = WSP(sw, index=index, id_order=index.tolist())
+    wsp = WSP(sw, index=index)
     return wsp
 
 
--- a/libpysal/weights/raster.py
+++ b/libpysal/weights/raster.py
@@ -107,9 +107,11 @@ def da2W(
     :class:`libpysal.weights.weights.W`
     """
     warn(
-        "It is best to use the WSP object when building weights"
-        " from a raster, as the da2WSP method is much faster"
-        " and far more memory efficient than the da2W method."
+        "You are trying to build a full W object from "
+        "xarray.DataArray (raster) object. This computation "
+        "can be very slow and not scale well. It is recommended, "
+        "if possible, to instead build WSP object, which is more "
+        "efficient and faster. You can do this by using da2WSP method."
     )
     wsp = da2WSP(da, criterion, z_value, coords_labels, k, include_nodata, n_jobs)
     w = wsp.to_W(**kwargs)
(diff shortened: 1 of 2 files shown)


MgeeeeK:
@darribas @slumnitz

libpysal/weights/weights.py
```diff
@@ -1378,7 +1378,19 @@ def __init__(self, sparse, id_order=None, index=None):
         else:
             index = pd.RangeIndex(self.n)
         self.index = index
-        self._cache = {}
+
+    @property
```
darribas, line 1382: Just to confirm, this only gets triggered when called, not built by default?

libpysal/weights/raster.py
```diff
@@ -104,8 +106,12 @@ def da2W(
     --------
     :class:`libpysal.weights.weights.W`
     """
-    wsp = da2WSP(da, criterion, z_value, coords_labels, k, include_nodata,
-                 n_jobs)
+    warn(
```
darribas, line 109: I'd say something like:

>  "You are trying to convert/build a full `W` object from `DataArray` (raster) object and this can be very slow and not scale well. It is recommended, if possible, to instead build `WSP` objects, which are more efficient and faster. You can do this using `da2WSP` instead of `da2W`"

That way you provide feedback as to what they can do and also explicitly warn about what can happen with `da2W`.

darribas (commented):
Very minor comments

libpysal/weights/weights.py
MgeeeeK: Yes, it'll behave just like other properties

darribas (approved):
This looks great to me! If @slumnitz approves too, I'm happy to merge and get it in before the end of the week!

PR #387, opened 2021-02-01
make build + make sync @docsrc

darribas:
This PR has a build of the docs created by:

```shell
cd docsrc
make build
make sync
```

I _think_ this is what was expected, but happy to amend.


jGaboardi (approved):
As is, this looks good to me. If we're interested in adding a `build_docs.yml` workflow like in [`tobler`](https://github.com/pysal/tobler/blob/master/.github/workflows/build_docs.yml), [`spopt`](https://github.com/pysal/spopt/blob/main/.github/workflows/build_docs.yml), etc. I'd be happy to put that together, as well.

PR #389, opened 2021-02-02
add docs action workflow

jGaboardi:
Following up with #387, this PR:
* adds GHA for building docs when a release is [triggered](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L16) 
* cleans and the `docs/` directory and removes the `docsrc/` directory
* updates `conf.py` and `Makefile`

This action will not run until a release is [triggered](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L16). After the first run of the workflow *Setting -> GitHub Pages -> Source* must be updated to `gh-pages/{root}`. Once this is done `pysal.org/libpysal` will be rebuilt automatically after each release.


martinfleis:
> Once this is done pysal.org/libpysal will be rebuilt automatically after each successful PR.

Hold on, that means that there will be no live docs of the latest stable release? Since this does not allow the switch between stable/latest as readthedocs does. I don't think that is wise. Can we rebuild the docs after each tag rather than each PR?

jGaboardi:
> there will be no live docs of the latest stable release?

Correct, this is how it is done currently in [`esda`](https://github.com/pysal/esda/blob/master/.github/workflows/build_docs.yml), [`segregation`](https://github.com/pysal/segregation/blob/master/.github/workflows/build_docs.yml), [`spaghetti`](https://github.com/pysal/spaghetti/blob/main/.github/workflows/build_docs.yml), [`spopt`](https://github.com/pysal/spopt/blob/main/.github/workflows/build_docs.yml), [`spreg`](https://github.com/pysal/spreg/blob/master/.github/workflows/build_docs.yml), and [`tobler`](https://github.com/pysal/tobler/blob/master/.github/workflows/build_docs.yml).

> Since this does not allow the switch between stable/latest as readthedocs does. I don't think that is wise.

Also correct. @sjsrey brought this up in pysal/pysal#1216

> Can we rebuild the docs after each tag rather than each PR?

After you comment here it just came to me that we could probably simply add the [`tags: - 'v*'`](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L15) like we use in`release_and_publish.yml` instead of [`branches: main/master`](https://github.com/pysal/tobler/blob/ad51c9af6b57f45187d8c54477156114566be33e/.github/workflows/build_docs.yml#L4). Doing this may be a nice intermediary solution until we figure out pysal/pysal#1216.

Thoughts @martinfleis?

martinfleis (approved):
The workflow recipe looks good to me now, not sure if there's more to check.

sjsrey:
> Following up with #387, this PR:
> 
>     * adds GHA for building docs when a release is [triggered](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L16)
> 
>     * cleans and the `docs/` directory and removes the `docsrc/` directory
> 
>     * updates `conf.py` and `Makefile`
> 
> 
> This action will not run until a release is [triggered](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L16). After the first run of the workflow _Setting -> GitHub Pages -> Source_ must be updated to `gh-pages/{root}`. Once this is done `pysal.org/libpysal` will be rebuilt automatically after each release.

As we are going to cut a release, I wanted to clarify what "After the first run..." means.

My understanding is this workflow will only be triggered on a release tag. Does this mean we need to do a temporary release, and then update settings and do the intended release?

sjsrey (commented):
Added a question in the comments.

jGaboardi:
> My understanding is this workflow will only be triggered on a release tag. Does this mean we need to do a temporary release, and then update settings and do the intended release?

I has been a bit since I made this PR, but I believe the answer is "yes." Moreover, it may be prudent to do a `v4.4.0post1` (and increment up) until we are are sure things are working properly. Also, I add a [manual workflow trigger to `spaghetti`](https://github.com/pysal/spaghetti/blob/42c9fb3184e9894daf80a54a1cb1ff1ea117cd90/.github/workflows/build_docs.yml#L8-L13) whereby the docs can be regenerated at will should they need a quick refresh. We can add that trigger here if we like, too.

sjsrey:
> > My understanding is this workflow will only be triggered on a release tag. Does this mean we need to do a temporary release, and then update settings and do the intended release?
> 
> I has been a bit since I made this PR, but I believe the answer is "yes." Moreover, it may be prudent to do a `v4.4.0post1` (and increment up) until we are are sure things are working properly. Also, I add a [manual workflow trigger to `spaghetti`](https://github.com/pysal/spaghetti/blob/42c9fb3184e9894daf80a54a1cb1ff1ea117cd90/.github/workflows/build_docs.yml#L8-L13) whereby the docs can be regenerated at will should they need a quick refresh. We can add that trigger here if we like, too.

can you add the trigger to this or, and then we can merge?

sjsrey (approved):

sjsrey:
> Following up with #387, this PR:
> 
>     * adds GHA for building docs when a release is [triggered](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L16)
> 
>     * cleans and the `docs/` directory and removes the `docsrc/` directory
> 
>     * updates `conf.py` and `Makefile`
> 
> 
> This action will not run until a release is [triggered](https://github.com/pysal/libpysal/blob/ea866112419bf0cfa44e1ccc613963e201aa5938/.github/workflows/release_and_publish.yml#L16). After the first run of the workflow _Setting -> GitHub Pages -> Source_ must be updated to `gh-pages/{root}`. Once this is done `pysal.org/libpysal` will be rebuilt automatically after each release.

Do we need to create the `gh-pages` branch?

jGaboardi:
We do not. There was [an error](https://github.com/pysal/libpysal/runs/2926108889?check_suite_focus=true#step:4:15) while the workflow ran.

jGaboardi:
We need to swap out the current function for the following (see [`spaghetti`](https://github.com/pysal/spaghetti/blob/42c9fb3184e9894daf80a54a1cb1ff1ea117cd90/docs/conf.py#L248-L249)):

```python
def setup(app):
    app.add_css_file("pysal-styles.css")
```

jGaboardi:
I will do this in a PR now.

jGaboardi:
This should be resolved in #413

PR #395, opened 2021-04-04
xarray geograph implementation

ljwolf:
This xarray-based implementation has `.neighbors`, `.weights`, and `.sparse`. 

Some tricks would be needed to move `.neighbors` from being a method to being indexable, but that's solvable (just need a placeholder class with `__getitem__` iirc

### DIFF
--- a/libpysal/graphs/wxa.py
+++ b/libpysal/graphs/wxa.py
@@ -1,0 +1,40 @@
+from xarray import DataArray
+from scipy import sparse
+import numpy
+from sklearn.preprocessing import LabelEncoder
+
+
+class W_XA:
+    def __init__(self, adjlist):
+        self.xarray = DataArray.from_series(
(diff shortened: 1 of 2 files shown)


libpysal/graphs/wxa.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            return self._cache["sparse"]
+        except KeyError:
+            coo = self.xarray.data
+            self._cache["sparse"] = sparse.csr_matrix((coo.data, coo.nonzero()))
+            return self.sparse
+
+
+if __name__ == "__main__":
+    import geopandas, libpysal
```
jGaboardi, line 35: Are considering a hard dependency for `geopandas` for this or is it simply for test?

sjsrey: simply for test

Issue #393, opened 2021-04-04
implement converters for


sjsrey:
- [ ] xarray 
- [ ] adjacency table-based on "dataframe"

ljwolf:
These are both completed. However, we need to fix/modify the `to_adjlist()` function to handle islands (xref #320, #341)

PR #397, opened 2021-04-04
Draft of adjacency table as df

sjsrey:
#393

### DIFF
--- a/libpysal/graphs/wdf.py
+++ b/libpysal/graphs/wdf.py
@@ -11,13 +11,11 @@ class W_DF:
 
     def neighbors(self, ix):
         # assume that the ix is not iterable, like currently done in W
-        return self.adjlist[self.adjlist.focal==ix].neighbor
-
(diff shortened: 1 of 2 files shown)


libpysal/graphs/wdf.py
```diff
@@ -0,0 +1,40 @@
...
+        return self.adjlist[self.adjlist.focal==ix].weight
+
+
+    @property
+    def sparse(self):
+        try:
+            return self._cache["sparse"]
+        except KeyError:
+            weights = self.adjlist.weight
+            idxs = self.adjlist.focal.unique()
```
ljwolf, line 28: This assumes that all focals are neighbours & vice versa. I think it needs something like
```python
le = LabelEncoder.fit(bumpy.hstack((self.adjlist.focal, self.adjlist.neighbor))
rows = le.transform(self.adjlist.focal)
cols = le.transform(self.adjlist.neighbor)
```
or the equivalent `numpy.searchsorted` on the set of `unique(focal.union(neighbour))`

Issue #375, opened 2021-01-07
pygeos-based contiguity


martinfleis:
Contiguity weights (`Queen` and `Rook`) can be significantly faster if using pygeos to extract coordinates. Note that this is not a drop-in replacement of the current function since that can handle shapely and PySAL geometries (I've never worked with those). But it is good to keep this option in mind.

I also think that some of those loops can be done in Numpy or Numba to get it even more performant. However, I initially wanted to figure out how to ease memory demands, but did not succeed on that yet.

This is based on code in here: https://github.com/pysal/libpysal/blob/bc37eb438b17270c3e222941950450e93ff433ed/libpysal/weights/_contW_lists.py#L61

```py
import numpy as np
import pygeos
import geopandas as gpd
from shapely.geometry import Point
import libpysal
import collections

# create geoms in a grid 200x200
size = 200
grid = gpd.points_from_xy(
    np.repeat(np.linspace(1, size, size), size), np.tile(np.linspace(1, size, size), size)
).buffer(0.5, cap_style=3)

# these two lines do the trick, the rest is essentially copy&paste
coords = pygeos.get_coordinates(grid.data)
ints = np.repeat(np.array(range(len(grid))), pygeos.get_num_coordinates(grid.data))

items = collections.defaultdict(set)
for i, vertex in enumerate(coords):
    items[tuple(vertex)].add(ints[i])

shared_vertices = []
for item, location in list(items.items()):
    if len(location) > 1:
        shared_vertices.append(location)

w = {}
for i in range(len(grid)):
    w[i] = set()

for vert_set in shared_vertices:
    for v in vert_set:
        w[v] = w[v] | vert_set
        try:
            w[v].remove(v)
        except:
            pass
```

Timing: 
20x20 grid
current `Wall time: 68.2 ms`
new `Wall time: 5.85 ms`

200x200 grid 
current `Wall time: 2.25 s`
new `Wall time: 676 ms`

200x200 grid of circles (dense geometry)
current `Wall time: 16 s`
new `Wall time: 13.9 s`

ljwolf:
That'd be a great contribution! 

I pieced together our current `_get_boundary_points` to allow for ducktyping of pysal polygons and shapely polygons as part of our migration to being able to work with `shapely`/`geopandas` without explicit type conversions. The original design direction was that we needed to support the pysal polygons alongside shapely polygons, and ensure the functions worked without shapely. 

[...]

PR #402, opened 2021-04-15
Edgelist-based contiguity constructor

ljwolf:
This stubs a edgelist-based constructor for contiguity. I've also included Numba implementations.

The basic algorithm follows @jlaura's existing builder, except I hash the edge/point before constructing the edgelist. I do this because (I think) this will be faster in the Numba implementation, and (in my experience) Numba gets a bit finicky with collections-within-collections. 

Currently, though, I get failures on `_fast_index_and_flip` on linux, but not on Mac. Just from experimentation, a no-hash edgelist version (i.e. just use the coordinates/edge tuples directly) outperformed the `_contW_lists.py` approach consistently. I'm troubleshooting the Numba hash-based implementation now.

The included 'geograph.py' script should run the relevant new classes.


ljwolf:
This also incorporates #375

libpysal/graph/_vertex_set_contiguity.py
```diff
@@ -0,0 +1,125 @@
...
+    result = numpy.ones(n_edges)
+    starts = numpy.cumsum(coords_per_geom)
+    on_point = 0
+    edge_ix = 0
+    for i in range(n_coords):
+        on_point += 1
+        index = indices[i]
+        offset = starts[i]
+        tail = coords[i]
+        if on_point == coords_per_geom[i]:  # would modulo be faster?
```
ljwolf, line 93: This is what causes the crash. I think this should be `coords_per_geom[index]`, since
- `index` represents the index of the geometry for point `I`
- `coords_per_geom` is the `(n,)` vector of counts of points per geometry

This was causing an index-out-of-bounds, basically, but this was not getting caught (for some reason) as an immediate fail.

ljwolf:
With 926d15e I get the following timings from 20 runs:
<table border="1" class="dataframe">
  <thead>
    <tr>
      <th></th>
      <th colspan="2" halign="left">edgelist_pyhash</th>
      <th colspan="2" halign="left">edgelist_py</th>
      <th colspan="2" halign="left">edgelist_nuhash</th>
      <th colspan="2" halign="left">dict</th>
      <th colspan="2" halign="left">elist_agree</th>
      <th colspan="2" halign="left">correct</th>
    </tr>
    <tr>
      <th></th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
    </tr>
    <tr>
      <th>dataset</th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>columbus.shp</th>
      <td>0.001637</td>
      <td>0.000067</td>
      <td>0.001626</td>
      <td>0.000058</td>
      <td>0.001089</td>
      <td>0.000041</td>
      <td>0.003976</td>
      <td>0.000169</td>
      <td>True</td>
      <td>0.0</td>
      <td>True</td>
      <td>0.0</td>
    </tr>
    <tr>
      <th>south.shp</th>
      <td>0.099791</td>
      <td>0.026529</td>
      <td>0.092008</td>
      <td>0.028093</td>
      <td>0.676044</td>
      <td>0.007085</td>
      <td>0.169440</td>
      <td>0.028031</td>
      <td>True</td>
      <td>0.0</td>
      <td>True</td>
      <td>0.0</td>
    </tr>
    <tr>
      <th>NAT.shp</th>
      <td>0.179908</td>
      <td>0.032130</td>
      <td>0.191044</td>
      <td>0.035152</td>
      <td>2.522227</td>
      <td>0.062529</td>
      <td>0.347574</td>
      <td>0.042780</td>
      <td>True</td>
      <td>0.0</td>
      <td>True</td>
      <td>0.0</td>
    </tr>
  </tbody>
</table>


Frankly, I'm quite surprised the numba strategy is so poor as things get bigger. it suggests that simply iterating over the points *in python* is faster than doing so in numba if the target is a set of tuples... even when using numba's `prange` to make the operation parallel. This would suggest the `|=` set operator is quite slow in numba.

ljwolf:
That, or, 

```python
at_hash = offsets[hashes == hash_]
```
is quite slow, since you're comparing the full set of coordinates/edges to each unique hash.

ljwolf:
Yeah... it's the equality I think, not the set. You can use a list instead of a set for `edgelist` in `_fast_index_and_flip` and you still have bad performance with large `n_geoms`. 

I just don't see how to do this in numba improves. 

At a high level, the original algorithm:
1. builds a lookup table mapping each point/edge to geometries that have that point/edge on their boundary. This is generally a dictionary with keys as tuples (either `(x,y)` or `((x1,y1),(x2,y2)`) and values as a set of indexes that use that point/edge. 
2. builds a lookup table mapping geometries to other geometries, based on whether they share a point/edge. This is generally keyed with indices (`int`) for each geometry with values as the collection of other geoms that share a point/edge (`[int, int, ...]` or `{int, int, ...}`)

This works great in pure python since dicts are fast and everything is hashable. Further, converting the last step to *append to a list* or *add to a set* the tuple `(I,j)`, rather than add `j` to `dict[I]`, works great, too. 

But, in numba, you [can't have a mutable container as the values of a dict](https://numba.pydata.org/numba-doc/dev/reference/pysupported.html?highlight=dict#typed-dict). So, that means no `{int : list}` or `{int: set}` dicts.
We could try something devious, doing `{int: str}` and using a `;` to separate ids? but that'd be hacky. 

To compensate, the implementation instead does:
1. get the unique hashes using `numpy.unique()`
2. for unique hashes, find indices that have that hash (the offending abovementioned `indices[hashes == hash_]`)
3. for each pair `i,j` of indices at the hash, update some output (either a list or a set) with `(I,j)` and `(j,i)`

libpysal/graph/_vertex_set_contiguity.py
```diff
@@ -0,0 +1,159 @@
...
+            raise Exception("Weight type {} Not Understood!".format(self.wttype))
+        if not self.numba:
+
+            hash_to_geom_lut = collections.defaultdict(set)
+            for i, hash_ in enumerate(hashes):
+                hash_to_geom_lut[hash_].add(indices[i])
+
+            edgelist = set()
+            shared_vertices = []
+            for hash_, neighbors in list(hash_to_geom_lut.items()):
```
ljwolf, line 70: I tried parallelising this using joblib+numba. I had a function that did the inner loop, mapped this over all available cores, and then united the results. 

That was nearly an order of magnitude slower.

ljwolf:
OK, final timings are here:
<table border="1" class="dataframe">
  <thead>
    <tr>
      <th>dataset</th>
      <th colspan="2" halign="left">edgelist_pyhash</th>
      <th colspan="2" halign="left">dict</th>
      <th colspan="2" halign="left">correct</th>
      <th colspan="2" halign="left">speedup</th>
      <th>n_coords</th>
      <th>n_geoms</th>
    </tr>
    <tr>
      <th></th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th>mean</th>
      <th>std</th>
      <th></th>
      <th></th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>columbus.shp</th>
      <td>0.000882</td>
      <td>0.000050</td>
      <td>0.003884</td>
      <td>0.000062</td>
      <td>True</td>
      <td>0.0</td>
      <td>3.415286</td>
      <td>0.200921</td>
      <td>1196</td>
      <td>49</td>
    </tr>
    <tr>
      <th>south.shp</th>
      <td>0.069331</td>
      <td>0.023092</td>
      <td>0.156798</td>
      <td>0.023751</td>
      <td>True</td>
      <td>0.0</td>
      <td>1.797178</td>
      <td>1.631756</td>
      <td>41126</td>
      <td>1412</td>
    </tr>
    <tr>
      <th>NAT.shp</th>
      <td>0.119465</td>
      <td>0.022743</td>
      <td>0.346447</td>
      <td>0.036635</td>
      <td>True</td>
      <td>0.0</td>
      <td>2.095172</td>
      <td>1.053413</td>
      <td>80563</td>
      <td>3085</td>
    </tr>
    <tr>
      <th>Chicago_parcels.shp</th>
      <td>5.251881</td>
      <td>0.207851</td>
      <td>24.208030</td>
      <td>0.309266</td>
      <td>True</td>
      <td>0.0</td>
      <td>3.617852</td>
      <td>0.221533</td>
      <td>3250087</td>
      <td>592521</td>
    </tr>
  </tbody>
</table>

Just moving to this edgelist as the "final" data structure (which is what `scipy.sparse` consumes by default and what we can use to build `xarray` weights), we get speedups on the order of 2x to 3x.

libpysal/graph/_vertex_set_contiguity.py
```diff
@@ -0,0 +1,102 @@
...
+        else:
+            raise Exception("Weight type {} Not Understood!".format(self.wttype))
+
+        hash_to_geom_lut = collections.defaultdict(set)
+        for i, hash_ in enumerate(hashes):
+            hash_to_geom_lut[hash_].add(indices[i])
+
+        edgelist = set()
+        shared_vertices = []
+        for hash_, neighbors in list(hash_to_geom_lut.items()):
```
ljwolf, line 56: doesn't need the "list()" around items.

libpysal/graph/wdf.py
```diff
@@ -2,9 +2,10 @@
 from scipy import sparse
 import numpy
 from sklearn.preprocessing import LabelEncoder
+from ._vertex_set_contiguity import ContiguityWeightsLists as _Vertex_Contiguity
 
 
-class W_DF:
+class W:
```
jGaboardi, line 8: During the sprint it was mentioned that we might want to move away from the *W* terminology for notating spatial weights. Not that it should be changed here, I am just curious if that is something still on the table for the future?

libpysal/graph/_vertex_set_contiguity.py
```diff
@@ -0,0 +1,102 @@
...
+            on_point = 0
+            continue
+        else:
+            head = coordinates[i + 1]
+
+        # ensure left-to-right ordering to avoid reversability issues
+        if head[0] > tail[0]:
+            tmp = head
+            head = tail
+            tail = tmp
```
martinfleis, line 98:
```suggestion
            head, tail = tail, head
```
bit shorter version of the same

martinfleis:
On a GeoDataFrame with 140k dense tessellation polygons (~200 points per polygon), I see around 3x speedup.

libpysal/graph/wdf.py
ljwolf: Yeah, I'm up for calling it a GeoGraph!

martinfleis:
@ljwolf I guess you already tried but can you somehow clean this PR so it is actually reviewable? At this point it may even be easier to get a clean branch :D.

ljwolf:
Yeah, I think the rebase pulled in a bunch of erroneous material from other commits. I'll revise and reopen.

Issue #282, opened 2020-04-27
Return, don't print, available examples. 


ljwolf:
As a general rule, we should avoid print in final user-facing code. 

It's normally used in 3 ways I think:
1. **Indicate the current interpretation of inputs or current status a function.** This is normally better served with (i) a warning if something is provided that is not as expected (more explicit, highlighted in the console/notebook, can be filtered when running in batch, contains information about where/when the warning is raised) (ii) using the [builtin `logging` module](https://docs.python.org/3/library/logging.html) for status messages in long-running programs (all the benefits of warning, plus that it can be saved if the session dies), or (iii) returning well-defined error-state values (numpy.nan, None, empty array/list). 
2. **Display values at the end of a computation that are not themselves of interest.** This is nearly always better as an explicit return because the user can build further on top of the function, kind of like how we should avoid `plt.show()`ing plots within functions since it precludes users building further upon them. 

So, here, 

```python
import libpysal
libpysal.examples.available()
```
if I wanted to figure out which were installed, I have to look through the return set. I can't sort, nor can I filter. But, if I got the dataframe back, not only would it be pretty-printed in the notebook, but it'd also be able to be filtered:
```python
datasets = libpysal.examples.available()
datasets.query('installed') # shows only installed datasets
datasets.Description.apply(lambda x: 'Chicago' in x) # give datasets with chicago in desc
```

jGaboardi:
resolved in #411

PR #411, opened 2021-06-14
Return a dataframe with info on available datasets

sjsrey:
Addresses #282 

Return instead of print.


jGaboardi (approved):

martinfleis (commented):
Can we add a tiny test checking that we actually return df?

PR #427, opened 2021-09-01
Graph module based on numba constructors

ljwolf:
This re-opens #402


libpysal/graph/wdf.py
```diff
@@ -0,0 +1,70 @@
+from xarray import DataArray
+from scipy import sparse
+import numpy
+from sklearn.preprocessing import LabelEncoder
+from ._vertex_set_contiguity import ContiguityGrapheightsLists as _Vertex_Contiguity
```
martinfleis, line 5:
```suggestion
from ._vertex_set_contiguity import ContiguityWeightsLists as _Vertex_Contiguity
```

libpysal/graph/_vertex_set_contiguity.py
```diff
@@ -0,0 +1,100 @@
...
+        self.wttype = wttype
+        self.contiguity()
+
+    def contiguity(self):
+
+        if self.wttype == QUEEN:  # points are hashed
+            hashes = _point_hash(self.coordinates)
+            indices = self.coord_indices
+        elif self.wttype == ROOK:  # edges are hashed
+            hashes = _edge_hash(self.coordinates)
```
martinfleis, line 45: `_edge_hash` requires `coordinates, indices, n_coordinates_per_geom`

(sorry for piecemeal suggestions, I am just playing with this more than doing a review)

ljwolf: very fine 😄 I do like keeping things WIP for these little bits of review... I think it can be a very fruitful use of small spare times.

ljwolf:
I think this should be closed!

PR #426, opened 2021-08-30
cast sw to csc and eliminate zeros

ljwolf:
Hello! Please make sure to check all these boxes before submitting a Pull Request
(PR). Once you have checked the boxes, feel free to remove all text except the
justification in point 5. 

1. [x] You have run tests on this submission locally using `pytest` on your changes. Continuous integration will be run on all PRs with [GitHub Actions](https://github.com/pysal/libpysal/blob/master/.github/workflows/unittests.yml), but it is good practice to test changes locally prior to a making a PR.
2. [x] This pull request is directed to the `pysal/master` branch.
3. [x] This pull introduces new functionality covered by
   [docstrings](https://en.wikipedia.org/wiki/Docstring#Python) and
   [unittests](https://docs.python.org/2/library/unittest.html)? 
4. [x] You have [assigned a
   reviewer](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) and added relevant [labels](https://help.github.com/articles/applying-labels-to-issues-and-pull-requests/)
5. [x] The justification for this PR is: 

Bug in `lat2SW` that does not remove structural zeros before returning the sparse matrix.

PR #428, opened 2021-09-01
Fix failure of "dia_matrix object is not subscriptable"

keszybz:
This occurs when running the code with numba unavailable.


keszybz:
#426 does this better, let's close this.

ljwolf:
Thank you for the contribution nonetheless! Sorry we were delayed in addressing it.

PR #430, opened 2021-09-05
Fix test_voronoi on other architectures.

QuLogic:
Floating point results shouldn't be checked exactly, as they might differ on other machines. This test currently fails on at least s390x and aarch64.


martinfleis (approved):
Thanks!

jGaboardi (approved):

PR #436, opened 2021-10-01
Update of the min_threshold_dist_from_shapefile function to support geopandas objects directly.

PhilipeRLeal:
5. [x] The justification for this PR is: The min_threshold_dist_from_shapefile function (from libpysal.utils) was updated so to support a geopandas GeoDataFrame or a geopandas.GeoSeries object directly. This Pull Request's update also ensured back-compatibility with the min_threshold_dist_from_shapefile elder behavior - of dealing with strings pointing towards shapefile filepaths.

## Implications:

None

### DIFF
--- a/libpysal/weights/util.py
+++ b/libpysal/weights/util.py
@@ -1082,8 +1082,9 @@ def get_points_array_from_shapefile(shapefile):
 
     Parameters
     ----------
-    shapefile     : string
-                    name of a shape file including suffix
+    shapefile     : string/geopandas.GeoSeries/geopandas.GeoDataFrame
+                    If a string, it must be a full pathname of 
+                    a shape file - including suffix
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


jGaboardi (approved):

libpysal/weights/util.py
```diff
@@ -1117,8 +1118,12 @@ def get_points_array_from_shapefile(shapefile):
            [ 8.33265837, 14.03162401],
            [ 9.01226541, 13.81971908]])
     """
-
-    f = psopen(shapefile)
+    if isinstance(shapefile, str):
+        f = psopen(shapefile)
+    elif isinstance(shapefile, gpd.GeoDataFrame):
+        f = shapefile['geometry']
```
martinfleis, line 1124:
```suggestion
        f = shapefile.geometry
```

GeoDataFrame can have an active geometry column named differently than `'geometry'`, we always need to access it via an attribute.

martinfleis (commented):
I have very mixed feelings here.

The change in this PR is perfectly fine code-wise (just see the note in the code).

But API-wise, this is a bit messy but that is more for a general discussion among libpysal team. `*from_shapefile` should be a different thing than `*from_dataframe`. In the ideal case, we wouldn't need any `*from_shapefile` or `*from_dataframe`. We should assume that the input is GeoDataFrame and leave the file IO to geopandas. 

There is a lot of legacy code in libpysal that should be cleaned and refactored based on geopandas and pygeos. Short term, we may just merge this assuming that these functions will eventually be deprecated anyway. But I would surely not go in the direction of shapefile==GeoDataFrame within our API.

PhilipeRLeal:
Dear @martinfleis,

thank's for the feedback. Your suggestion was completely right. The suggestion was accepted in this new commit.

SIncerely,

PR #437, opened 2021-10-06
REF: update for shapely 2.0

martinfleis:
This PR adapts the code to the upcoming shapely 2.0, which in practice means eliminating deprecation warnings coming from shapely 1.8rc1.

See the migration guide for details - https://shapely.readthedocs.io/en/latest/migration.html. My guess is that we will have to do similar actions in other repositories as well.

You can install shapely 1.8rc1 from conda-forge:

```
mamba update -c conda-forge/label/shapely_dev shapely
```


martinfleis:
> deprecate things like geotable, shapely_extension, and cg.shapes!

Yes please!

PR #438, opened 2021-10-18
add gabriel weights and more performant voronoi generation

ljwolf:
This adds a revised `numba`-accelerated version of the Voronoi constructor (which avoids building the Voronoi polygons explicitly and instead opts to work directly with the Delaunay graph) and implements a Gabriel graph constructor (*edit: and a relative neighborhood constructor*) based on the new Voronoi/Delaunay constructor. The Gabriel graph is a subset of the Delaunay triangulation where the circle whose diameter is edge *ij* contains no other point *k*.

### The math for the Gabriel graph

Given the fast `QHull` Delaunay constructor, we can just start from the Delaunay triangulation and remove links *ij* if *k*, some other neighbor of *i*, is within the circle whose diameter is *ij*. 

To do this, we first can say (by [Thales' Theorem](https://en.wikipedia.org/wiki/Thales%27s_theorem)) that any triangle *ijk* with a side *ij* that is a circle diameter will be a *right* triangle when the non-diameter point *k* is *also* on the circle. This means that when *k* is *on the boundary* of the circle with *ij* as a diameter, then we have the Pythagorean relationship, *dij^2 = dik^2 + djk^2*. However, if *dij^2* is ever bigger than *dik^2 + djk^2*, then we know that triangle *ijk* is obtuse, and that *k* is "within" the circle whose diameter is *ij*. So, for focal *i* and each pair of neighbors *(j*,*k)*, prune Delaunay edge *ij* when *dij^2 > dik^2 + djk^2*. 

This needs:
- [ ] tests
- [x] comparison with R's [gabrielgraph](https://r-spatial.github.io/spdep/reference/graphneigh.html)
- [x] **decision**: if we use this graph constructor for Delaunay/Voronoi, it comes as about an order of magnitude speedup. But, it (a) cannot handle "clipping" the voronoi polygons to an arbitrary study extent (or, indeed, to a convex or alpha-hull) and (b) will then never be exactly equivalent to the Rook graph of a clipped Voronoi Diagram. But, this has the conceptual simplicity that it *only* converts the Delaunay triangulation into a graph, rather than building the voronoi cells, then clipping them, then constructing that Rook graph. So, input is needed on which we should use @sjsrey @martinfleis 
```
        Old Voronoi: 0.83s
        Raw Gabriel: 1.37s
        Compile: 1.88s
        New Delaunay: 0.04s
        Ref Gabriel: 0.07s
```

### DIFF
--- a/libpysal/weights/gabriel.py
+++ b/libpysal/weights/gabriel.py
@@ -167,7 +167,7 @@ class Relative_Neighborhood(Delaunay):
     **kwargs    :   keyword argument list
         keyword arguments passed directly to weights.W
     """
-    def __init__(self, coordinates, **kwargs):
+    def __init__(self, coordinates, binary=True, **kwargs):
         try:
             from numba import njit
         except ModuleNotFoundError:
@@ -181,6 +181,8 @@ class Relative_Neighborhood(Delaunay):
             edges, dt.points, return_dkmax=False
         )
         row, col, data = zip(*output)
+        if binary:
+            data = numpy.ones_like(col, dtype=float)
         sp = sparse.csc_matrix((data, (row, col))) #TODO: faster way than this?
         tmp = WSP(sp).to_W() 
         W.__init__(self, tmp.neighbors, tmp.weights, **kwargs)
--- a/libpysal/weights/gabriel.py
+++ b/libpysal/weights/gabriel.py
@@ -1,8 +1,13 @@
 from scipy.spatial import Delaunay as _Delaunay
 from scipy import sparse
-from numba import njit
 from libpysal.weights import W, WSP
-import joblib, pandas, pygeos, numpy
+import pandas, numpy
+
+try:
+    from numba import njit
+except ModuleNotFoundErrror:
+    from libpysal.common import jit as njit
+
 
 # delaunay graphs and their subgraphs
 
@@ -28,7 +33,9 @@ class Delaunay(W):
 
     @classmethod
     def from_dataframe(cls, df, **kwargs):
-        return cls(pygeos.get_coordinates(pygeos.centroid(df.geometry.values.data)))
+        centroids = df.geometry.centroid
+        point_array = numpy.column_stack((centroids.x.values, centroids.y.values))
+        return cls(point_array)
 
 
 class Gabriel(Delaunay):
@@ -153,22 +160,6 @@ def _filter_relativehood(edges, coordinates, return_dkmax=False):
             r.append(dkmax)
 
     return out, r
-
-
-class Mutual_Reachability(W):
-    def __init__(
-        self, coordinates, n_max=None, n_core=5, method="kdtree", metric="euclidean"
-    ):
-        # check to see if metric is supported by method using tree.valid_metrics
-        # so, kdtree can only do minkowski
-        # balltree can handle more
-        # approx can handle a ton, but wait to import pynndescent unless method = approx
-        # raise an error if the requested metric is not supported
-
-        # then, compute the core distances for each point
-        # finally, for all pairs of points, compute the distances for the nearest n_max
-        # points (possibly all points), and censor that by the core distance.
-        raise NotImplementedError()
 
 
 if __name__ == "__main__":
--- a/libpysal/weights/gabriel.py
+++ b/libpysal/weights/gabriel.py
@@ -43,6 +43,26 @@ class Gabriel(Delaunay):
         W.__init__(self, gabriel_neighbors, **kwargs)
 
 
+class Relative_Neighborhood(Delaunay):
+    def __init__(self, coordinates, return_dkmax=False, **kwargs):
+        """
+        via toussaint, https://doi.org/10.1016/0031-3203(80)90066-7
+        """
+        edges, dt = self._voronoi_edges(coordinates)
+        output, dkmax = _filter_relativehood(
+            edges, dt.points, return_dkmax=return_dkmax
+        )
+        if not return_dkmax:
+            del dkmax
+        row, col, data = zip(*output)
+        sp = sparse.csc_matrix((data, (row, col))) #TODO: faster way than this?
+        tmp = WSP(sp).to_W() 
+        W.__init__(self, tmp.neighbors, tmp.weights, **kwargs)
+
+
+
+#### utilities
+
 @njit
 def _edges_from_simplices(simplices):
     edges = []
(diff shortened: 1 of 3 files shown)


ljwolf:
Results replicate with `spdep::gabrielgraph`!

martinfleis:
> cannot handle "clipping" the voronoi polygons to an arbitrary study extent

I had the same issue with scipy.spatial.Voronoi in momepy and went around it by adding external points to the input. `momepy.Tessellation`, which is in its purest form a Voronoi tessellation based on points generated along the polygon boundary, requires a `limit` polygon, specifying the external limit of tessellation (see http://docs.momepy.org/en/stable/user_guide/elements/tessellation.html). We then take a convex hull of the limit, buffer it by twice the size of bounds and add points along the boundary of this buffer to the mix along those coming from buildings. That way we ensure that all tessellation cells always reach the limit unaffected by these external points and none of them ends in infinity as cells leading to infinity are only those based on the buffered hull. Those are dropped anyway (and remaining are clipped to `limit`).

ljwolf:
> > cannot handle "clipping" the voronoi polygons to an arbitrary study extent
> 
> I had the same issue with scipy.spatial.Voronoi in momepy and went around it ...

yeah, in [the current implementation](https://github.com/pysal/libpysal/blob/5b386cfd182f131c174ef10f789e9c49acc7fd4d/libpysal/weights/contiguity.py#L480), we do something similar, allowing an automated clip to the bounding box, the convex hull, or the concave hull... But, this requires us to build the faces as polygons, which is generally slower than iterating over the tessellation in `numba`... 

But, there are quite a few cases where a clipped Voronoi diagram is not the same as a non-clipped one, thus the "Delaunay-like" triangulation implied by `Voronoi()` is not actually the same as the Rook graph of the clipped voronoi polygons. 

I think what we may do is:
1. keep the `Voronoi` around,
2. put it in the same module as this Delaunay & Mutual Neighbor stuff
3. write a vignette explaining why these aren't the same, and refer users to that vignette for the difference.

martinfleis:
Ah, sure, that won't help Delaunay.

Since we do `voronoi_frames` + `Queen` under the hood now, I think we can deprecate `weights.Voronoi`. Users can do it themselves if they need a clipped version.

I am just not sure if I would use Delaunay constructor myself, knowing that it can result in some super weird neighbours, like two opposite corners of an array of points. Everything on the edges will be a bit sketchy.

ljwolf:
> I am just not sure if I would use `Delaunay` constructor myself, knowing that it can result in some super weird neighbors

Yeah, definitely. But, my worry is that, in general, we can't get the gabriel graph from the clipped voronoi queen graph, since it's only a subset of the delaunay graph. 

Mathematically, I think there's a relationship between our "concave hull" alpha shape, the delaunay graph, and the gabriel graph, but I'm not sure how to characterize it off the top of my head. I think we need a proof that the radius defining the limiting concave hull is always larger than the largest gabriel neighbor radius. 

So, I put together a [small gist](https://gist.github.com/ljwolf/d7be4aeed8b0aba79b7dc9c1ec905035) to look into it. In this example, I simulate a banana-shaped pattern (where clipped Voronoi will be very different from the Delaunay triangulation), and check whether the Gabriel graph is a subgraph of the Queen Alpha-Hull clipped Voronoi.

download

When I simulate 1000 random patterns with this shape, the gabriel is always a subset of the clipped Queen graph.

ljwolf:
I should be clear: I'm in favor of keeping the Voronoi constructor around, but maybe "hiding" the raw delaunay one, or making `libpysal.weights.Delaunay` another name for `libpysal.weights.Voronoi` with `clip=None` as default.

martinfleis:
Ahh, I should have paid more attention to the Gabriel graph. So the idea is to use `weights.Gabriel` where you now use clipped `weights.Voronoi`? That sounds good to me.

I wouldn't even hide `Delanuay`, just properly document it so users are aware of potential consequences.

ljwolf:
This has tests and is passing. It only needs docstrings now and maybe a numba warning if computed w/o acceleration

libpysal/weights/gabriel.py
```diff
@@ -0,0 +1,265 @@
+from scipy.spatial import Delaunay as _Delaunay
+from scipy import sparse
+from libpysal.weights import W, WSP
+import pandas, numpy
+
+try:
+    from numba import njit
+except ModuleNotFoundErrror:
+    from libpysal.common import jit as njit
```
martinfleis, line 9: Can we add that warning here? I think it is useful to give users indication that it can be faster.

libpysal/weights/gabriel.py
```diff
@@ -0,0 +1,265 @@
...
+            pandas.DataFrame(numpy.asarray(list(edges)))
+            .sort_values([0, 1])
+            .drop_duplicates()
+            .values
+        )
+        return edges, dt
+
+    @classmethod
+    def from_dataframe(cls, df, **kwargs):
+        centroids = df.geometry.centroid
```
martinfleis, line 36: Wouldn't be better to check for geometry type and raise if it is not Point here? Rather than silently downcasting all other geometries to a centroid?

I think that the user should be aware of this step and should make a conscious decision on how to do this approximation (centroid, representative_point, a centre of the inscribed circle...). We can suggest it in the error message.

libpysal/weights/gabriel.py
```diff
@@ -0,0 +1,265 @@
...
+    def __init__(self, coordinates, return_dkmax=False, **kwargs):
+        """
+        via toussaint, https://doi.org/10.1016/0031-3203(80)90066-7
+        """
+        edges, dt = self._voronoi_edges(coordinates)
+        output, dkmax = _filter_relativehood(
+            edges, dt.points, return_dkmax=return_dkmax
+        )
+        if not return_dkmax:
+            del dkmax
```
martinfleis, line 63: Even if `return_dkmax=True`, you don't actually return it, do you?

libpysal/weights/gabriel.py
```diff
@@ -0,0 +1,265 @@
...
+        if prune:
+            continue
+        out.append((i, j, dij))
+        if return_dkmax:
+            r.append(dkmax)
+
+    return out, r
+
+
+if __name__ == "__main__":
```
martinfleis, line 165: I am not checking anything below this assuming this will be removed.

libpysal/weights/gabriel.py
ljwolf: I'd rather not add it at import, but I'll add a warning on first use.

libpysal/weights/gabriel.py
ljwolf: Great, yes, will remove. Was just there for development.

libpysal/weights/gabriel.py
ljwolf: Nope; that was another development option I thought of keeping before deciding these should probably follow the same class-oriented design. The `dkmax` array serves as a useful measure of the local density, but this can be gotten directly from the helper function.

libpysal/weights/gabriel.py
ljwolf: Yes, great idea, I think this is good to raise. I think we may do this silently elsewhere in the library (e.g. [KNN](https://github.com/pysal/libpysal/blob/master/libpysal/weights/distance.py#L297))...

ljwolf:
@martinfleis, what would you think about exposing the momepy restricted voronoi tesselation in this PR, too, with a simple try/except wrapper class? 

I think it'd help for discoverability/completeness to expose it at the `weights` location, as we also should with the perimeter-weighted rook...

martinfleis:
> @martinfleis, what would you think about exposing the momepy restricted voronoi tesselation in this PR, too, with a simple try/except wrapper class?

How would you do that? Create a tesselation GeoDataFrame and do Queen/Rook on top of that as we do in existing `Voronoi`? In principle that is why I guess, just people may be surprised that it is way slower than the other methods. And I have to check how does it behave when you pass point geometry type, that is not expected now.

ljwolf:
Yes, that's exactly what I was thinking, under a `Voronoi_Tesselation` class... But, on reflection, I will stub this for a next release.

martinfleis:
Thinking about it a bit more, it may be better to leave `momepy.Tessellation` out of this for now. It currently expects projected buildings on input (in meters) and we may want to try to generalise it a bit before plugging it in libpysal.

ljwolf:
Yep, I totally agree. I think it could be generalized, but I don't want that to get in the way of releasing this... so merge here & then I'll cut releases of this + esda this weekend!

libpysal/weights/gabriel.py
```diff
@@ -0,0 +1,280 @@
...
+    coordinates :   array of points, (N,2)
+        numpy array of coordinates containing locations to compute the
+        delaunay triangulation
+    **kwargs    :   keyword argument list
+        keyword arguments passed directly to weights.W
+    
+    Notes
+    -----
+    The Delaunay triangulation can result in quite a few non-local links among
+    spatial coordinates. For a more useful graph, consider the weights.Voronoi
```
martinfleis, line 37: Do you also want to point to Gabriel? Maybe using See also?

libpysal/weights/gabriel.py
```diff
@@ -0,0 +1,280 @@
...
+        edges = (
+            pandas.DataFrame(numpy.asarray(list(edges)))
+            .sort_values([0, 1])
+            .drop_duplicates()
+            .values
+        )
+        return edges, dt
+
+    @classmethod
+    def from_dataframe(cls, df, **kwargs):
```
martinfleis, line 70: Can we add a short docstring here mentioning that geometry must be Points?

martinfleis (commented):
Thanks for the changes! Can you also add a test for the `Relative_Neighborhood`?

libpysal/weights/gabriel.py
ljwolf: Yep, done!

libpysal/weights/gabriel.py
ljwolf: Yes, mentioned in the docstring as an alternative to the clipped Voronoi

ljwolf:
OK, all should be addressed plus a new test for the Relative Neighborhood Graph. I've also forced that to use binary weights (instead of distances) so that it behaves like the rest of the graphs in the module.

martinfleis (approved):
Thanks! I think it is ready to go.

martinfleis:
One more thing - we need to add it to docs (`docs/api.rst`).

sjsrey (approved):

PR #439, opened 2021-10-23
Bug fix for parallel execution environments

schmidtc:
1. [No] Don't have dev environment setup, but tested functionality locally.
5. [Yes] The justification for this PR is: Have been hitting a rare bug here when deploying applications with many worker processes in a Kubernetes environment.  A bit of a race condition exists here where multiple workers attempt to create the directory but fail because another worker already created it, resulting in a crash loop.


ljwolf:
Thanks @schmidtc! We'll be shooting for a point release of this within a week 😄

PR #449, opened 2022-01-22
DOC: Update docs for Gabriel and related enhancements

sjsrey:


docs/conf.py
```diff
@@ -124,7 +124,7 @@
 #
 html_theme_options = {
     # Navigation bar title. (Default: ``project`` value)
-    "navbar_title": project,
```
jGaboardi, line 127: What's reasoning for replacing `project` with `"libpysal"`? `project = "libpysal"` is defined in line 73 so that it can be used throughout `conf.py`.

docs/conf.py
```diff
@@ -287,47 +287,3 @@ def setup(app):
     "python": ("https://docs.python.org/3.8/", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/reference/", None),
 }
-
-
-# This is processed by Jinja2 and inserted before each notebook
```
jGaboardi, line 292: Why are we getting rid of the notebook rendering for tutorials?

sjsrey: I see what happened. I had moved the conf.py over from the old docsrc which predated these changes. Will revert, update and resubmit this.

sjsrey:
will fix and resubmit

PR #442, opened 2021-12-16
ENH: Moving to appdirs for example storage

sjsrey:
Using [appdirs](https://github.com/ActiveState/appdirs) to handle local storage of examples.

PR #451, opened 2022-01-23
Update testing, add Py310, etc.

jGaboardi:
This PR:
* removes unneeded `ci/travis/*` files (see [here](https://github.com/pysal/libpysal/pull/442#issuecomment-1019323303))
* adds in `pytest-xdist` for multicore testing
* adds in testing against Python `3.10`
* trims down testing matrix (`3.{7,8,9,10}` on Ubuntu; `3.10` on macOS, Windows)
* removes mention of support for Python `3.6`
* etc.


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos: 
--   repo: https://github.com/ambv/black
-    rev: stable
-    hooks: 
+-   repo: https://github.com/psf/black
+    rev: 20.8b1
```
martinfleis, line 3: Why this specific pin? It is quite old.

sjsrey (approved):

.pre-commit-config.yaml
jGaboardi: Good point. The specific version is because [`stable` is not recommended anymore](https://github.com/ljvmiranda921/comments.ljvmiranda921.github.io/issues/18#issuecomment-870697993), but that version is old as you say and should be updated with `pre-commit autoupdate`. I'll push up that fix.

martinfleis (approved):

PR #454, opened 2022-01-24
revert shapely hard dependecy

### LINKED ISSUE #453: Shapely Dependancy Added
[This commit](https://github.com/pysal/libpysal/commit/cd4b809de38ea0df1c4d2564b72617fcd6f81c4b) adds a dependancy on shapely.

martinfleis:
Closes #453 

Since this a regression of some sort, I'd recommend issuing 4.6.1 release.

I have also included a CI environment that does not have any optional dependencies so we can catch this issue next time before merging and releasing.

### DIFF
--- a/libpysal/cg/tests/test_voronoi.py
+++ b/libpysal/cg/tests/test_voronoi.py
@@ -5,6 +5,14 @@ import numpy as np
 from ..voronoi import voronoi, voronoi_frames
 from ..shapes import Polygon, asShape
 
+import pytest
+
+try:
+    import shapely
+    HAS_SHAPELY = True
+except ImportError:
+    HAS_SHAPELY = False
+
 
 class Voronoi(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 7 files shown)


martinfleis:
I also had to do some adaptation in tests to make sure they're properly skipped if the required optional dep is missing.

libpysal/cg/tests/test_voronoi.py
```diff
@@ -29,6 +37,7 @@ def test_voronoi(self):
 
         np.testing.assert_array_almost_equal(vertices, self.vertices)
 
+    @pytest.mark.skipif(not HAS_SHAPELY, reason="shapely needed")
```
jGaboardi, line 40: would the [`@requires` decorator](https://github.com/pysal/libpysal/blob/8aa08b1c58d6622eddb39885d78c34d75f9d27b6/libpysal/common.py#L115) that is defined in `common.py` work here?

```python
@requires("shapely")
def test_voronoi_frames(self):
    ....

```

jGaboardi (approved):

libpysal/cg/tests/test_voronoi.py
martinfleis: Not here, that would work if the normal not the test function required shapely. I can't even use it in `get_points_array` because there's a no-shapely path in there as well.

jGaboardi:
Shall we merge or get another review from @sjsrey?

martinfleis:
> I should have also checked this before the tag for v4.6.1. Happy to make 4.6.2 ?

I forgot that it is still there. Yes, I would do 4.6.2 with this.

ljwolf:
on it

Issue #445, opened 2022-01-19
Add support for `DataArray` objects read through `rioxarray`


darribas:
The `raster` functionality in for weights building currently supports `DataArray` objects created through `xarray.open_rasterio`:

```python
import xarray
from libpysal.weights import Queen
pop = xarray.open_rasterio(
               'https://geographicdata.science/book/_downloads/5263090bd0bdbd7d1635505ff7d36d04/ghsl_sao_paulo.tif'
)
w = Queen.from_xarray(pop)
```

The code above correctly picks up that missing data are specified with -200 and removes them from the calculation of the weights:

```python
w.n
> 97232
```

This is not the case when we use `DataArray` objects build through the (generally recommended) `rioxarray.open_rasterio`:

```python
import rioxarray
pop2  = rioxarray.open_rasterio(
               'https://geographicdata.science/book/_downloads/5263090bd0bdbd7d1635505ff7d36d04/ghsl_sao_paulo.tif'
)
w2 = Queen.from_xarray(pop2)
```

which does _not_ pick up the missing data:

```python
w.n
> 194688
```

My hunch is that this is because the builder picks up missing values _only_ through the `attrs` object:

https://github.com/pysal/libpysal/blob/0f03a313896de6af096ff92105e54da86bb78368/libpysal/weights/raster.py#L229-L233

But `rioxarray` does not populate that general object, instead logs missing data under `pop2.rio.nodata`:

```python
pop.attrs
> {'transform': (250.0, 0.0, -4482000.0, 0.0, -250.0, -2822000.0),
 'crs': '+proj=moll +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs=True',
 'res': (250.0, 250.0),
 'is_tiled': 0,
 'nodatavals': (-200.0,),
 'scales': (1.0,),
 'offsets': (0.0,),
 'AREA_OR_POINT': 'Area',
 'grid_mapping': 'spatial_ref'}

pop2.attrs
> {'_FillValue': -200.0, 'scale_factor': 1.0, 'add_offset': 0.0}

pop2.rio.nodata
> -200.0
```

I think we should support both formats, read through vanilla `xarray` and with `rioxarray` loaded in the session too. The former is a more generic approach, the latter is a more robust one for rasters.

[...]

PR #455, opened 2022-01-29
Fix#445

darribas:
justification in point 5. 

4. [x] You have [assigned a
5. [x] The justification for this PR is: Fix#445

### DIFF
--- a/libpysal/weights/raster.py
+++ b/libpysal/weights/raster.py
@@ -173,7 +173,7 @@ def nodata_from_attrs(attrs):
             returned
     """
     candidates = [
-            '_FillValue', 'missing_value', 'fill_value', 'nodata'
+            '_FillValue', 'missing_value', 'fill_value', 'nodata', 'nodatavals'
     ]
     for i in candidates:
         if i in attrs:
--- a/libpysal/weights/raster.py
+++ b/libpysal/weights/raster.py
@@ -149,16 +149,39 @@ def get_nodata(da):
     try:
         return da.rio.nodata
     except:
-        candidates = [
-                '_FillValue', 'missing_value', 'fill_value', 'nodata'
-        ]
-        for i in candidates:
-            if i in da.attrs:
-                if type(i) == tuple:
-                    return da.attrs[i][0]
-                else:
-                    return da.attrs[i]
+        nodata = nodata_from_attrs(da.attrs)
+        if nodata is not None:
+            return nodata
         return None
+
+def nodata_from_attrs(attrs):
+    """
+    Identify nodata value in a `DataArray.attrs`
+
+    NOTE: follows guidance from https://corteva.github.io/rioxarray/stable/getting_started/nodata_management.html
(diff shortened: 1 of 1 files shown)


libpysal/weights/raster.py
```diff
@@ -128,6 +128,60 @@ def da2W(
...
+            Value used for nodata pixels. If no value is available, `None` is
+            returned
+    """
+    try:
+        return da.rio.nodata
+    except:
+        nodata = nodata_from_attrs(da.attrs)
+        if nodata is not None:
+            return nodata
+        return None
```
martinfleis, line 155: This is redundant, isn't? 
```suggestion
        return nodata_from_attrs(da.attrs)
```

If it is `None` it returns `None`, if not it returns a value.

martinfleis (commented):
Can you try to increase test coverage to cover more of these options? Codecov reports some of these lines as unchecked currently.

libpysal/weights/raster.py
```diff
@@ -128,6 +128,60 @@ def da2W(
...
+
+    Returns
+    -------
+    nodata : int/float
+            Value used for nodata pixels. If no value is available, `None` is
+            returned
+    """
+    try:
+        return da.rio.nodata
+    except:
```
snowman2, line 151: Hopefully this is helpful: https://www.flake8rules.com/rules/E722.html

darribas: Thanks very much! This got me thinking a more specific way could be to check for `rio` by:

```python
if hasattr(da, 'rio'):
    return da.rio.nodata
else:
```

Would that be preferable? Otherwise I can catch the exception as in the link shared.

Thanks very much for the code review and my non-developer traits ;-)

snowman2: I think both `hasattr` and catching `AttributeError` are both valid options in this scenario. The benefit of `hasattr` is that you can easily check for a specific attribute.

Side note: IMO, the best developer trait is the willingness to learn and improve because we are all learning :-)

darribas: Added!

ljwolf:
What does this need to merge?

martinfleis:
We should test the use case from #445 this is supposed to fix. It is not being covered by current tests.

martinfleis (approved):
I updated this and added a simple test taken from the issue. I also verified that the builder in `graph` behaves as expected and is not affected by this issue (though we may want to revise #741). lgtm

libpysal/weights/raster.py
```diff
@@ -133,6 +133,58 @@ def da2W(
...
 
 
+def get_nodata(da):
+    """
+    Identify nodata value in a `DataArray`
+
+    NOTE: follows guidance from https://corteva.github.io/rioxarray/stable/getting_started/nodata_management.html
+    ...
+
+    Arguments
```
jGaboardi:
```suggestion
    Parameters
```

Issue #456, opened 2022-01-30
alpha shape multipolygon - is it expected behavior to have filled in holes?


cornhundred:
Hi, I would like to know if it is expected behavior for the alpha shape to be returned with filled in holes? This example [colab notebook](https://colab.research.google.com/drive/1KCCnbwTvg-omi8Fz9UHq0bn1NSfTPl90?usp=sharing) illustrates this

```
from libpysal.cg import alpha_shape, alpha_shape_auto
from shapely.geometry import MultiPolygon, Polygon
import numpy as np
import pandas as pd

# generate random matrix
num_cols = 2
num_rows = 1000
np.random.seed(seed=100)
pts = np.random.rand(num_rows, num_cols) * 100

inv_alpha = 3.5
poly = alpha_shape(pts, 1/inv_alpha)

multi_poly = MultiPolygon(poly.values)
multi_poly
```

<img width="369" alt="Screen Shot 2022-01-29 at 11 38 23 PM" src="https://user-images.githubusercontent.com/8352840/151686839-95716899-76a7-46f7-a5ff-e56b8bcb1d97.png">

The alpha shape polygon geoseries appears to include polygons that intersect the holes in the alpha shape. Is this expected and if so can we easily remove these in order to combine these polygons into a MultiPolygon?

- Platform information:
```python
>>> import os; print(os.name, os.sys.platform);print(os.uname())

posix linux
posix.uname_result(sysname='Linux', nodename='b49e8e01b8be', release='5.4.144+', version='#1 SMP Tue Dec 7 09:58:10 PST 2021', machine='x86_64')
```
- Python version: 
```python
>>> import sys; print(sys.version)

3.7.12 (default, Sep 10 2021, 00:21:48) 
[GCC 7.5.0]
```
- SciPy version:
```python
>>> import scipy; print(scipy.__version__)

1.4.1
```
- NumPy version:
```python
>>> import numpy; print(numpy.__version__)

1.19.5
```

ljwolf:
Hey, yes the "extra" polygons are holes:

![image](https://user-images.githubusercontent.com/2250995/151958152-6e461369-1d4f-40d3-89fd-8c203d692491.png)

This results in some confusing behavior if you assemble a multipolygon using all the alpha shapes as regular polygons... some of the polygons need to be treated as holes, and others as rings.

[...]

PR #457, opened 2022-02-01
[WIP] Filter holes from alpha shape returns

ljwolf:
Noted by @cornhundred in #456, we've got some surprising behavior for alpha shapes with holes. This attempts to fix this in the `alpha_shape()` function by (1) checking if there are any "holes" in the shapely polygons and, if there are, (2) we only keep polygons that contain an input point. 

This solves the problem OK, but I think there's a bit of an issue: 
Figure_1

Note that this results in two separate polygons-with-holes on the left, but on the right, we retain one of the holes because a point falls within it... I think this happens because the "correct" polygon for that hole would have a zero-area "cut in" that only includes that point. When we use `shapely.ops.polygonize`, we lose that cut in, and thus omit the vertex from the alpha shape. 

Consider this box with a "cut in" from [1,1] to [.5,.5]:

```python
lines = [
         LineString([[0, 0], [0, 1]]), # 1
         LineString([[0, 1], [1, 1]]), # 2
         LineString([[0, 0], [1, 0]]), # 6
         LineString([[1, 0], [1, 1]]), # 3
         LineString([[1, 1], [0.5, 0.5]]), # 4 
         LineString([[0.5, 0.5], [1, 1]]), # 5
        ]
```
upon polygonization, we lose that cut-in:
```python
from shapely.ops import polygonize
geom = polygonize(lines)
geom.wkt
```
`geom` is now `'POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))'` with no mention of the cut-in. 
This is good, since a polygon with this kind of cut-in is an invalid polygon, but it complicates our testing here... not all holes are empty. 

The other way to do this would be to do an all-pairs `within` check for the polygons using `geoms.sindex.query_bulk()`. I was hoping the point-in-poly woudl be faster, but... but you never know. I'll check that & re-submit.

### DIFF
--- a/libpysal/cg/alpha_shapes.py
+++ b/libpysal/cg/alpha_shapes.py
@@ -427,7 +427,7 @@ def alpha_geoms(alpha, triangles, radii, xys):
 
 
 @requires("geopandas", "shapely")
-def alpha_shape(xys, alpha, filter_holes=True):
+def alpha_shape(xys, alpha):
     """Alpha-shape delineation (Edelsbrunner, Kirkpatrick & Seidel, 1983) from a collection of points
 
     Parameters
@@ -483,8 +483,7 @@ def alpha_shape(xys, alpha, filter_holes=True):
     radii = r_circumcircle_triangle(a_pts, b_pts, c_pts)
     del triangles, a_pts, b_pts, c_pts
     geoms = alpha_geoms(alpha, triangulation.simplices, radii, xys)
-    if filter_holes:
-        geoms = _filter_holes(geoms, xys)
+    geoms = _filter_holes(geoms, xys)
     return geoms
 
 
--- a/libpysal/cg/alpha_shapes.py
+++ b/libpysal/cg/alpha_shapes.py
@@ -481,7 +481,7 @@ def alpha_shape(xys, alpha):
     radii = r_circumcircle_triangle(a_pts, b_pts, c_pts)
     del triangles, a_pts, b_pts, c_pts
     geoms = alpha_geoms(alpha, triangulation.simplices, radii, xys)
-    return geoms
+    return _filter_holes(geoms, xys)
 
 
 def _valid_hull(geoms, points):
@@ -703,6 +703,12 @@ def _construct_centers(a, b, radius):
     else:
         return down_x, down_y
 
+def _filter_holes(geoms, points):
+    if (geoms.interiors.apply(len) > 0).any():
+        from geopandas import points_from_xy
+        has_points, _ = points_from_xy(*points.T).sindex.query_bulk(geoms.values.data, predicate='contains')
+        geoms = geoms[geoms.index.isin(has_points)]
+    return geoms
 
 if __name__ == "__main__":
 
--- a/libpysal/cg/alpha_shapes.py
+++ b/libpysal/cg/alpha_shapes.py
@@ -6,10 +6,12 @@ by Tim Kittel (@timkittel) available at:
 
 Author(s):
     Dani Arribas-Bel daniel.arribas.bel@gmail.com
+    Levi John Wolf levi.john.wolf@gmail.com
 """
 
 import numpy as np
 import scipy.spatial as spat
+from scipy import sparse
 
 from ..common import requires, jit, HAS_JIT
 
@@ -380,7 +382,8 @@ def alpha_geoms(alpha, triangles, radii, xys):
     geoms : GeoSeries
         Polygon(s) resulting from the alpha shape algorithm. The GeoSeries
         object remains so even if only a single polygon is returned. There is
-        no CRS included in the object.
+        no CRS included in the object. Also note that holes are included in addition
+        to exteriors. 
 
     Examples
     --------
@@ -424,7 +427,7 @@ def alpha_geoms(alpha, triangles, radii, xys):
 
 
 @requires("geopandas", "shapely")
-def alpha_shape(xys, alpha):
+def alpha_shape(xys, alpha, filter_holes=True):
     """Alpha-shape delineation (Edelsbrunner, Kirkpatrick & Seidel, 1983) from a collection of points
 
     Parameters
@@ -481,7 +484,9 @@ def alpha_shape(xys, alpha):
     radii = r_circumcircle_triangle(a_pts, b_pts, c_pts)
     del triangles, a_pts, b_pts, c_pts
     geoms = alpha_geoms(alpha, triangulation.simplices, radii, xys)
-    return _filter_holes(geoms, xys)
+    if filter_holes:
+        geoms = _filter_holes(geoms, xys)
+    return geoms
 
 
 def _valid_hull(geoms, points):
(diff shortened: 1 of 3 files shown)


ljwolf:
OK, I've switched to a zorder search method: 
1. Extract the exterior ring of all shapes `geoms`. Call this `shells`. 
2. Compute `shells.sindex.query_bulk(geoms, predicate='within')` to determine which shells each geometry falls within. 
3. Construct the "within" graph adjacency matrix from these edges. 
4. Compute the row sum of this adjacency matrix. *This represents the number of shells in which a geometry falls, self-inclusive,* also known as the "z-order". 
5. Select all geometries with an odd z-order, because all holes have an even z-order and all non-holes are odd. Reprising the code comment: 
>        # A z-order of 1 means the polygon is only inside if its own exterior. This means it's a polygon.
>        # A z-order of 2 means the polygon is inside of exactly one other exterior. Because
>        #   the hull generation method is restricted to be planar, this means the polygon is a hole. 
>        # In general, an even z-order means that the polygon is always exactly matched to one exterior. 
>        #   This means that the polygon forms the exterior of a hole. 
>        # In general, an odd z-order means that there is an uneven number of exteriors. 
>        #   This means the polygon is filled. 

Figure_1

ljwolf:
@cornhundred, this should be good to go, and solves your reference case. Is there any other known pattern you'd want to test on?

cornhundred:
Thanks @ljwolf! The random matrix with holes should cover our use case. Does it return a MultiPolygon or a list of Polygons?

Also, I started to scope out how to allow alternate alpha shape metrics (radius, distance, etc) https://github.com/cornhundred/libpysal/tree/alt_alpha_metrics I'll make a pull request when I've cleaned it up more and meet the pull request requirements (testing etc). For our case we would like to set the threshold based on max triangle length.

ljwolf:
It keeps the return type as a GeoSeries. Would you prefer a single (multi)polygon? 

The `alpha_shape_auto` returns the single shapely polygon, so keeping the return type consistent might be a good idea. However, that'd involve changing the return type of `alpha_shape`... which would break things.

cornhundred:
Ok, keeping it the way it is should be fine. It is easy enough to calculate a MultiPolygon from it. We also want to calculate properties of the individual polygons (area, number of holes, etc) so that would be easier if it returns a list of discontiguous polygons.

darribas:
Hello,

Thanks for this @ljwolf for the PR and @cornhundred for bringing it up and chipping in!

I've looked over and it seems all fine, very clever trick!

A couple of thoughts:

1. Would it be worth to add a test to test both options in `alpha_shape`?
2. Right now the holes are discarded, which is probably fine in most applications. Would it be worth considering using the holes to generate "holed" shapes? I'm not sure that's the name, but a `MultiPolygon` where inside the hole is not part of the polygon? That could also be an option (`filter_holes=False/'remove'/'include'`, or something like that).
3. It'd be great to add documentation for the `filter_holes` parameter in `alpha_shape`.

cornhundred:
Hey @darribas, those are great ideas I'm happy to try and help address them. We definitely care about the holes but I'm not sure if it makes sense to pass them as separate polygons rather than derive them later from polygons.

For some background, we're working with single-cell spatial transcriptomics data - see [mouse liver data release](https://info.vizgen.com/mouse-liver-access) and example image below

Vizgen MERFISH Mouse Liver Map 

The holes in our case can be blood vessels (or artifacts from sectioning) and alpha shapes seem like a great way to quantify them. Also here's a [Colab notebook from a brain dataset](https://twitter.com/vizgen_inc/status/1464626128484806663?t=msC12F2gZ7lnn57a6g648w&s=19) where we interactively visualize our data.

martinfleis:
>  I'm not sure that's the name, but a MultiPolygon where inside the hole is not part of the polygon? 

That can still be a Polygon as long as it has a single part (with N holes in it but not another part within that hole).

ljwolf:
> holed shapes

The shapes we return are already holed. We currently return polygons with holes *and also* the holes themselves as separate extra polygons... Below shows our current behavior:
Figure_1
Currently, we return all of these shapes mixed together in a single GeoSeries. We don't differentiate which is the "hull" (with holes) and which are extra hole polygons. Further, you can't immediately tell which polygons represent the "hull" and which are just extra "holes," hence the z-order search. 

> `False/'ignore'/'include'`

Upon reflection, I actually we should remove the option and always filter. Two reasons:
1. **80% use case**: People looking for the `alpha_shape()` are probably expecting one (or more) polygons that represent the alpha hull of the input point set. In this case, returning duplicate hole polygons (with or without labelling) is giving the user "extra" information they didn't ask for, aren't interested in, and can't tell apart from what they *are* interested in: the alpha hull. 
2. **Promote "good" patterns**: If they do want them, then we should push users towards standard geopandas functionality (e.g. `result.interiors.explode()`) to extract holes from polygons. The information they want is already there in the holey polygons in the *filtered* geoseries.

ljwolf:
I've added a `test_holes()` case to the `test_ashapes.py` now. If we are OK w/ filtering (no option) then I will add a note/discussion in the docstring about holes and commit. Otherwise, I have the docstring for the option ready to go, and can modify the test easily to check the options.

cornhundred:
Hi @ljwolf, I agree with the always filtering logic. Since the holes are properties of specific polygons within the alpha shape we would want users to obtain holes with metadata about their 'parent' polygon, but that should be outside the scope of the alpha shape method.

cornhundred:
Hi, just checking on the status of this pull request. I can try to make an additional pull request to have the option for different alpha shape distance thresholds this week so maybe it makes sense to wait on that before making a potential new release.

ljwolf:
A determination about the proper behavior of the option from a reviewer (e.g. @darribas @martinfleis @sjsrey) would get this moving again. Happy to raise this on our monthly developer meeting friday if there is no assent by then!

cornhundred:
Great thanks @ljwolf appreciate it.

libpysal/cg/alpha_shapes.py
```diff
@@ -380,7 +382,8 @@ def alpha_geoms(alpha, triangles, radii, xys):
     geoms : GeoSeries
         Polygon(s) resulting from the alpha shape algorithm. The GeoSeries
         object remains so even if only a single polygon is returned. There is
-        no CRS included in the object.
+        no CRS included in the object. Also note that holes are included in addition
+        to exteriors.
```
martinfleis, line 386: This is probably not the case with `_filter_holes` anymore.

martinfleis (approved):
I agree with @ljwolf. I believe that we should return the polygons that represent the alpha shape and not those that represent holes. As said, those are already included as `interiors` so we would be just duplicating the same information.

Apart from that one note in the docstring, this seems to be ready to be merged.

libpysal/cg/alpha_shapes.py
ljwolf: Ah, so, this is slightly confusing. 
There are two functions:
-  `alpha_shape(xys, alpha)`, which returns a GeoSeries for a given `alpha` and input points `xys`.
- `alpha_geoms(alpha, triangles, radii, xys)`, which does the same, but using the delaunay triangulation and its corresponding circumcircle radii. 

I have only filtered the `alpha_shape()` output. `alpha_geoms()` still can return holes. 

I didn't want to add the z-order search into `alpha_geoms()` because it's used repeatedly in a performance-critical loop by `alpha_shape_auto()`.  That function constructs a single output Polygon that covers the input point set. For its particular case, the holes returned by `alpha_geoms()` won't matter because of the search procedure used in `alpha_shape_auto()`. So, if the z-order search were introduced in `alpha_geoms`, it would introduce an unnecessary (but slight) performance penalty.

Thinking about this, the "right" solution would be to make `alpha_geoms()` private to `alpha_shape_auto()`, but we don't know if anyone's using that function. I'm, thinking to solve this:
- make an unfiltered private version of `alpha_geoms()` to preserve `alpha_shape_auto()` speed
- add a filter to `alpha_geoms()`, too.
- deprecate the public `alpha_geoms()` method. 

Thoughts?

cornhundred: It seems reasonable to me to make `alpha_geoms()` private since users will be directed to use `alpha_shape()`.

martinfleis: `alpha_geoms` is not a part of public API as far as I can tell. It is not mentioned in the API documentation and it is not listed in `__all__` in this file. So I'd say that we do not have to deprecate it and can change as we wish.

ljwolf: Awesome. Then I'll add an underscore to the name to be extra careful, and then merge this. 

Expect a point release with these changes tomorrow UK time @cornhundred

cornhundred: Thanks!!

ljwolf:
tests are now failing in another part of the code. Looks like a networkx issue in Python 3.7. will investigate & merge if I can figure it out.

ljwolf:
OK, well, networkx dropped support for python 3.7; we'll need to drop that too, then, or skip the test.

ljwolf:
Or, I suppose we could pin to networkx 2.6 for Python 3.7. Will investigate.

cornhundred:
Thanks!

PR #471, opened 2022-06-12
Remove non-existant parameter `tolerance` from `voronoi_frames` doc comment

john-h-kastner:
This parameter was added in 33b500b but subsequently dropped in b5cde8a without updating the doc comment.


jGaboardi:
@john-h-kastner Good eye. Thanks for the contribution!

jGaboardi (approved):

PR #472, opened 2022-06-15
update nx conversions to be 3.0 compatible

mjschwenne:
Greeting! I am one of the NetworkX core developers. We are preparing to move to NetworkX 3.0 soon and that mostly involves removing a lot of deprecated code. As part of this update, we are dropping support for numpy matrices and scipy sparse matrices in favor of numpy arrays and scipy sparse arrays. While removing the old code from our own repo we discovered that some of our examples, which use libpysal, weren't working since functions such as `nx.from_scipy_sparse_matrix` and `nx.to_numpy_matrix` were still being used internally to libpysal. 

I've composed a simple PR which should ensure that libpysal remains compatible with networkx 3.0 by using the drop-in replacements for the removed functions. All of the tests are passing for me locally, so the changes don't seem to cause any issues. 

If you have any more questions about the upcoming networkx 3.0 update please ask myself, @rossbar, @dschult or @jarrodmillman.

### DIFF
--- a/libpysal/weights/tests/test_nx.py
+++ b/libpysal/weights/tests/test_nx.py
@@ -13,17 +13,17 @@ except ImportError:
 class Test_NetworkXConverter(ut.TestCase):
     def setUp(self):
         self.known_nx = nx.random_regular_graph(4, 10, seed=8879)
-        self.known_amat = np.array(nx.to_numpy_matrix(self.known_nx))
+        self.known_amat = nx.to_numpy_array(self.known_nx)
         self.known_W = lat2W(5, 5)
 
     def test_round_trip(self):
         W_ = W.from_networkx(self.known_nx)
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks! How far backward is this compatible? We may want to bump a min version of networkx supported. Though I'm not sure we mention it anywhere right now.

mjschwenne:
The replacement functions `to_scipy_sparse_array` and `from_scipy_sparse_array` were added in [networkx 2.7](https://networkx.org/documentation/stable/release/release_2.7.html).

martinfleis:
This change actually requires scipy 1.8 or newer (Feb 22). Since we do not require it now, I will add a flag to skip the test.

martinfleis (approved):
Thanks!

jGaboardi (approved):

PR #480, opened 2022-10-22
max val in rtree.silhouette_coeff

jGaboardi:
This PR provides an overdo solution to a rough edge in `cg.rtree`. 

xref https://github.com/pysal/spopt/issues/57

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos: 
 -   repo: https://github.com/psf/black
-    rev: 21.12b0
+    rev: 22.6.0
     hooks:
     - id: black
-      language_version: python3.9
+      language_version: python3
--- a/libpysal/cg/rtree.py
+++ b/libpysal/cg/rtree.py
@@ -59,17 +59,17 @@ class Rect(object):
 
     def overlap(self, orect):
         """Return the overlapping area of two rectangles.
-        
+
         Parameters
         ----------
         orect : libpysal.cg.Rect
             Another rectangle.
-        
+
         Returns
         -------
         overlapping_area : float
             The area of the overlap between ``orect`` and ``self``.
-        
+
         """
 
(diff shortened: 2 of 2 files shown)


jGaboardi:
Tests are passing locally. [Failure](https://github.com/pysal/libpysal/actions/runs/3301624340/jobs/5447307217#step:4:1744) due to a fresh issue in `geopandas` through `fiona`.

martinfleis:
Can we test this?

jGaboardi:
> Can we test this?

Probably a good idea to... Let me try to carve out some time to look into it unless you already have an idea?

martinfleis:
> unless you already have an idea

nope

jGaboardi:
In order to exactly test this we need to know of a situation where "there are corner cases when silhoutte for two clusters are exactly the same" (see [here](https://github.com/pysal/spopt/issues/57#issue-650666401)). I have [pinged](https://github.com/pysal/spopt/issues/57#issuecomment-1292343760) the OPs for a their use case, not sure how likely we are to hear back since the first post was nearly 5 years ago...

martinfleis:
Tricky, yeah. 

Out of curiosity, where is Rtree used in spopt? Can we use STRtree from shapely 2.0 or `rtree` package instead? Both are C++, so likely much faster.

jGaboardi:
> Out of curiosity, where is Rtree used in spopt?

Seems as it is actually not even being used in `spopt` at all.

jGaboardi:
> Is it used anywhere? Is there a value in keeping it around or is that something that could be deprecated?

Good question...

ljwolf:
I believe it's only used in the "locators" code in libpysal, which I also don't think is used. +1 for deprecation of that and depending code.

martinfleis:
> I believe it's only used in the "locators" code in libpysal

I've seen that but that whole module seems to be deprecated and can be removed.

Issue #485, opened 2022-10-28
allow for individual downloads and access to  remote urls in examples


sjsrey:
- [ ] allow for get_path() to list remote paths, too
- [ ] enhance to only download the requested data

#484

jGaboardi:
This issue can be closed out and the three check boxes in #484 can be ticked off, correct?

sjsrey:
Good catch.

jGaboardi:
Can confirm this works in the wild: pysal/splot#165

PR #486, opened 2022-10-31
ENH: more granularity in example handling

sjsrey:
Address #485

### DIFF
--- a/examples.ipynb
+++ b/examples.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 46,
+   "execution_count": 1,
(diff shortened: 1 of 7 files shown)


examples.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
```
jGaboardi, line 2: Is this more of a scratch notebook? I am wondering if it needs to be included here?

sjsrey: It gets pulled into the docs. So not scratch.

jGaboardi: ah, OK.

jGaboardi (approved):

PR #491, opened 2022-11-21
Fix Linux user path test

QuLogic:
3. [n/a] This pull introduces new functionality covered by
4. [ ] You have [assigned a
5. [ ] The justification for this PR is: 

There is no requirement that user directories start with `/home`. In
fact, if `XDG_DATA_HOME` is set, there's no guarantee of anything
checked in this test.


sjsrey (approved):
Nice catch.

PR #494, opened 2022-11-30
Move WSP2W into weights.py

sjsrey:
`WSP2W` was not imported so https://github.com/pysal/libpysal/blob/master/libpysal/weights/weights.py#L235 would fail. However, importing results in problems with circular imports. So the solution here is to add that function to `weights.py`.

This also starts some deprecation.


sjsrey:
Ignore, this breaks other things.

Issue #495, opened 2022-11-30
WSP2W missing in weights.py


sjsrey:
As a result, [this](https://github.com/pysal/libpysal/blob/master/libpysal/weights/weights.py#L235) will fail.

However, importing WSP2W into weights raises other errors due to circular imports.

I can in-line the logic in `from_WSP` rather than calling `WSP2W`.

PR #496, opened 2022-12-16
Implements W.from_WSP

sjsrey:
Addresses #495


libpysal/weights/weights.py
```diff
@@ -232,7 +232,69 @@ def from_shapefile(cls, *args, **kwargs):
...
+
+        Create a standard PySAL W from this sparse weights object.
+
+        >>> w = W.from_WSP(wsp)
+        >>> w.n
+        10
+        >>> print(w.full()[0][0])
+        [0 1 0 0 0 1 0 0 0 0]
+        """
+        WSP.sparse
```
jGaboardi, line 275: Is there a reason for calling `WSP.sparse` here?

libpysal/weights/weights.py
```diff
@@ -232,7 +232,69 @@ def from_shapefile(cls, *args, **kwargs):
...
+        indices = WSP.sparse.indices
```
jGaboardi, line 276: Seems like `indices` does not need to be assigned here and `WSP.sparse.indices` can be used directly in line 282:

```python 
indices = [id_order[i] for i in WSP.sparse.indices]
```

libpysal/weights/weights.py
sjsrey: good catch. I was pulling this over from https://github.com/pysal/libpysal/blob/master/libpysal/weights/util.py#L788 .

Will update here and there.

libpysal/weights/weights.py
sjsrey: Good catch. Will change.

jGaboardi (approved):

jGaboardi:
Maybe pulled the approval trigger too fast. We should figure a work around for failures in 3.7.

sjsrey:
> Maybe pulled the approval trigger too fast. We should figure a work around for failures in 3.7.

Or we could consider dropping support for 3.7?

jGaboardi:
> > Maybe pulled the approval trigger too fast. We should figure a work around for failures in 3.7.
> 
> 
> 
> Or we could consider dropping support for 3.7?

I 100% support this.

knaaptime:
cool with me. its outside the 'last three' release window we commit to

PR #503, opened 2023-01-03
fix: Import-time fetching of example dataset metadata by refactoring top-level __init__.py 

### LINKED ISSUE #499: Import of libpysal causes libpysal examples to download which slows down initialization or completely hangs
Thank you for filing this issue! To help troubleshoot this issue, please follow
the following directions to the best of your ability before submitting an issue.
Feel free to delete this text once you've filled out the relevant requests. 

Please include the output of the following in your issue submission. If you don't know how to provide the information, commands to get the relevant information from the Python interpreter will follow each bullet point.

Feel free to delete the commands after you've filled out each bullet. 

- Platform information:
```python
>>> import os; print(os.name, os.sys.platform);print(os.uname())
>>> posix.uname_result(sysname='Linux', nodename='*', release='5.15.76-1-MANJARO', version='#1 SMP PREEMPT Sat Oct 29 14:22:16 UTC 2022', machine='x86_64')
```
- Python version: 
```python
>>> import sys; print(sys.version)
>>> 3.10.8 (main, Oct 13 2022, 21:13:48) [GCC 12.2.0]
```
- SciPy version:
```python
>>> import scipy; print(scipy.__version__)
>>> 1.9.3
```
- NumPy version:
```python
>>> import numpy; print(numpy.__version__)
>>> 1.24.0
```

Importing libpysal, such as https://github.com/pysal/momepy/blob/99cf555f3a8c93356c280ccfc5d13111d0518427/momepy/weights.py#L4, will import the whole libpysal library. This includes the examples module https://github.com/pysal/libpysal/blob/284d2ae3d8d2084e6c74901d79cc0e13029c6cce/libpysal/__init__.py#L26, which on import attempts to download some remote datasets https://github.com/pysal/libpysal/blob/284d2ae3d8d2084e6c74901d79cc0e13029c6cce/libpysal/examples/remotes.py#L110 .

This has caused `requests.get` to hang here https://github.com/pysal/libpysal/blob/284d2ae3d8d2084e6c74901d79cc0e13029c6cce/libpysal/examples/remotes.py#L24

Is there a way around this?

I have also opened a issue here https://github.com/pysal/momepy/issues/451

iwishiwasaneagle:
Refactor import process to ensure that __init__ files for the individual modules are not run at import time if not required which closes #499 

justification in point 5. 

1. [ x] You have run tests on this submission locally using `pytest` on your changes. Continuous integration will be run on all PRs with [GitHub Actions](https://github.com/pysal/libpysal/blob/master/.github/workflows/unittests.yml), but it is good practice to test changes locally prior to a making a PR.
2. [x ] This pull request is directed to the `pysal/master` branch.
4. [ ] You have [assigned a

#499

### DIFF
--- a/libpysal/__init__.py
+++ b/libpysal/__init__.py
@@ -20,10 +20,6 @@ io
 weights
     Tools for creating and manipulating weights
 """
-from . import cg
-from . import io
-from . import weights
-from . import examples
 
 from . import _version
 


iwishiwasaneagle:
I was unsure whether to just remove `from . import examples` or also the other options. I went for the latter to start a discussion. Tests still pass but I'm worried the import API will change too dramatically. The other option is to just remove `examples` from being imported but this leads to an inconsistent import API.

iwishiwasaneagle:
I see the https://github.com/pysal/libpysal/actions/runs/3828396150/workflow#L16 doesn't like this change. Will update the CI runs after a discussion on the above.

sjsrey:
> I see the https://github.com/pysal/libpysal/actions/runs/3828396150/workflow#L16 doesn't like this change. Will update the CI runs after a discussion on the above.

If you add to this PR by changing:

https://github.com/pysal/libpysal/blob/d291e014538f0c8862b70bd0119745aa8c55424a/.github/workflows/unittests.yml#L16

to
```
 FETCH_EXAMPLES: python -c 'import libpysal.examples; libpysal.examples.fetch_all()' 
```

The CI may proceed.

sjsrey:
> I was unsure whether to just remove `from . import examples` or also the other options. I went for the latter to start a discussion. Tests still pass but I'm worried the import API will change too dramatically. The other option is to just remove `examples` from being imported but this leads to an inconsistent import API.

Limiting the changes to the `examples` module seems like the way to go here. Changing this for the other modules may break downstream use cases, even though our tests are green.

libpysal/__init__.py
```diff
@@ -20,10 +20,6 @@
 weights
     Tools for creating and manipulating weights
 """
-from . import cg
```
sjsrey, line 23: leave in lines 23-25 (i.e., keep the imports for cg, io, and weights) as they don't do any polling of remotes.

knaaptime:
i think even removing `examples` from the init will cause some downstream breakage (i know for sure it will break some of the tests in other subpackages where we routinely do stuff like [`from libpysal.examples import load_example`](https://github.com/pysal/segregation/blob/b3643f7af0aeaed1caf5a39b4e57c19a5c5f3459/segregation/tests/test_absolute_concentration.py#L2), which would need to be updated.

i dont think its necessarily a dealbreaker, but alternatively, we could do a try/except to allow polling to fail or hardcode the paths and skip the polling step

sjsrey:
@knaaptime good points on side-effects from making the import of `examples` explicit and not automatic.

I can work on refactoring the remote logic so that no polling is required on init.

The import would be automatic but without the cost of requests. Requests would only happen on a load.

iwishiwasaneagle:
Yup, good point @knaaptime. However, any networking happening at import time should be removed.

sjsrey:
ref #504

PR #504, opened 2023-01-04
Remove automatic polling of remotes

sjsrey:
This addresses #499 by no longer polling the remotes on init.

Remotes would be curated prior to each release instead.


knaaptime (approved):
sweet

iwishiwasaneagle (approved):

martinfleis:
One more thing on this. GeoPandas is deprecating `datasets` module (https://github.com/geopandas/geopandas/issues/2751) and we want to offer a convenient way of replacing this with something like https://github.com/martinfleis/xyzdata. I've used the data we have in here right now as a test and thought that the package may eventually live under pysal org and replace `libpysal.examples`. Thoughts?

PR #506, opened 2023-01-06
draft perimeter weights from seg

### LINKED ISSUE #80: ENH: shared perimeter contiguity weighting.
When we're coming from geodataframes, we can probably implement a shared perimeter weighting scheme pretty easily. This means that, instead of contiguity being a binary relation, we'd assign the weight to the adjacency graph according to how much of the polygon's perimeter is shared along their shared edge. 

For Queen weights, this would necessarily reduce them down to Rook weights, so we would implement this only for Rook. 

0. do a first pass for binary contiguity
1. for neighbors:
    1. `poly_focal.intersection(poly_neighbor).length / poly_focal.boundary.length` would give the asymmetric perimeter weight
    2. `poly_focal.intersection(poly_neighbor).length / (poly_focal.boundary.length + poly_neighbor.length)` would give a symmetrized perimeter weight

This would be an interesting addition to the library. It could be implemented as a function in `util`, and then applied at the end of initialisation for `Rook.from_shapefile` or `from_dataframe` if `perimeter=True`.

knaaptime:
first draft moving @ljwolf's perimeter-weighting weights code over from segregation

resolves #80

### DIFF
--- a/libpysal/weights/contiguity.py
+++ b/libpysal/weights/contiguity.py
@@ -1,16 +1,18 @@
 import itertools
 
 import numpy
+import pandas as pd
 
 from ..cg import voronoi_frames
 from ..io.fileio import FileIO
 from ._contW_lists import ContiguityWeightsLists
+from .raster import da2W, da2WSP
 from .util import get_ids, get_points_array
 from .weights import WSP, W
-from .raster import da2W, da2WSP
 
 try:
     from shapely.geometry import Point as shapely_point
+
     from ..cg.shapes import Point as pysal_point
 
     point_type = (shapely_point, pysal_point)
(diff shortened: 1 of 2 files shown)


libpysal/weights/contiguity.py
```diff
@@ -631,3 +647,64 @@ def buildContiguity(polygons, criterion="rook", ids=None):
         return Queen(polygons, ids=ids)
     else:
         raise Exception('Weights criterion "{}" was not found.'.format(criterion))
+
+
+def _return_length_weighted_w(w, data):
```
sjsrey, line 652: data is not defined

knaaptime: here its an input to the `_return_length_weighted_w` function being defined, (the `df` argument from w.from_dataframe()` is what gets passed in that position in 194

sjsrey: should the docstring be updated though?

knaaptime: ah, right. I'm not terribly worried about this particular docstring since its now a private function, but i forgot to include the extra `perimeter` argument docstring for the `from_dataframe` method

knaaptime: just occurred to me, might we want an option to normalize the perimeter weights? i.e. the weight for unit i is $*shared boundary ij* / *boundary i*$?

knaaptime: that would be different than row-standardized weights for observations on the periphery, but probably not a super useful enhancement

Issue #484, opened 2022-10-28
libpysal deprecation 


sjsrey:
##### weights

- [  ] keep everything in distance & contiguity, but reorganize the voronoi, delaunay, gabriel, and relative neighborhood into a triangulations.py
- [  ] move wsp out of spint
- [  ] move spint weights into spint itself, since no one else has used it since then
- [  ] move perimeter-weighted contiguity into contiguity
- [  ] move some stuff from momepy 
- [  ] deprecate adjtools.py

#### raster

- [  ] migrate da2W/da2WSP into W.from_da() methods instead, so deprecate the functions. 

#### util & user

- [  ] WSP2W gets deprecated in favour of WSP.to_w()
- [  ] full2W gets deprecated in favour of a W.from_dense()
- [  ] full() gets deprecated in favour of a W.to_dense()
- [  ] fill_diagonal() is retained
- [  ] keep higher_order, shimbel gets re-written to use csgraph.floydwarshall
- [  ] comb, remap_ids, get_ids, get_points_array_from_shapefile gets deprecated
- [  ] consolidate higher_order, order, and heigher_order_sp to one function
- [  ] min_threshold_distance move to cg
- [  ] lat2SW, hexlat2W moves to same module as lat2W
- [  ] min_threshold_dist, build_lattice_shapefile, spw_from_gal, w_local_cluster, spw_from_gal deprecated
- [  ] attach_islands, nonplanar, fuzzy is kept
- [  ] neighbor_equality should be enhanced, made to recognize the difference between "equality" and "isomorphism"
    - [  ] two weights are equal when their matrix representations are the same (i.e. they're ordered the same)
    - [  ] two weights are isomorphic when the ids are equal, links are present in both, and weights are equal.
- [  ] the rest of weights should be kept (w_sets, lag_spatila/categorical ,etc. )

#### cg

- [  ] all of sphere will be deprecated
- [  ] all of shapes will get deprecated
- [  ] all of standalone will get deprecated
- [  ] locators are already deprecated and should be removed
- [  ] kdtree should be deprecated, point users to BallTree with haversine metric as an alternative to Arc_KDTree
- [  ] deprecate ops
- [  ] deprecate shapely_ext
- [  ] deprecate segmentLocator

#### io

- [  ] deprecate all io functionality not reading weights 
- [  ] keep weights readers, wrap into a W.from_file() method that works kind of like the existing implementation

#### examples
- [  ] keep the module
- [  ] allow for get_path() to list remote paths, too
[...]

PR #507, opened 2023-01-12
[WIP] shared perimeter-weighted contiguity

### LINKED ISSUE #80: ENH: shared perimeter contiguity weighting.

knaaptime:
supercedes #506 ; resolves #80


knaaptime:
@sjsrey the last commit implements the logic we talked about today. I can add a test case, or we could have you sanity-check again?

libpysal/weights/contiguity.py
```diff
@@ -748,3 +774,68 @@ def buildContiguity(polygons, criterion="rook", ids=None):
...
+    merged["geometry_neighbor"] = gpd.GeoSeries(merged.geometry_neighbor)
+
+    # Getting the shared boundaries
+    merged["shared_boundary"] = merged.geometry_focal.intersection(
+        merged.set_geometry("geometry_neighbor")
+    )
+
+    # Putting it back to a matrix
+    if perimeter_standardize:
+        merged['weight'] = merged.set_geometry("shared_boundary").length / total_boundary_length
```
martinfleis, line 828:
```suggestion
        merged['weight'] = merged["shared_boundary"].length / total_boundary_length
```
as above

libpysal/weights/contiguity.py
```diff
@@ -748,3 +774,68 @@ def buildContiguity(polygons, criterion="rook", ids=None):
...
+    else:
+        merged["weight"] = merged.set_geometry("shared_boundary").length
```
martinfleis, line 830:
```suggestion
        merged["weight"] = merged["shared_boundary"].length
```

libpysal/weights/contiguity.py
```diff
@@ -748,3 +774,68 @@ def buildContiguity(polygons, criterion="rook", ids=None):
...
+        suffixes=("_focal", "_neighbor"),
+    )
+    # Transforming from pandas to geopandas
+    merged = gpd.GeoDataFrame(merged, geometry="geometry_focal")
+    total_boundary_length = merged.boundary.length
+    merged["geometry_neighbor"] = gpd.GeoSeries(merged.geometry_neighbor)
+
+    # Getting the shared boundaries
+    merged["shared_boundary"] = merged.geometry_focal.intersection(
+        merged.set_geometry("geometry_neighbor")
```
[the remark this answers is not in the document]
knaaptime: thanks

sjsrey (approved):

knaaptime:
ive got a notebook to go with this, but running into issues when i create a perimeter-weighted W with `use_index=True` after setting a meaningful index. I think the issue has to do with how ids are handled in `to_adjlist` but still investigating

cc: @martinfleis @ljwolf

knaaptime:
```python
import geopandas as gpd
from libpysal.examples import load_example
from libpysal.weights import Rook

us = gpd.read_file(examples.get_path("us48.shp"))

old = Rook.from_dataframe(us)
new = Rook.from_dataframe(us.set_index('STATE_FIPS'), use_index=True)
```
Screenshot 2023-01-13 at 2 45 08 PM
Screenshot 2023-01-13 at 2 45 16 PM

martinfleis:
Did I mess up #477?

knaaptime:
cant say for certain, but i think all of that is ok. I think we may just need to update the to_adjlist method

knaaptime:
im not sure whether its the `to_adjlist` method to blame, actually. I can get those tables above to match if i reset the index prior to return in the current to_adjlist method, but the problem persists

[this notebook](https://github.com/knaaptime/libpysal/blob/perim_update/notebooks/shared_perimeter_weights.ipynb) runs fine as-is. But, if i explicitly set the index to something like STATE_FIPS in cell 25, then `use_index=True` with the perimeter weights, the code will run fine, but will produce incorrect results. Still not sure why

sjsrey:
It could be due to `to_adjlist` relying on `id_order`  https://github.com/pysal/libpysal/blob/3a09c422b6650c217357a7f0fe5ade0799cf87db/libpysal/weights/weights.py#L435

which we have not yet [deprecated](https://github.com/pysal/libpysal/issues/484).

sjsrey:
Maybe changing: 

https://github.com/pysal/libpysal/blob/3a09c422b6650c217357a7f0fe5ade0799cf87db/libpysal/weights/weights.py#L434-L440

to

```
        focal_ix, neighbor_ix = self.sparse.nonzero()
        weights = self.sparse.data
        adjlist = pandas.DataFrame(
            {focal_col: focal_ix, neighbor_col: neighbor_ix, weight_col: weights}
```
?

sjsrey:
I think the `to_adjlist` issue is sorted with #510:

image

knaaptime:
that gets the alignment back in shape so the weights are correct, but we lose the index itself (in the updated example 'focal' is an integer sequence, not fips codes)

sjsrey:
This is because `to_ajlist` is grabbing the indices from the w.sparse attribute, which only has integer indices.

knaaptime:
got it

so in the new api, we take the `ids` kwarg and deprecate the idvariable and id_order. But then we never expose `W.ids` publicly. I'm not sure whether that was intentional, but it means when we go back and forth from sparse, we have touble redoing the logic e.g. [here](https://github.com/pysal/libpysal/blob/6e805a5652856a99c98ab8426958d5975f0a72b4/libpysal/weights/weights.py#L1664). We could use use `w.neighbors.keys()` in place of id_order there since we know the dict is sorted, (but in that case are we keeping `id_order_set` or some other indicator that we could test against? otherwise we need to update the dict blindly even if unnnecessary)

sjsrey:
> got it
> 
> so in the new api, we take the `ids` kwarg and deprecate the idvariable and id_order. But then we never expose `W.ids` publicly. I'm not sure whether that was intentional, but it means when we go back and forth from sparse, we have touble redoing the logic e.g. [here](https://github.com/pysal/libpysal/blob/6e805a5652856a99c98ab8426958d5975f0a72b4/libpysal/weights/weights.py#L1664). We could use use `w.neighbors.keys()` in place of id_order there since we know the dict is sorted, (but in that case are we keeping `id_order_set` or some other indicator that we could test against? otherwise we need to update the dict blindly even if unnnecessary)

This was a good catch, as it raises a number of issues that are related to the planned deprecation.

knaaptime:
if we are getting rid of `id_order_set`, and we've decided that exposing `W.ids` is redundant, i guess when roundtripping to sparse, we could do a simple test with something like `list(w.neighbors.keys()) == list(range(len(w.neighbors))` and remap ids if false?

sjsrey:
> ```python
> import geopandas as gpd
> from libpysal.examples import load_example
> from libpysal.weights import Rook
> 
> us = gpd.read_file(examples.get_path("us48.shp"))
> 
> old = Rook.from_dataframe(us)
> new = Rook.from_dataframe(us.set_index('STATE_FIPS'), use_index=True)
> ```
> 
> Screenshot 2023-01-13 at 2 45 08 PM
> 
> Screenshot 2023-01-13 at 2 45 16 PM
> 
> i thnk the neighbor relationships here are correct, but the indices get screwed up (which causes issues for the perimeter weights, as they rely on a join)

`to_adjlist` sorts before returning. If it did not sort, the returned df would align with what you are expecting:

```
   focal neighbor  weight
0      53       16     1.0
1      53       41     1.0
2      30       38     1.0
3      30       46     1.0
4      30       56     1.0
..    ...      ...     ...
205    12       01     1.0
206    12       13     1.0
207    26       55     1.0
208    26       18     1.0
209    26       39     1.0
```

We could add a kw argument to implement this.

knaaptime:
i think we also need to store (and expose publicly) the `ids` property. We will need that as a mapping any time we roundtrip through a WSP

knaaptime:
i think #511 solves the core issue, so this is probably ready as well.

ljwolf:
To make sure, this has already landed in [`libpysal.graph`, right?](https://github.com/pysal/libpysal/blob/main/libpysal/graph/_contiguity.py#L186) This should be closed if I'm reading that correctly, imho, since we should probably freeze new features in `libpysal.weights`.

knaaptime:
yup, this is stale

libpysal/weights/contiguity.py
```diff
@@ -748,3 +774,68 @@ def buildContiguity(polygons, criterion="rook", ids=None):
...
+    total_boundary_length = merged.boundary.length
+    merged["geometry_neighbor"] = gpd.GeoSeries(merged.geometry_neighbor)
+
+    # Getting the shared boundaries
+    merged["shared_boundary"] = merged.geometry_focal.intersection(
+        merged["geometry_neighbor"]
+    )
+
+    # Putting it back to a matrix
+    if perimeter_standardize:
```
knaaptime, line 827: @ljwolf do we want to keep the option to standardize when the boundary isnt exhausted? (so the denom is boundary_i instead of \sum{sharedboundary_ij})?

PR #509, opened 2023-01-13
Remove sklearn dependency

### LINKED ISSUE #508: [BUG] now scikit-learn is a hard dependency
[`scikit-learn`](https://github.com/pysal/libpysal/blob/master/libpysal/weights/weights.py#L13) became a hard dependency in [#496](https://github.com/pysal/libpysal/commit/b5dfdd24595a4ce90c912a32e303a65d22724843#diff-e06ce570feae14d00eeca26d532fad05393c2ccee32dd82cbcffc135dd8243fdR13), but it never got migrated from [`requirements_plus_conda.txt`](`https://github.com/pysal/libpysal/blob/master/requirements_plus_conda.txt#L10`) to [`requirements.txt`](https://github.com/pysal/libpysal/blob/master/requirements.txt). CI is passing because [`scikit-learn` is getting pulled in there](https://github.com/pysal/libpysal/actions/runs/3897083329/jobs/6654378097#step:3:884), but this is causing failure downstream where it is [not being installed via `pip`](https://github.com/pysal/inequality/actions/runs/3898079286/jobs/6656437828#step:5:263) for the bleeding edge `libpysal`. 

So, the question now is:
1. Should `scikit-learn` be promoted to a hard dependency?; or
2. Should we move it to a method level import in [`W.to_sparse()`](https://github.com/pysal/libpysal/blob/ad1afcb1c54eb4263c583ebf3a55a9c76a13973a/libpysal/weights/weights.py#L478)?

sjsrey:
Rolling our own encoder to avoid a heavy dependency.

resolves #508

### DIFF
--- a/libpysal/weights/tests/test_weights.py
+++ b/libpysal/weights/tests/test_weights.py
@@ -3,7 +3,7 @@ import tempfile
 
 import unittest
 import pytest
-from ..weights import W, WSP
+from ..weights import W, WSP, _LabelEncoder
 from .. import util
 from ..util import WSP2W, lat2W
 from ..contiguity import Rook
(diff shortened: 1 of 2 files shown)


jGaboardi:
resolves #509

jGaboardi (approved):

jGaboardi:
All good to merge from my end. Shall we wait for other reviewers or pull the trigger?

martinfleis:
> resolves https://github.com/pysal/libpysal/issues/508

@jGaboardi that needs to be in the first post to work automatically. just fyi

martinfleis (approved):
I am fine with it, just wondering if it is not an overkill and calling `np.unique` and `np.searchsorted` twice in `to_sparse` directly wouldn't be enough. But it may be useful elsewhere.

PR #510, opened 2023-01-16
Removing internal use of id_order in to_adjlist

sjsrey:
We are going to deprecate `id_order` in the future, so the internal use here can be removed to address #507.

### DIFF
--- a/libpysal/weights/weights.py
+++ b/libpysal/weights/weights.py
@@ -432,12 +432,9 @@ class W(object):
 
         links = []
         focal_ix, neighbor_ix = self.sparse.nonzero()
-        names = np.asarray(self.id_order)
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Seems all good

knaaptime (approved):

martinfleis:
This will probably need to be revised as it broke tests in esda https://github.com/pysal/esda/issues/239 and momepy https://github.com/pysal/momepy/issues/461

martinfleis:
See the reproducer in https://github.com/pysal/momepy/issues/461#issuecomment-1396591258

martinfleis:
At least the one in momepy is fixed by #511 (haven't seen that before).

PR #511, opened 2023-01-17
Handle set_index properly in to_adjlist

sjsrey:
xref #507

### DIFF
--- a/libpysal/weights/tests/test_adjlist.py
+++ b/libpysal/weights/tests/test_adjlist.py
@@ -160,3 +160,30 @@ class Test_Adjlist(ut.TestCase):
         usv = np.array(["53", "53", "30", "30", "30"])
         np.testing.assert_array_equal(unsorted_al.focal.values[:5], usv)
         np.testing.assert_array_equal(sorted_al.focal.values[:5], sv)
+        
+    def test_ids(self):
+        df = geopandas.read_file(examples.get_path("columbus.dbf")).head()
+        df["my_id"] = range(3, len(df) + 3)
+        W = weights.Queen.from_dataframe(df, ids="my_id")
+        W_adj = W.to_adjlist(drop_islands=True)
+        for i in range(3, 8):
+            assert i in W_adj.focal
+            assert i in W_adj.neighbor
+        for i in W_adj.focal:
+            assert i in list(range(3, len(df) + 3))
+        for i in W_adj.neighbor:
+            assert i in list(range(3, len(df) + 3))
+
+    def test_str_ids(self):
+        df = geopandas.read_file(examples.get_path("columbus.dbf")).head()
+        snakes = ["mamba", "boa", "python", "rattlesnake", "cobra"]
+        df["my_str_id"] = snakes
+        W = weights.Queen.from_dataframe(df, ids="my_str_id")
+        W_adj = W.to_adjlist(drop_islands=True)
+        for i in snakes:
+            (W_adj.focal == i).any()
+            (W_adj.neighbor == i).any()
+        for i in W_adj.focal:
+            assert i in snakes
+        for i in W_adj.neighbor:
+            assert i in snakes
--- a/libpysal/weights/tests/test_adjlist.py
+++ b/libpysal/weights/tests/test_adjlist.py
@@ -145,3 +145,18 @@ class Test_Adjlist(ut.TestCase):
             np.testing.assert_allclose(
                 data, mapped["_".join(("subtract", name))].values
             )
+
+    def test_sort(self):
+        from libpysal import examples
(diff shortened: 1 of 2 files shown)


libpysal/weights/weights.py
```diff
@@ -432,6 +437,9 @@ def to_adjlist(
 
         links = []
         focal_ix, neighbor_ix = self.sparse.nonzero()
+        idxs = np.array(list(self.neighbors.keys()))
+        focal_ix = idxs[focal_ix]
+        neighbor_ix = idxs[neighbor_ix]
```
martinfleis, line 442: We don't need to do this if there are no custom ids set. I would add 

```py
self.ids = ids
```
to the `W.__init__` and then have this done conditionally after

```py
if self.ids:
    ...
```

You can potentially also use the ids directly here but that does not matter much

```py
        if self.ids:
            ids = np.asarray(self.ids)
            focal_ix = ids.take(focal_ix)
            neighbor_ix = ids.take(neighbor_ix)
```
Using keys from neighbours is probably safer option anyway.

libpysal/weights/tests/test_adjlist.py
```diff
@@ -145,3 +145,18 @@ def test_map(self):
...
+        us = geopandas.read_file(examples.get_path("us48.shp"))
+        w = Rook.from_dataframe(us.set_index("STATE_FIPS"), use_index=True)
+        unsorted_al = w.to_adjlist(sort_joins=False)
+        sorted_al = w.to_adjlist(sort_joins=True)
+        sv = ["01"] * 4
+        sv.append("04")
+        sv = np.array(sv)
+        usv = np.array(["53", "53", "30", "30", "30"])
+        np.testing.assert_array_equal(unsorted_al.focal.values[:5], usv)
+        np.testing.assert_array_equal(sorted_al.focal.values[:5], sv)
```
martinfleis, line 162:
```suggestion
        np.testing.assert_array_equal(sorted_al.focal.values[:5], sv)
        
    def test_ids(self):
        df = geopandas.read_file(examples.get_path("columbus.dbf")).head()
        df["my_id"] = range(3, len(df) + 3)
        W = weights.Queen.from_dataframe(df, ids="my_id")
        W_adj = W.to_adjlist(drop_islands=True)
        for i in range(3, 8):
            assert i in W_adj.focal
            assert i in W_adj.neighbor
        for i in W_adj.focal:
            assert i in list(range(3, len(df) + 3))
        for i in W_adj.neighbor:
            assert i in list(range(3, len(df) + 3))

    def test_str_ids(self):
        df = geopandas.read_file(examples.get_path("columbus.dbf")).head()
        snakes = ["mamba", "boa", "python", "rattlesnake", "cobra"]
        df["my_str_id"] = snakes
        W = weights.Queen.from_dataframe(df, ids="my_str_id")
        W_adj = W.to_adjlist(drop_islands=True)
        for i in snakes:
            (W_adj.focal == i).any()
            (W_adj.neighbor == i).any()
        for i in W_adj.focal:
            assert i in snakes
        for i in W_adj.neighbor:
            assert i in snakes
```

I haven't noticed this PR and made a fix myself (see the other comment), so at least the tests I made may be included here.

knaaptime:
martin's review echos my last two comments in #507, so do we also want to expose the ids attribute and only conditionally updated the adjlist when they dont match? (as a part of this PR?)

martinfleis:
> do we also want to expose the ids attribute

`self.ids` shall be identical as `neigbors.keys` so not sure if it is worth exposing as a public attribute.

> only conditionally updated the adjlist when they don't match

I'd surely do that (as above)

martinfleis:
Okay, nothing against that.

knaaptime:
or--when converting to WSP, we could stash w.neighbors.keys as a private attribute, then use that when going back `WSP` --> `W` (in which case we just leave everything as-is)

martinfleis:
It seems safer to have an attribute always available than that.

knaaptime:
yeah agree. And looking a bit more closely, we use `id_order` a lot in the sparse stuff, so a straight swap-out for `ids` seems like the best choice there also

knaaptime (approved):
lets go ahead and merge this. I've got [another branch](https://github.com/knaaptime/libpysal/tree/rm_idorder) that includes the enhancements we discussed in this thread, but touches a bunch of other stuff to get rid of `id_order`, so lets get this in and i can rebase that for review

knaaptime:
yeah but it'll be in a separate pr from [this branch](https://github.com/knaaptime/libpysal/tree/rm_idorder?rgh-link-date=2023-01-20T00%3A16%3A47Z) as that one does away with id_order

PR #514, opened 2023-02-11
Reverting w.id_order

sjsrey:
#510 appears to result in several [esda failures](https://github.com/pysal/esda/issues/239) when `shapely>=2.0b1`.

This reverts #510 to make debugging those failures more transparent.


sjsrey:
Closing this, as there were more recent changes #511 that are dependent on #510 so reverting the latter would induce more breakage.

PR #516, opened 2023-02-13
TEST: 3.11 dev coverage

sjsrey:
This adds testing for shapely_dev under 3.11 to help with debugging of:

- https://github.com/pysal/esda/issues/239
- https://github.com/pysal/spopt/issues/346

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -24,6 +24,7 @@
             ci/39.yaml,
             ci/310.yaml,
             ci/310_shapely_dev.yaml
+            ci/311_shapely_dev.yaml
          ]
          include:
            - environment-file: ci/310.yaml
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -23,7 +23,7 @@
             ci/38.yaml,
             ci/39.yaml,
             ci/310.yaml,
-            ci/310_shapely_dev.yaml
+            ci/310_shapely_dev.yaml,
             ci/311_shapely_dev.yaml
          ]
          include:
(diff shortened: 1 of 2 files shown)


jGaboardi (approved):

.github/workflows/unittests.yml
```diff
@@ -23,7 +23,8 @@
             ci/38.yaml,
             ci/39.yaml,
             ci/310.yaml,
-            ci/310_shapely_dev.yaml
+            ci/310_shapely_dev.yaml,
```
martinfleis, line 26: It is no longer `dev` what we're installing.

ci/311_shapely_dev.yaml
```diff
@@ -0,0 +1,32 @@
+name: test
+channels:
+  - conda-forge
+  - conda-forge/label/shapely_dev
```
martinfleis, line 4:
```suggestion
```

not needed, shapely 2 is properly released now

martinfleis (commented):
We can remove the dev flag and channel, shapely 2 is properly out.

knaaptime (approved):
with martin's suggestions, it looks good to me

.github/workflows/unittests.yml
jGaboardi: @martinfleis , so should we rename the environment? Perhaps `shapely_2`?

martinfleis: I'd say it is just a normal latest now.

jGaboardi: Ah, OK. So we can simply remove the `310_shapely_dev.yaml` file and the `310_shapely_dev` test, correct?

martinfleis: I think that if we had one `minimal` with the minimal versions we require and then everything else running on the latest (packages, not Python), it is enough.

PR #521, opened 2023-02-26
CI: revamp CI matrix and set bump minimal requirements

### LINKED ISSUE #520: Update CI envs with a true minimum and dev.
xref https://github.com/pysal/libpysal/pull/516#discussion_r1117983927

As is done in [`momepy`](https://github.com/pysal/momepy/tree/main/ci/envs), we should update the CI environments to have one with all minimum versions supported and one with bleeding edge development versions. Considering the currently support python versions it would look something like:
* `py38-min`
  * All dependencies pinned to exactly the minimum supported versions, which will likely take some trouble shooting because are using [`>=` in the CI env](https://github.com/pysal/libpysal/blob/main/ci/38.yaml).
  * After our actual minimums are proven through CI, we will need to update in the `requirements*.txt`, `environment.yml`, etc.
* `py39`, `py310`,  `py311`
  *  all dependencies with a `>=`  only where absolutely need, which is probably necessary currently
* `py311-dev`
  * same as `py311` above, but installing the bleeding edge versions of some dependencies via `pip git+...`
    * `geopandas`
    * others?

martinfleis:
Closes #520 

Will update the requirements files once I check this actually works.


jGaboardi (approved):
Awesome job on this @martinfleis. Conditional approval based on passing CI. Please merge when ready.

PR #528, opened 2023-05-24
fix indexing issue with adjlist construction

ljwolf:
see [my comment in pysal/esda#245](https://github.com/pysal/esda/pull/245#issuecomment-1561815887) for rationale and the fix implemented here.


jGaboardi:
Possibly related to and fix for https://github.com/pysal/spopt/issues/346

martinfleis (approved):
Thanks!

We really need to get rid of ordering... It is a can of worms and I'm sure this is not the only bug that is in there.

PR #529, opened 2023-05-24
[DO NOT MERGE] Change to the new sparse arrays and see what breaks

ljwolf:
scipy is [moving to a new sparse array type](https://docs.scipy.org/doc/scipy/reference/sparse.html), and eventually deprecating the sparse matrix type. We need to check what breaks when we use it. Right now, I see a few failures locally, so I want to get this online for others to see.


martinfleis:
@ljwolf can we close this given the Graph is using sparse array and there's likely no apetite to make changes in weights?

ljwolf:
yep, fine!

Issue #526, opened 2023-05-12
NumbaDeprecationWarning: The implicit default value nopython is changing


snowman2:
```
libpysal/cg/alpha_shapes.py:165: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
```

PR #535, opened 2023-07-27
Fixing #526

robna:
justification in point 5. 

1. [ ] You have run tests on this submission locally using `pytest` on your changes. Continuous integration will be run on all PRs with [GitHub Actions](https://github.com/pysal/libpysal/blob/master/.github/workflows/unittests.yml), but it is good practice to test changes locally prior to a making a PR.
2. [ ] This pull request is directed to the `pysal/master` branch.
4. [ ] You have [assigned a
5. [x] The justification for this PR is: just doing what the deprecations warning says: `@jit` decorators in alpha_shapes.py that are currently without any arguments now explicitly specify `nopython=False`

### DIFF
--- a/libpysal/cg/alpha_shapes.py
+++ b/libpysal/cg/alpha_shapes.py
@@ -38,7 +38,7 @@ EPS = np.finfo(float).eps
 __all__ = ["alpha_shape", "alpha_shape_auto"]
 
 
-@jit
+@jit(nopython=False)
 def nb_dist(x, y):
     """numba implementation of distance between points `x` and `y`
 
@@ -164,7 +164,7 @@ def r_circumcircle_triangle(a_s, b_s, c_s):
     return r2
 
 
-@jit
+@jit(nopython=False)
 def get_faces(triangle):
     """Extract faces from a single triangle
 
@@ -198,7 +198,7 @@ def get_faces(triangle):
     return faces
 
 
-@jit
+@jit(nopython=False)
 def build_faces(faces, triangles_is, num_triangles, num_faces_single):
     """Build facing triangles
 
@@ -260,7 +260,7 @@ def build_faces(faces, triangles_is, num_triangles, num_faces_single):
     return faces
 
 
-@jit
+@jit(nopython=False)
 def nb_mask_faces(mask, faces):
     """Run over each row in `faces`, if the face in the following row is the
     same, then mark both as False on `mask`


knaaptime:
thanks for this!

robna:
It was apparent that some functions inside alpha_shapes.py already carry a `@jit` decorator with explicit `nopython=True` 

I.e.:
https://github.com/pysal/libpysal/blob/a2e3902a9dea2c9b43ddcf9fbbc8cc8e2c92e7ec/libpysal/cg/alpha_shapes.py#L77-L78
https://github.com/pysal/libpysal/blob/a2e3902a9dea2c9b43ddcf9fbbc8cc8e2c92e7ec/libpysal/cg/alpha_shapes.py#L129-L130
https://github.com/pysal/libpysal/blob/a2e3902a9dea2c9b43ddcf9fbbc8cc8e2c92e7ec/libpysal/cg/alpha_shapes.py#L682-L683

These are not changed by this PR. However, someone more knowledgeable in numba and in this project should review and decide, whether the lines where I now specified `nopython=False` should better also receive a `nopython=True`.

robna:
I added one more commit to the PR regarding another deprecation warning:

voronoi.py imports geopandas (if available) and creates geodataframes, where a `'geometry'` column is manually assigned. This will lead to errors with future versions of geopandas.

```
libpysal/cg/voronoi.py:173: FutureWarning: You are adding a column named 'geometry' to a GeoDataFrame constructed without an active geometry column. Currently, this automatically sets the active geometry column to 'geometry' but in the future that will no longer happen. Instead, either provide geometry to the GeoDataFrame constructor (GeoDataFrame(... geometry=GeoSeries()) or use `set_geometry('geometry')` to explicitly set the active geometry column.
```

Instead, the geometry is now specified in the gdf creation via `gpd.GeoDataFrame(geometry = gpd.GeoSeries( ...`

Please review,
thank you.

martinfleis:
`nopython=True` is recommended and we should use `nopython=False` only if the former fails to compile. ([source](https://numba.pydata.org/numba-doc/latest/user/5minguide.html#what-is-nopython-mode)). So I would try to set them all to True and see if CI complains.

knaaptime:
the default is switching from nopython=False to nopython=True, so these are the ones that raise the warning. In the upcoming versions of numba, the default will switch so the tests will fail for anything that requires python (I played around with this in esda yesterday)

robna:
actually, both, `@jit` (without arguments) and `@jit(nopython=False)` are raising a warning. 

`NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit...`

and

`NumbaDeprecationWarning: The keyword argument 'nopython=False' was supplied. From Numba 0.59.0 the default is...`

But also for me, no matter what way I set them, there are no severe problems. Just the warnings.

knaaptime (approved):
explicit is better than implicit, so we'll want to include `nopython=False` anyway. I vote to merge

PR #536, opened 2023-07-28
Test graph class

martinfleis:
This is just a start, no need to merge yet. I'll continue later today a bit and then after the weekend.

### DIFF
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -10,10 +10,11 @@ def _neighbor_dict_to_edges(neighbors, weights=None):
     that the any self-loops have a weight of zero.
     """
     idxs = pd.Series(neighbors).explode()
+    idxs = idxs.fillna(pd.Series(idxs.index, index=idxs.index))  # self-loops
     heads, tails = idxs.index.values, idxs.values
     tails = tails.astype(heads.dtype)
     if weights is not None:
-        data_array = pd.Series(weights).explode().values
+        data_array = pd.Series(weights).explode().fillna(0).values
         if not pd.api.types.is_numeric_dtype(data_array):
             data_array = pd.to_numeric(data_array)
     else:
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -705,6 +705,9 @@ class Graph(_Set_Mixin):
                 i \in N_j \ \& \ j \in N_i
 
             where :math:`N_j` is the set of neighbors for :math:`j`.
+            I.e. ``True`` requires equality of the weight to consider
+            two links equal, ``False`` requires only a presence of a link
+            with a non-zero weight.
 
         Returns
         -------
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1,4 +1,5 @@
 from functools import cached_property
+import math
 
 import numpy as np
 import pandas as pd
@@ -572,7 +573,7 @@ class Graph(_Set_Mixin):
         transformation = transformation.upper()
 
         if self.transformation == transformation:
-            return self
+            return self.copy()
 
         if transformation == "R":
             standardized = (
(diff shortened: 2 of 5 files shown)


martinfleis:
I have included `to_W` and `from_W` in this PR to allow easy roundtripping. We can use `to_W` in downstream for an easy compatibility layer.

martinfleis:
This is now ready for a review and merge. It contains test suite for most of the contents of `base.py`. It does not include tests of any of the `build_*` constructors.

It also contains a number of fixes I've found through tests and a few new methods I needed (W I/O. copy).

libpysal/graph/_utils.py
```diff
@@ -10,10 +10,11 @@ def _neighbor_dict_to_edges(neighbors, weights=None):
     that the any self-loops have a weight of zero.
     """
     idxs = pd.Series(neighbors).explode()
+    idxs = idxs.fillna(pd.Series(idxs.index, index=idxs.index))  # self-loops
     heads, tails = idxs.index.values, idxs.values
     tails = tails.astype(heads.dtype)
     if weights is not None:
-        data_array = pd.Series(weights).explode().values
+        data_array = pd.Series(weights).explode().fillna(0).values
```
jGaboardi, line 17: I am guessing since this is a very internal function that hard coding a `0` here is the only acceptable value for filling `NA`?

libpysal/graph/base.py
```diff
@@ -47,25 +114,69 @@ def from_old_w(cls, w):
...
 
         Parameters
         ----------
         sparse : scipy.sparse array
+            sparse representation of a graph
+        focal_ids : list-like, default None
+            list-like of ids for focal geometries that is mappable to
+            positions from sparse. If None, the positions are used as labels.
+        neighbor_ids : list-like, default None
+            list-like of ids for focal geometries that is mappable to
```
jGaboardi, line 161: Should this one read `focal geometries` or `neighbor geometries`?

libpysal/graph/base.py
```diff
@@ -471,17 +594,17 @@ def transform(self, transformation):
...
-                    / np.sqrt(self._adjacency.weight.groupby(level=0).transform("sum"))
-                )
-                .fillna(0)
-                .values
-            )  # island comes as NaN -> 0
+            s = self._adjacency.weight.groupby(level=0).transform(
+                lambda group: group / math.sqrt((group**2).sum())
+            )
+            nQ = self.n / s.sum()
+            standardized = (s * nQ).fillna(0).values  # isolate comes as NaN -> 0
```
jGaboardi, line 601: Same here for hard coding a `0` for filling `NA`. Would there never be any chance a user would want/need to use another value for filling `NA`?

libpysal/graph/base.py
```diff
@@ -582,6 +705,9 @@ def asymmetry(self, intrinsic=True):
                 i \in N_j \ \& \ j \in N_i
 
             where :math:`N_j` is the set of neighbors for :math:`j`.
+            I.e. ``True`` requires equality of the weight to consider
```
jGaboardi, line 708: Very small nitpick, but I think this should be "e.g." (could be wrong though).

libpysal/graph/tests/test_base.py
```diff
@@ -0,0 +1,718 @@
...
+        G = graph.Graph.from_dicts(self.neighbor_dict_int)
+        pd.testing.assert_frame_equal(G._adjacency, self.adjacency_int_binary)
+
+        G = graph.Graph.from_dicts(self.neighbor_dict_str)
+        pd.testing.assert_frame_equal(G._adjacency, self.adjacency_str_binary)
+
+    @pytest.mark.parametrize("y", [3, 5])
+    @pytest.mark.parametrize("rook", [True, False])
+    @pytest.mark.parametrize("id_type", ["int", "str"])
+    def test_from_dicts_via_W(self, y, id_type, rook):
```
jGaboardi, line 320: Very minor nit: Perhaps we should swap the `"rook"` and `"id_type"` markers to match the order of the positional arguments in `test_from_dicts_via_W`?

jGaboardi (changes requested):
Mostly minor stuff. Very good work. I am guessing yall decided to swap `islands` terminology to `isolates` in one of the recent sprints?

jGaboardi:
Curious about the Ubuntu 3.8 and Windows 3.11 failures.

libpysal/graph/_utils.py
martinfleis: Always 0. Weights of an isolate (island) are `[]` which gets converted to missing by `explode`. We always need to encode that with 0.

libpysal/graph/base.py
martinfleis: Nope, always 0. When a weight is 0, these ops result in it being np.nan at one point but all we want is to keep a 0. For both isolates and zero-weight neighbours.

martinfleis:
>  I am guessing yall decided to swap islands terminology to isolates in one of the recent sprints?

As per @ljwolf's suggestion to match the [ecosystem](https://networkx.org/documentation/stable/reference/algorithms/isolates.html#module-networkx.algorithms.isolate).


> Curious about the Ubuntu 3.8 and Windows 3.11 failures.

They're not related to these changes (apart from that one in 3.8 I'll look into) so I ignored them for now.

martinfleis:
There's also some flaky-ness with pytest-xdist but none of that related to this PR, so I suppose nothing stops from merging this.

martinfleis:
Merging in the spirit of fact-paced development :)

PR #537, opened 2023-07-28
Geographs _triangulation weights 

sjsrey:
This has complete working drafts of the triangulation weights.

## To do

- [ ] insert coincident points check
- [ ] tests
- [ ] use case notebook illustrating coincident point policy and new api

### DIFF
--- a/libpysal/graph/_triangulation.py
+++ b/libpysal/graph/_triangulation.py
@@ -9,16 +9,19 @@ from ._kernel import _kernel_functions
 from ._utils import _validate_geometry_input
 from .base import Graph
 
+from libpysal.cg import voronoi_frames
+
 try:
     from numba import njit
 except ModuleNotFoundError:
     from libpysal.common import jit as njit
 
-_VALID_GEOMETRY_TYPES = ("Point",)
+_VALID_GEOMETRY_TYPES = ("Point")
 
 __author__ = """"
 Levi John Wolf (levi.john.wolf@gmail.com)
 Martin Fleischmann (martin@martinfleischmann.net)
+Serge Rey (sjsrey@gmail.com)
 """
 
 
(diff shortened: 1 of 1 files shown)


libpysal/graph/_triangulation.py
```diff
@@ -269,27 +282,25 @@ def voronoi(
...
     cells, generators = voronoi_frames(coordinates, clip=clip)
-    if contiguity_type == "vertex":
+    if contiguity_type == "v":
         w = _vertex_set_intersection(cells, ids=ids)
-    elif contiguity_type == "queen":
+    elif contiguity_type == "q":
         w = _queen(cells, ids=ids)
-    elif contiguity_type == "rook":
+    elif contiguity_type == "r":
         w = _rook(cells, ids=ids)
```
martinfleis, line 294: This is unnecessary. We should have a keyword that controls if it is rook or queen, ideally consistent with contiguity builders which is currently `rook=True/False` (but happy to have a discussion about better options), and pass it to `_vertex_set_intersection`. We do not need to use _queen and _rook in here as they bring no benefit whatsoever. The geometry coming from voronoi is always topologically correct, so we can always depend on faster vertex-based method over shapely's predicates.

libpysal/graph/_triangulation.py
```diff
@@ -94,6 +101,10 @@ def delaunay(coordinates, ids=None, bandwidth=numpy.inf, kernel="boxcar"):
         axis=1
     ).squeeze() ** 0.5
     weights = _kernel_functions[kernel](distances, bandwidth)
+    # check for coincident points which result in
+    # dropped points from the triangulation and the
+    # misalignment of the weights and the attribute array
+
     return Graph.from_arrays(head, tail, weights)
```
martinfleis, line 108: We should probably discuss if these private methods should return the Graph or the arrays. In contiguity, we always  return `heads, tails, weights` tuple and Graph itself gets constructed in `Graph.build_contiguity`. It, in practice, does not matter much but we should be consistent. And given we want to push users towards `Graph.build_*` path, I suggest returning the tuple of arrays in all these constructor functions.

libpysal/graph/_triangulation.py
```diff
@@ -305,6 +316,8 @@ def _edges_from_simplices(simplices):
     that are then converted into the six non-directed edges for
     each simplex.
     """
+    dupes = {}
```
martinfleis, line 319:
```suggestion
```

not used

martinfleis (commented):
A few notes in the code. 

I would also merge this sooner than later and leave the to-do tasks listed in the first post for follow-up PRs to speed things up.

martinfleis:
I'll merge this and open a PR on top of these changes. We can do the to-do in the follow-ups.

PR #541, opened 2023-08-03
TST: fix tests of the Graph class

martinfleis:
There are some flaky tests and I suppose that something does not play nice with pytest-xdist, so I disabled it now.

Some linting also happened.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -14,7 +14,7 @@
      runs-on: ${{ matrix.os }}
      env:
        FETCH_EXAMPLES: python -c 'import libpysal; libpysal.examples.fetch_all()'
-       RUN_TEST: pytest -v -n auto libpysal --cov libpysal --cov-config .coveragerc --cov-report xml --color yes --cov-append --cov-report term-missing
+       RUN_TEST: pytest -v libpysal --cov libpysal --cov-config .coveragerc --cov-report xml --color yes --cov-append --cov-report term-missing
      #timeout-minutes: 25
      strategy:
        matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/unittests.yml
```diff
@@ -14,7 +14,7 @@
      runs-on: ${{ matrix.os }}
      env:
        FETCH_EXAMPLES: python -c 'import libpysal; libpysal.examples.fetch_all()'
-       RUN_TEST: pytest -v -n auto libpysal --cov libpysal --cov-config .coveragerc --cov-report xml --color yes --cov-append --cov-report term-missing
```
martinfleis, line 17: This fixed those issue. I suppose that the reason for them in the first place was download of datasets from geodatasets that was attempted by both workers at the same time, which resulted in invalid files.

@jGaboardi are we fine with disabling pytest-xdist? at least temporarily? Or would you rather rewrite those tests in some way to avoid the issue?

jGaboardi: Yes, let's temporarily disable `pytest-xdist`. We can re-evaluate as time permits, but it does seem like much of the value with `pytest-xdist` is for local testing and VMs where there are 4 or more cores. The two cores here are providing some speedup, but also headaches like we are seeing here...

jGaboardi:
LOL at the [bba0557](https://github.com/pysal/libpysal/pull/541/commits/bba05574ccde079ae7150b66f08adcfe2233cbd8) commit message.

.github/workflows/unittests.yml
```diff
@@ -14,7 +14,7 @@
...
+       RUN_TEST: pytest -v libpysal --cov libpysal --cov-config .coveragerc --cov-report xml --color yes --cov-append --cov-report term-missing
```
jGaboardi, line 17: Let's maybe keep the `-n` parameter and simply set it as `1`? That way we can keep the logic in there *and* `pytest-xdist` "disabled"

jGaboardi (approved):

Issue #540, opened 2023-07-30
BUG: weights.W.asymmetry does not support custom ids


martinfleis:
```py
>>> neighbors={"a":["b","c","d"], "b":["b","c","d"], "c":["a","b"], "d":["a","b"]}
>>> weights_d={"a":[1,1,1], "b":[1,1,1], "c":[1,1], "d":[1,1]}
>>> w=weights.W(neighbors,weights_d)
>>> w.asymmetry()
[(0, 1), (1, 0)]
```

This should be a ``list`` of ``(i,j)`` tuples, as per docstring but it is capturing integer ids coming from sparse.

It should be this:

```
[("a", "b"), ("b", "a")]
```

jGaboardi:
Locally I've added the following to after [line 1280](https://github.com/pysal/libpysal/blob/e743c48c8bd3b74b423202ef5bb9b7316e9ac14e/libpysal/weights/weights.py#L1280):

```python
if isinstance(self.id_order[0], str):
    id_order_array = np.asarray(self.id_order)
    ids = tuple(id_order_array[ids[i]] for i in (0, 1))
```

This returns the correct ID labels if the labels are strings. Another thing we may consider though is if IDs are not strings, but numeric values that are not a 0-indexed range...

jGaboardi:
Tests pass locally when simply leaving out the conditional:

```python
id_order_array = np.asarray(self.id_order)
ids = np.nonzero(wd)
ids = tuple(id_order_array[ids[i]] for i in (0, 1))
```

martinfleis:
> Another thing we may consider though is if IDs are not strings, but numeric values that are not a 0-indexed range...

That is another valid case we should support. This one can be especially confusing. Your W is indexed by int, you get back int but it refers to position, not the value.

jGaboardi:
2 more complications with indices that are not both (a) monotonically increasing integers and (b) 0-based:

* With the above array-based solution, if there are mixed-numeric types all values will be coerced to floats. This is probably a super edge case, but it exists nonetheless.
* The sorting that is happening in line [L1285](https://github.com/pysal/libpysal/blob/e743c48c8bd3b74b423202ef5bb9b7316e9ac14e/libpysal/weights/weights.py#L1285) may not give the intended result (and does the even matter?)

jGaboardi:
[...]

PR #544, opened 2023-08-07
expand graph.Graph constructors and other functionality

martinfleis:
Very much WIP but wanted to open this for discussion. This partially handles isolates but not always. Distance band and knn are used to compute distance matrix that is then (optionally) passed via kernel. 

@ljwolf I am not sure if this is the implementation you had in mind, but this seems to also work well.

includes:

- `Graph.build_distance_band`
- reworked kernels

edit: Also includes:

- `Graph.build_block_contiguity()`
- Parquet IO as `graph.read_parquet` and `Graph.to_parquet` mirroring (geo)pandas
- `Graph.lag()`
- `__getitem__`


libpysal/graph/_kernel.py
```diff
@@ -157,6 +140,44 @@ def _kernel(
...
+    return D
+
+
+def _distance_band(coordinates, threshold, ids=None):
+    coordinates, ids, _ = _validate_geometry_input(
+        coordinates, ids=ids, valid_geometry_types=_VALID_GEOMETRY_TYPES
+    )
+    tree = spatial.KDTree(coordinates)
+    dist = tree.sparse_distance_matrix(tree, threshold, output_type="ndarray")
+    return sparse.csr_array((dist["v"], (dist["i"], dist["j"])))
```
martinfleis, line 176: This is a trick how to get a sparse array instead of a sparse matrix out of `sparse_distance_matrix`. I suppose that scipy will have `sparse_distance_array` eventually but this is fairly cheap anyway.

libpysal/graph/base.py
```diff
@@ -497,6 +497,87 @@ def build_triangulation(
...
+
+        # drop diagnoal
+        counts = adjacency.index.value_counts()
+        no_isolates = counts[counts > 1]
+        adjacency = adjacency[
+            ~(
+                adjacency.index.isin(no_isolates.index)
+                & (adjacency.index == adjacency.neighbor)
+            )
+        ]
```
martinfleis, line 578: These bits should ideally be handled within `_kernel` and properly. This does not always work.

ljwolf:
Yep, this is exactly what I had in mind!

martinfleis:
This one is ready for review. Not everything properly works (isolates in kernel-based stuff is sketchy in some and plain wrong  in other cases) but I'd leave that for follow-ups. I've left TODO comments for those.

I will be leaving for a week on Saturday and will not be able to respond to anything, so it would be best if this is merged by Saturday morning (my time) to ensure I am not a blocker.

Tests are mostly checking dtypes and expected shapes but not necessarily values and all possible paths. That will need to be done as follow-ups, especially for constructors.

libpysal/graph/base.py
```diff
@@ -497,6 +520,109 @@ def build_triangulation(
...
+        cls, data, threshold, binary=True, alpha=-1.0, kernel=None, bandwidth=None
+    ):
+        """Generate Graph from geometry based on a distance band
+
+        Parameters
+        ----------
+        data : numpy.ndarray, geopandas.GeoSeries, geopandas.GeoDataFrame
+            geometries containing locations to compute the
+            delaunay triangulation. If a geopandas object with Point
+            geoemtry is provided, the .geometry attribute is used. If a numpy.ndarray
```
jGaboardi, line 534:
```suggestion
            geometry is provided, the .geometry attribute is used. If a numpy.ndarray
```

libpysal/graph/base.py
```diff
@@ -497,6 +520,109 @@ def build_triangulation(
...
+            with shapely geoemtry is used, then the coordinates are extracted and used.
```
jGaboardi, line 535:
```suggestion
            with shapely geometry is used, then the coordinates are extracted and used.
```

libpysal/graph/base.py
```diff
@@ -497,6 +520,109 @@ def build_triangulation(
...
+                kernel=lambda distances, alpha: np.power(distances, alpha),
+                metric="precomputed",
+                ids=ids,
+                bandwidth=alpha,
+            )
+        sp.setdiag(0)
+
+        adjacency = cls.from_sparse(sp, ids)._adjacency
+
+        # drop diagnoal
```
jGaboardi, line 591:
```suggestion
        # drop diagonal
```

jGaboardi (commented):
Looks good as far as I can tell, considering your [comments above](https://github.com/pysal/libpysal/pull/544#issuecomment-1673737505).

libpysal/graph/_kernel.py
```diff
@@ -156,6 +139,51 @@ def _kernel(
...
+        coordinates = numpy.fliplr(numpy.deg2rad(coordinates))
+    query = _prepare_tree_query(coordinates, metric, p=p)
+    D_linear, ixs = query(coordinates, k=k + 1)
+    self_ix, neighbor_ix = ixs[:, 0], ixs[:, 1:]
+    D_linear = D_linear[:, 1:]
+    self_ix_flat = numpy.repeat(self_ix, k)
+    neighbor_ix_flat = neighbor_ix.flatten()
+    D_linear_flat = D_linear.flatten()
+    if metric == "haversine":
+        D_linear_flat * 6371  # express haversine distances in kilometers
```
jGaboardi, line 165: Should we add a note to the docstrings of exposed functions that the returned `haversine` distance is in kilometers (as opposed to meters)?

libpysal/graph/base.py
```diff
@@ -497,6 +520,109 @@ def build_triangulation(
...
+            )
+        else:
+            sp, ids = _kernel(
+                dist,
+                kernel=lambda distances, alpha: np.power(distances, alpha),
+                metric="precomputed",
+                ids=ids,
+                bandwidth=alpha,
+            )
+        sp.setdiag(0)
```
ljwolf, line 587: This is not necessarily valid, but I can understand why this is here. Many kernel weights require self-weighting. It would be good to provide a way to control this.

martinfleis: This was likely related to some isolate handling. That is still to-do as a follow up so I'll just add a note here to reconsider this.

libpysal/graph/base.py
```diff
@@ -598,18 +724,18 @@ def sparse(self):
             (
                 self._adjacency.weight.values,
                 (
-                    self._adjacency.index.map(self.id2i),
-                    self._adjacency.neighbor.map(self.id2i),
+                    self._adjacency.index.map(self._id2i),
+                    self._adjacency.neighbor.map(self._id2i),
```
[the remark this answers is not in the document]
martinfleis: yes. do you have a better suggestion?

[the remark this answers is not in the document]
ljwolf: No no! I was just wondering, given I'll need a way to deal with this when re-ordering the adjacency list after expanding cliques.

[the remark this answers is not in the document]
jGaboardi: 2 things here from my end:
* Regarding the public vs. private discussion for `id2i`, was this talked about anywhere else besides [here](https://github.com/pysal/libpysal/pull/542#issuecomment-1668456231)? Since `id2i` is currently exposed in the `W` implementation, I am curious what the argument is for making it private in the new `Graph` implementation.
* Also, I wanted to bring up the idea again to discuss the potential usefulness of an `i2id` property (or `_i2id` if we stick with the current private schema). Currently that translated lookup is only used within [`asymmetry`](https://github.com/pysal/libpysal/blob/e579820a444ac70395e079b99601e636c6728ecc/libpysal/graph/base.py#L808), but I am wondering if anyone can think of other situations where it might be helpful? If we will be using a translated lookup anywhere else, then another property  is surely warranted IMHO. 
  * xref: https://github.com/pysal/libpysal/issues/540#issuecomment-1667093292

ljwolf:
I would like to merge this so that I can rebase #548... every triangulation can pass through kernel, so we need the kernel implementation stabilised before I can finish up the triangulation stuff.

@jGaboardi your review doesn't have any further outstanding requests, right? just questions about `id2i`?  If so, can I merge this?

martinfleis:
@ljwolf Go ahead. We can open issues for API discussions.

PR #548, opened 2023-08-10
Intercept Coincident points

ljwolf:
This is my start at intercepting coincident points. I still need to
- [x] implement the jitter solution
- [ ] re-order the `new_adj` table according to the derived original ids
- [x] (possibly) implement these checks in `_triangulation.py` as a wrapper function? I think we could do this easily by intercepting the input & output of the wrapped function. 
- [ ] ~implement for knn weights~
- [x] tests

I will hold off on implementing these for `_distance.py` until @martinfleis's changes land and we decide if the decorator strategy is the way to go.


libpysal/graph/_utils.py
```diff
@@ -22,6 +63,24 @@ def _neighbor_dict_to_edges(neighbors, weights=None):
...
+def _validate_coincident(geoms):
+    """
+    Identify coincident points and create a look-up table for the coincident geometries. 
+    """
+    valid_coincident_geom_types = set(("Point",))
+    if not set(geoms.geom_type) <= valid_coincident_geom_types:
+        raise ValueError(
+            f"coindicence checks are only well-defined for geom_types: {valid_coincident_geom_types}"
+            )
+    wkb = geoms.to_frame('geometry').assign(hash = geoms.apply(lambda x: x.wkb)).reset_index(names='index')
```
martinfleis, line 75: You don't need wkb, shapely 2 geometries are hashable so both `geoms.nunique()` and `groupby('geometry')` will work directly.

If you want WKB for some reason, use `to_wkb` to make the conversion vectorized, rather than lambda.

libpysal/graph/_utils.py
```diff
@@ -22,6 +63,24 @@ def _neighbor_dict_to_edges(neighbors, weights=None):
...
+    if len(wkb) > len(wkb.hash.unique()):
```
martinfleis, line 76:
```suggestion
    if len(wkb) > wkb.hash.nunique(): 
```

libpysal/graph/_utils.py
knaaptime: i've been using `groupby('geometry')` of late... that rocks.

ljwolf: I was using the wkb as a "string" identifier for the ids of the points, since I didn't want to assume that using the geometries themselves as ids would work... but I suppose that's a robust test of our indexing code...

ljwolf:
I've implemented the jitter solution using displacements in a circle with a radius of the `resolution` of the input dtype (if a float), or the resolution of `float32` (`1e-6`) if the input are integers or float168. This should be the smallest possible movement required to de-dupe points. 

After some simulations, I don't see any improvement in the agreement of `_delaunay()` for smaller displacements (up to `1e-22`): for anything at or this `resolution` value, the agreement between any two jittered tessellations is about 68%.... not very stable. 


```python
import geopandas, geodatasets, numpy
from libpysal.graph._triangulation import _delaunay, _validate_coincident
from libpysal.graph._utils import _build_coincidence_lookup
from libpysal.graph.base import Graph
from tqdm import trange

gdf = (
    geopandas.read_file(geodatasets.get_path("geoda groceries"))
    .explode(index_parts=False)
    .reset_index(drop=True)
)

coincident = geopandas.pd.concat((gdf.head(5), gdf.head(2), gdf.head(1)), axis=0).reset_index(drop=True)
coincident.index = (
    coincident.index.astype(str).str.rjust(3, '0') 
    + '-' 
    + coincident.Chain.str.replace(" ", "").str.replace("/", "")
    )

jgraph = _delaunay(coincident.geometry, coincident='jitter')

def iou(a,b, normalized=True):
    n_together = len(a.intersection(b))
    if normalized:
        return n_together / len(a.union(b))
    else:
        return n_together

reps = [set(zip(*jgraph[0:2]))] + [set(zip(*_delaunay(coincident.geometry, coincident='jitter')[0:2])) for _ in trange(999)]
sims_lt = [iou(reps[i], reps[j]) for i in trange(1000) for j in range(i+1)]
sims = numpy.empty((1000,1000))
sims[*numpy.tril_indices(1000)] = sims_lt
sims += sims.T
numpy.fill_diagonal(sims, 1)
```

libpysal/graph/_utils.py
ljwolf: This is now quite a bit simplified by using the geometry alone and using the synthetic ids from the groupby.

ljwolf:
Great that we're doing `kernel` as a thin class. This makes basically all of `_triangulate.py` possible to implement as *just* building a sparse distance matrix. I have very good feelings about this API design!

ljwolf:
Not sure why this is replaying all of Martin's commits from #544? We can wait to rebase I think.

martinfleis:
```sh
git fetch
git merge upstream/geographs
```

does not resolve that? How did you update your branch after we merged the other PR?

ljwolf:
I did an interactive rebase to solve some commit conflicts, so I think that might have done it. I've tried the merge commit again now, so we'll see if that fixes it.

ljwolf:
Worst comes to worst, I'll close this, create a new branch from `upstream/master`, and cherry-pick the relevant commits. 

I think the issue is that the rebase put your commits in the middle of the commit chain.

martinfleis:
looks okay now

libpysal/graph/_kernel.py
```diff
@@ -61,6 +60,7 @@ def _kernel(
     k=None,
     ids=None,
     p=2,
+    taper=False,
```
ljwolf, line 63: This should default to true.

ljwolf:
For the ordering of input `from_arrays()`, we need to document the ordering that happens implicitly from `pandas.concat((focal, neighbor)).unique()`. It's not sorting, but it's useful to document in the method so that we know where we might have to re-arrange the `Graph` based on node input (e.g. from a dataframe) that comes through `from_arrays()`, like most of `_triangulation.py` as well as `_contiguity.py`

libpysal/graph/_utils.py
```diff
@@ -2,6 +2,73 @@
...
+    longitude coordinates, this will distort according to a plateé carree
+    prjection, jittering slightly more in the x direction than the y direction. 
+    """
+    rng = np.random.default_rng(seed=seed)
+    dtype = coordinates.dtype
+    if dtype not in (np.float32, np.float64):
+        # jittering requires us to cast ints to float
+        # and the rng.random generator only works with float32 and float64
+        dtype = np.float32
+    # the resolultion is the approximate difference between two floats
```
jGaboardi, line 23:
```suggestion
    # the resolution is the approximate difference between two floats
```

libpysal/graph/_utils.py
```diff
@@ -2,6 +2,73 @@
...
 import shapely
+from itertools import permutations
+
+def _jitter_geoms(coordinates, geoms, seed=None):
+    """
+    Jitter geometries based on the smallest required movements to induce 
+    uniqueness. For each point, this samples a radius and angle uniformly 
+    at random from the unit circle, rescales it to a circle of values that
+    are extremely small relative to the precision of the input, and
+    then and displaces the point. For a non-euclidean geometry, like latitude
```
jGaboardi, line 13:
```suggestion
    then displaces the point. For a non-euclidean geometry, like latitude
```

libpysal/graph/_utils.py
```diff
@@ -2,6 +2,73 @@
...
+    longitude coordinates, this will distort according to a plateé carree
+    prjection, jittering slightly more in the x direction than the y direction.
```
jGaboardi, line 15:
```suggestion
    projection, jittering slightly more in the x direction than the y direction. 
```

libpysal/graph/_triangulation.py
```diff
@@ -23,8 +29,75 @@
...
+                    f"There are {len(coincident_lut)} "
+                    f"unique locations in the dataset, but {len(geoms)} observations."
+                    "This means there are multiple points in the same location, which"
+                    " is undefined for this graph type. To address this issue, consider setting "
+                    " `coincident='clique' or consult the documentation about coincident points."
+                )
+            elif coincident == "jitter":
+                coordinates, geoms = _jitter_geoms(coordinates, geoms)
+            elif coincident == "clique":
+                input_coordinates, input_ids, input_geoms = coordinates, ids, geoms
```
jGaboardi, line 54: * Are `input_coordinates`, `input_ids`, and `input_geoms` actually used anywhere?
* Should `_induce_cliques()` be used within this chunk or only down in L92?

libpysal/graph/_utils.py
```diff
@@ -23,6 +90,20 @@ def _neighbor_dict_to_edges(neighbors, weights=None):
     return heads, tails, data_array
 
 
+def _build_coincidence_lookup(geoms):
+    """
+    Identify coincident points and create a look-up table for the coincident geometries. 
+    """
+    valid_coincident_geom_types = set(("Point",))
```
jGaboardi, line 97: Does this need to be a `set`? Would `["Point"]` work just as well?

libpysal/graph/_triangulation.py
```diff
@@ -23,8 +29,75 @@
...
+            distances = _vec_euclidean_distances(coordinates[heads_ix], coordinates[tails_ix]).squeeze()
+            sparse_D = sparse.csc_array((distances, (heads_ix, tails_ix)))
+            if bandwidth == "auto":
+                bandwidth = _optimize_bandwidth(sparse_D, kernel)
+            _k = _kernel(
+                sparse_D, 
+                metric='precomputed', 
+                kernel=kernel, 
+                bandwidth=bandwidth, 
+                taper=False
```
jGaboardi, line 78: Should `taper` be passed into `tri_with_validation()` as a keyword argument? And added to each constructor call within `Graph.build_triangulation()` (plus a docstring entry there?)

libpysal/graph/_triangulation.py
```diff
@@ -23,8 +29,75 @@
...
+            weights = _k.data
+        adjtable = pandas.DataFrame.from_dict(
+            dict(
+                focal = heads, neighbor = tails, weight = weights
+            )
+        )
+
+        if (n_coincident > 0) & (coincident == "clique"):
+            # note that the kernel is only used to compute a fill value for the clique. 
+            # in the case of the voronoi weights. Using boxcar with an infinite bandwidht
```
jGaboardi, line 89:
```suggestion
            # in the case of the voronoi weights. Using boxcar with an infinite bandwidth
```

libpysal/graph/base.py
```diff
@@ -445,6 +445,7 @@ def build_triangulation(
         kernel="boxcar",
         clip="extent",
         rook=True,
+        coincident='raise'
```
jGaboardi, line 448: Here's starter docstring entry for `coincident`:

```
coincident : str, optional (default "raise")
    Method for handling coincident points. Options include
    ``'raise'`` (raising an exception when coincident points are present),
    ``'jitter'`` (randomly displace coincident points to produce uniqueness), and
    ``'clique'`` (induce clues and fully-connected sub cliques for coincident points).
```

libpysal/graph/_triangulation.py
ljwolf: It doesn't hurt, but I would not. Users who ask for a kernel weighted triangulation intend for the triangulation links weighted by a kernel function. If the kernel is zero for a link, the *triangulation* should still preserve those links since the triangulation is the underlying graph.

Regardless, if taper is included as an argument, this still must be  False here, otherwise the kernel output can become misaligned to the triangulation.

ljwolf: Actually, we should just be able to pass taper through and return the kernel output. That would probably be safest for the future too. 

My issue earlier was that I was keeping the heads/tails from the triangulation and stacking the weights back onto them. The size of _k.data changes depending on taper, but we can just use the kernel output directly with taper as a keyword, set to false by default

knaaptime:
what's the authorship model we're shooting for?

ljwolf:
Happy to do whatever most devs find useful! I was thinking we would add to a module-level author string, like in current `libpysal`?

knaaptime:
>Happy to do whatever most devs find useful! I was thinking we would add to a module-level author string, like in current libpysal?

cool with me, i just noticed things are pretty inconsistent. Some files have authors, others dont, so wasnt sure whether we had a convention. Tbh, it always feels mildly awkward to me when an open-source file has an authorship line. On the one hand, the "original author" should definitely get credit for the first draft. On the other hand, since the codebase is collectively maintained and edited, what constitutes a contribution *large enough* that someone would feel comfortable adding their name to `__authors__`? When someone opens a PR that adds a new function to a file that comprises over 50% of that file's lines, but the functionality is minimal, is that person an author?  At some level, it feels like the git log is the fundamental authorship record, but im curious how others feel


completely unrelated: It just occurred to me that we can induce cliques with voronoi graphs that have coincident points. Maybe that `_induce_cliques` method is already applied, but since we'd only discussed in the context of knn and distanceband, thought i'd raise

ljwolf:
> completely unrelated 

Indeed, this currently *only* deals with the triangulations, including voronoi. I need slightly different logic to handle KNN, since the interception depends on the value of k, but all triangulation weights have to get handled, even if there's only one coincident pair.

ljwolf:
@sjsrey the example I'm using for testing coincident points is the following: 

```python
import geopandas, geodatasets, numpy
from libpysal.graph._triangulation import _delaunay, _validate_coincident
from libpysal.graph._kernel import _knn
from libpysal.graph._utils import _build_coincidence_lookup
from libpysal.graph.base import Graph
from tqdm import trange

gdf = (
    geopandas.read_file(geodatasets.get_path("geoda groceries"))
    .explode(index_parts=False)
    .reset_index(drop=True)
)

coincident = geopandas.pd.concat((gdf.head(5), gdf.head(2), gdf.head(1)), axis=0).reset_index(drop=True)
coincident.index = (
    coincident.index.astype(str).str.rjust(3, '0') 
    + '-' 
    + coincident.Chain.str.replace(" ", "").str.replace("/", "")
    )

max_n, lut = _build_coincidence_lookup(coincident.geometry)

dgraph = _delaunay(lut.geometry)
graph = _delaunay(coincident.geometry, coincident='clique')

dgraph = Graph.from_arrays(*dgraph)
graph = Graph.from_arrays(*graph)

jgraph = _delaunay(coincident.geometry, coincident='jitter')

# raises
knn = _knn(coincident.geometry)
knn = _knn(coincident.geometry, coincident='jitter')
knn = _knn(coincident.geometry, coincident='clique')
```

libpysal/graph/_utils.py
ljwolf: It should remain set because we're doing a subset comparison immediately afterwards, checking whether the input geometry types are a subset of the supported geometry types. 

In theory, coincidence is well defined for all geometry types, so I figured this function would evolve if there was a need to have more geometry types.

libpysal/graph/_triangulation.py
ljwolf: No, `input_coordinates`, `input_ids`, and `input_geoms` aren't currently used again, but we probably will need at least `input_ids` later when we implement re-ordering the output. Not done yet.

martinfleis:
Just for reference, this is my solution to sorting.

```py
df = pd.DataFrame({"focal": heads, "neighbor": tails})
# df = df.sample(frac=1).reset_index(drop=True) 
sorted_index = df.applymap(coincident.index.unique().tolist().index).sort_values(["focal", "neighbor"]).index
return heads[sorted_index], tails[sorted_index], weights[sorted_index]
```

martinfleis:
@ljwolf have you intentionally removed all the kernel functionality from triangulation here?

ljwolf:
>it's all in the decorator

yep. I think a similar approach would be useful for KNN/distance band? rather than doing knn truncation in kernel for precomputed weights, we make knn and distance band similar to the triangulators: ways to build a generic distance matrix, which then gets passed to kernel for weighting. 

it keeps the behaviour very consistent, and reduces the amount of unique class/constructor-based code.

knaaptime:
thats ideal because it allows 

`Graph.build_kernel(gdf, threshold=2000, kernel='gaussian')` and

`Graph.from_sparse(distance_matrix).build_kernel(threshold=2000, kernel='gaussian')`, 

where distance_matrix is a travel time matrix from routingpy or pandana

ljwolf:
Indeed, already can with _kernel(distance_matrix, metric="precomputed", kernel="gaussian") so I think Graph.build_kernel(distance_matrix, metric="precomputed", kernel="gaussian") would work. 

The simplification would be that kernel *wouldnt* take a "k" argument, because that is a knn-style operation.

martinfleis:
I am not sure if I am a big fan of this decorator approach. It does not result in the most legible code.

> `Graph.from_sparse(distance_matrix).build_kernel(threshold=2000, kernel='gaussian')`

This would result in inconsistent API. `build_kernel` builds Graph from external object, not `self`. What are you looking for is `transform()` which should support a custom kernel. What we have there now is a subset of kernels anyway.

knaaptime:
got it, so the appropriate way is 

`Graph.build_kernel(distance_matrix, metric="precomputed", threshold=2000, kernel="gaussian")`

equivalent to 

`Graph.build_kernel(gdf, threshold=2000, kernel="gaussian")`


building kernel from the sparse versus building kernel from the implied distances in the gdf.coordinates

martinfleis:
Yes, that is how it currently works. But I can also imagine

```py
Graph.from_sparse(distance_matrix).transform(transformation="kernel", threshold=2000, kernel='gaussian')
```

knaaptime:
i think we need a `from_adjacency`. Currently the only way to build from a pandas adjlist is to give that directly to the init

martinfleis:
> i think we need a `from_adjacency`. Currently the only way to build from a pandas adjlist is to give that directly to the init

or using `from_arrays(df['focal'], df['neighbor'], df['weight'])` but feel free to write `from_adjacency` as well. That surely won't hurt.

knaaptime:
>or using from_arrays(df['focal'], df['neighbor'], df['weight'])

that assumes they are sorted the same way

knaaptime:
(the within-focal sorting of the neighbors may be different for each focal, so you wont get the right sparse)

martinfleis:
True. Then the constructor method may be a wise thing to do.

knaaptime:
https://gist.github.com/knaaptime/60bcf942508a47f6028ab1bfd1f3fa26

knaaptime:
update: looks like i was using old geographs code. This mostly works now (see updated gist)

https://gist.github.com/knaaptime/60bcf942508a47f6028ab1bfd1f3fa26

only thing is the user has to know how to reshape the adjlist to matrix and reindex the rows/cols

martinfleis:
I think it is a good idea to have a constructor where we do this sorting ourselves. We just need to be clear that this happens as we won't have the original df as a reference as with the `build_*` constructors.

knaaptime:
im happy to add a `_from_adjlist` constructor, cause i think we should have one anyway, and it can include the reindexing logic, which should be pretty costless.

But this still doesnt solve the problem raised above. To get a kernel weight based on precomputed network distances, a user still has to manually convert to a matrix from an adjcency. Technically you can do the signature

`Graph.build_kernel(distance_matrix, metric="precomputed", threshold=2000, kernel="gaussian")`, but its hard for us to help you get an adjlist into the right `distance_matrix` unless we allow something like Martin's transform

`Graph.from_sparse(distance_matrix).transform(transformation="kernel", threshold=2000, kernel='gaussian')`. I'm not sure i love that convention, but dont have a better idea at the moment

knaaptime:
> I think it is a good idea to have a constructor where we do this sorting ourselves. We just need to be clear that this happens as we won't have the original df as a reference as with the build_* constructors.

as of [this commit](https://github.com/pysal/libpysal/blob/6a4c004268119ab92ff7c87a866b869cd36f78ee/libpysal/graph/base.py#L1080), we're always reordering the inner index, (we take the row index as given and reorder the column index in the sparse), right? so i think that's just a stipulation of the Graph in general?

martinfleis:
That function is not used yet. I need to look into that and figure out if that's actually true in all cases. I think it is but don't want to say it for sure before testing.

knaaptime:
ok, let me know what you find. After working through this in #513, i ended up pretty confident that we always need to reorder columns to match row order

PR #551, opened 2023-08-22
BUG: fix Graph.to_W

martinfleis:
I've noticed that when ids are not ordered, they get ordered within groupby which messes up the conversion to `W`. This seems to fix it.


jGaboardi (approved):

jGaboardi:
LGTM. Let's merge after Windows passes.

jGaboardi:
(as a side note, I've never even heard of `pandas.factorize()`. You really do learn something new every day!)

Issue #534, opened 2023-07-18
Weights sprint planning


ljwolf:
I'm starting this thread to begin discussions about the upcoming weights sprint. The current experimental implementation is in the [`geographs` branch, the `weights/experimental` module](https://github.com/pysal/libpysal/tree/geographs/libpysal/weights/experimental). The `ROADMAP.md` will need updating...

martinfleis:
Things we have right now:

- Base class
- Generic constructors (`from_arrays`, `from_sparse`, `from_old_w` (for ease of testing), `from_weights_dict`, `from_dicts`), all channeled through `from_arrays`.
- Contiguity constructors (`from_contiguity`)
- A bunch of properties
- Transformation
- Handling of islands
- basics tooling for kernel constructors
- unfinished tooling for triangulation constructors
- some bit of set ops but @ljwolf will know the state of that better
- spatial lag

What needs to be done:
- hook kernel constructors to the base class
- finish triangulation and hook to the base class
- **tests**
    - tests of the functionality independent of current codebase
    - tests against the current implementations. these will be gone at some point but we need to ensure that the results from old and new match
- I/O
- plotting
- lag_categorical
- finish set operations
- fuzzy_contiguity
- bunch of there utils we have now

Where we're blocked by scipy.sparse development
- components
- higher_order

My suggestion would be to focus on testing right now. The code is becoming too large and it shall be thoroughly covered before we start moving forward.

ljwolf:
I will try to finish scipy/scipy#18544, which should unblock higher order. We will need to also identify why connected components fails in the csgraph side, but that should be doable tomorrow. 

If necessary, we can build on top of the matrix classes for now and migrate once the arrays are feature complete?

martinfleis:
csgraph does seem to generally require matrix for now.

> If necessary, we can build on top of the matrix classes for now and migrate once the arrays are feature complete?

Maybe as on-the-fly conversion when necessary?

ljwolf:
We are currently not sure if there is an "adaptive" bandwidth parameter allowed in the `kernel()` function.

ljwolf:
Not going to be able to complete scipy/scipy#18544 today, but will aim to next sprint.

knaaptime:
TLDR: 
- whats a the intended signature for instantiating a kernel-based W from a dataframe?
- whats the signature for instantiating from a precomputed distance matrix?



[...]

PR #554, opened 2023-08-28
Coindicent tests

martinfleis:
This includes proper sorting within triangulation and tests for that.

I believe that we need to sort only by focal and only in cases where the constructor has a tendency to mess up the order. The rest shall always give equal results, so no more sorting shall be needed.

@ljwolf are cliques finished? The implementation does not seem to work, the resulting adjacency is always much shorter than it should be (even shorter than the one before inducing cliques, so something is wrong).

I have also included tests for cliques but since those are failing, they're now marked with xfail. When the implementation is fixed, they should pass (maybe a minor reorder of expected tails will be needed).

@knaaptime on the sorting, I will need to be convinced that we always need to sort by the use case where not sorting everything by default causes trouble.

### DIFF
--- a/libpysal/graph/_kernel.py
+++ b/libpysal/graph/_kernel.py
@@ -230,7 +230,6 @@ def _distance_band(coordinates, threshold, ids=None):
     tree = spatial.KDTree(coordinates)
     dist = tree.sparse_distance_matrix(tree, threshold, output_type="ndarray")
     sp = sparse.csr_array((dist["v"], (dist["i"], dist["j"])))
-    sp.eliminate_zeros()
     return sp
 
     # TODO: handle co-located points
--- a/libpysal/graph/_triangulation.py
+++ b/libpysal/graph/_triangulation.py
@@ -111,8 +111,12 @@ def _validate_coincident(triangulator):
             adjtable = _induce_cliques(adjtable, coincident_lut, fill_value=fill_value)
 
         # ensure proper sorting
-        mapping = dict(zip(ids, range(len(ids))))
-        sorted_index = adjtable.focal.map(mapping).sort_values().index
+        sorted_index = (
+            adjtable[["focal", "neighbor"]]
+            .applymap(list(ids).index)
+            .sort_values(["focal", "neighbor"])
+            .index
+        )
 
         # return data for Graph.from_arrays
         return heads[sorted_index], tails[sorted_index], weights[sorted_index]
--- a/libpysal/graph/_triangulation.py
+++ b/libpysal/graph/_triangulation.py
@@ -37,7 +37,13 @@ def _validate_coincident(triangulator):
 
     @wraps(triangulator)
     def tri_with_validation(
-        coordinates, ids=None, coincident="raise", kernel=None, bandwidth=None, **kwargs
+        coordinates,
+        ids=None,
+        coincident="raise",
+        kernel=None,
+        bandwidth=None,
+        seed=None,
+        **kwargs,
     ):
         # validate geometry input
         coordinates, ids, geoms = _validate_geometry_input(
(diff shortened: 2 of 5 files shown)


ljwolf:
It should have been complete. The result wasn't incomplete on the test set I posted in the PR? I don't quite know what would be missing if it's now not working.

libpysal/graph/_kernel.py
```diff
@@ -230,7 +230,6 @@ def _distance_band(coordinates, threshold, ids=None):
     tree = spatial.KDTree(coordinates)
     dist = tree.sparse_distance_matrix(tree, threshold, output_type="ndarray")
     sp = sparse.csr_array((dist["v"], (dist["i"], dist["j"])))
-    sp.eliminate_zeros()
```
jGaboardi, line 233: Should this be completely removed or perhaps skipped by default and performed on conditional?

knaaptime:
>@knaaptime on the sorting, I will need to be convinced that we always need to sort by the use case where not sorting everything by default causes trouble.

i'm with you, but maybe we should revisit our design goals now that this has come full circle. When we originally sketched this out, we were trying to [hide issues of sorting to the user](https://github.com/pysal/libpysal/pull/513#discussion_r1204957311), since conceptually the graph is unordered. The goal was that

> not supporting _any sorting whatsoever_ will make sure that the input order is never broken, so long as we can avoid `pandas`'s predilection for it... [pysal/libpysal@geographs/libpysal/weights/experimental/base.py#L216](https://github.com/pysal/libpysal/blob/geographs/libpysal/weights/experimental/base.py?rgh-link-date=2023-05-26T18%3A59%3A25Z#L216)

but it's clear there are cases where we need to sort because the input order is sometimes broken. So we've implemented sorting in cases like inducing cliques, where we need to know the sort order to expand the neighbor set. In that case, the goal is to respect the focal index and then reorder the neighbors slightly, so we can manage the data internally. The neighbor sorting is invisible to the user, and it happens without notification, because the presumption is the user really doesnt care about the neighbor sorting (within each focal)... this part of the graph *is* unordered from the users' perspective because the only time alignment matters is when converting to a matrix representation, and they dont really need to encounter that part.

So what i'm getting at is, in some cases we're saying the user doesnt need to know about this part, and we'll handle it the best way, but *in principle* "[we] will need to be convinced that we always need to sort by the use case where not sorting everything by default causes trouble."

so, in some cases "we're handling it, because it doesnt matter to the user", but in other cases, "the user is correct and their data should not be manipulated". 

personally, i think its preferable to say "we're always handling the inner index (order of the neighbors)" because the graph is unordered, except with respect to focal. If it's unordered from the user's perspective, then internally we should just be consistent about reshaping into the best representation (in which case, i'd say we just commit to sorting the neighbors so we have an actual policy)

knaaptime:
(to me, the best way to resolve the ordered/unordered compromise is to say "the focal index is the only order that matters". Since the neighbor order is arbitrary, both levels of the graphy are always ordered along the focal index. That gives a consistent policy with the best data layout)

libpysal/graph/_kernel.py
martinfleis: Removed from here. We deal with the diagonal later. Though that logic should maybe happen here...

You would add a keyword specifying if self-weight shall be included?

jGaboardi: Do you mean something like the [`taper`](https://github.com/pysal/libpysal/blob/6a4c004268119ab92ff7c87a866b869cd36f78ee/libpysal/graph/_kernel.py#L150) keyword that is happening in `_kernel()`?

martinfleis: yes

martinfleis:
I have added sorting on `['focal', 'neighbor']` to ensure this passes and can be merged. We can potentially refactor depending on the result of discussions in #534.

knaaptime:
im gonna go ahead and merge this and explore the [sparse input stuff](https://github.com/pysal/libpysal/issues/534#issuecomment-1703056589)

PR #555, opened 2023-09-01
add from_adjacency and handle sorting

knaaptime:
for discussion

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -657,8 +657,14 @@ class Graph(_Set_Mixin):
         # set isolates to 0 - distance band should never contain self-weight
         adjacency.loc[~adjacency.index.isin(no_isolates.index), "weight"] = 0
         # re-sort using canonical order
-
-        return cls(adjacency)
+        adjacency = (
+            adjacency.reset_index()
+            .set_index(["focal", "neighbor"])
+            .reindex(ids, level=0)
+            .reindex(ids, level=1)
+            .reset_index()
+        )
+        return cls(adjacency.set_index("focal"))
 
     @classmethod
     def build_block_contiguity(cls, regimes):
@@ -781,16 +787,17 @@ class Graph(_Set_Mixin):
         scipy.sparse.COO
             sparse representation of the adjacency
         """
-        return sparse.coo_array(
-            (
-                self._adjacency.weight.values,
-                (
-                    self._adjacency.index.map(self._id2i),
-                    self._adjacency.neighbor.map(self._id2i),
-                ),
-            ),
-            shape=(self.n, self.n),
+        ids = list(self._id2i.keys())
+        # convert to multiindex series with sparse dtype
+        adj = (
+            self.adjacency.reset_index()
+            .set_index(["focal", "neighbor"])["weight"]
+            .astype("Sparse[float]")
+            .reindex(ids, level=0)
+            .reindex(ids, level=1)
         )
+        # pivot to COO sparse matrix and cast to array
+        return sparse.coo_array(adj.sparse.to_coo()[0])
 
     @cached_property
     def _id2i(self):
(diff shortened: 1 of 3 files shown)


knaaptime:
failures are expected

libpysal/graph/base.py
```diff
@@ -622,10 +669,15 @@ def build_distance_band(
...
-
-        return cls(adjacency)
+        # re-sort using canonical order
+        adjacency = (
+            adjacency.reset_index()
+            .set_index(["focal", "neighbor"])
+            .reindex(ids, level=0)
+            .reindex(ids, level=1)
+            .reset_index()
+        )
```
martinfleis, line 679: We don't need the sorting logic here if it is in the `__init__`, no?

libpysal/graph/base.py
```diff
@@ -748,16 +800,17 @@ def sparse(self):
...
+        # convert to multiindex series with sparse dtype
+        adj = (
+            self.adjacency.reset_index()
+            .set_index(["focal", "neighbor"])["weight"]
+            .astype("Sparse[float]")
+            .reindex(ids, level=0)
+            .reindex(ids, level=1)
         )
+        # pivot to COO sparse matrix and cast to array
+        return sparse.coo_array(adj.sparse.to_coo(sort_labels=True)[0])
```
martinfleis, line 813: If we use the `_id2i` here anyway, what is the benefit of reindexing, casting to pandas sparse, to scipy sparse matrix and then to a sparse array over the direct constructor we had here?

libpysal/graph/base.py
```diff
@@ -1096,11 +1149,9 @@ def _arrange_arrays(heads, tails, weights, ids=None):
     to heads *first*, then sorted within the tails.
     """
     if ids is None:
-        ids = numpy.unique(numpy.hstack((heads, tails)))
+        ids = np.unique(np.hstack((heads, tails)))
```
martinfleis, line 1152: Given the requirement for the presence of every observation in focal, at least as a self-loop, the `hstack` is not needed anymore.

libpysal/graph/base.py
knaaptime: Yeah this can go

libpysal/graph/base.py
knaaptime: i also think that whenever possible we should rely on indexing rather than mapping by hand, as indices are really the benefit that the adjlist-backed implementation provides. Frankly, the multiindexed series is what we should really use under the hood at all times

martinfleis: We originally started with a Series with a MultiIndex but there was something funny about sorting messing up sparse in some cases. But we may have prematurely dropped the idea and resorted to DataFrame. I also think that a Series would be optimal data structure. Also from performance perspective. 

Given how much the code has changed since then and that we do this sorting now, we may want to revise that and go back to Series. It should not be a complex change. We need to adapt `from_arrays` and `init` and maybe throw in a couple of `reset_index` here and there.

ljwolf: All for revisiting that. We jettisoned it because of sorting issues, but it's the correct data structure and we started out intending to keep @knaaptime's core approach

martinfleis: I think that the requirement to keep a MultiIndex lexicographically sorted to do anything performant was the main issue. pandas does not like indices that are not sorted (see this [section](https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html#sorting-a-multiindex) in their docs). In some cases, they even raise for unsorted indices, though that may not be our case. 

While we do have indices sorted, they are sorted according to the order of appearance in the original dataframe, which is not sorted in pandas terms. So while I would like to store adjacency as a mutli-indexed series, I am not sure it is possible and if it is possible, if it is actually a good idea.

martinfleis: That said, I am also happy to give it a try so we're not just guessing here.

PR #557, opened 2023-09-05
adjacency as mutliindex series

martinfleis:
This builds on top of @knaaptime's #555 and refactors `_adjacency` from a DataFrame to a Series with a MultiIndex. 

All seems to be good 🎉

I still want to look at performance but it is promising.

closes #555

### DIFF
--- a/ci/38-minimal.yaml
+++ b/ci/38-minimal.yaml
@@ -5,7 +5,7 @@ dependencies:
   - python=3.8
   - beautifulsoup4=4.10
   - jinja2=3.0
-  - pandas=1.3
+  - pandas=1.4
   - scipy=1.8
   - xarray=0.18
   # testing
--- a/libpysal/graph/_parquet.py
+++ b/libpysal/graph/_parquet.py
@@ -23,7 +23,7 @@ def _to_parquet(G, destination, **kwargs):
         import pyarrow.parquet as pq
     except (ImportError, ModuleNotFoundError):
         raise ImportError("pyarrow is required for `to_parquet`.")
-    table = pa.Table.from_pandas(G._adjacency)
+    table = pa.Table.from_pandas(G._adjacency.to_frame())
 
     meta = table.schema.metadata
     d = {"transformation": G.transformation, "version": libpysal.__version__}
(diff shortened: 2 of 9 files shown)


knaaptime:
sweet! I would be curious to see the benchmarks as well. We don't really do much indexing/slicing internally, so i'd be kinda surprised if we run into any serious performance issues with the multiindex (and copy/view shouldn't be a problem since the Graph is immutable). I havent tested thoroughly, but afaict the conversion to sparse is much faster this way. If it pans out, we can do the same thing to consume in `from_sparse` using `pd.Series.sparse.from_coo()`

martinfleis:
This comparison is not even close.

I have a large weighted distance band graph with nearly 2 million edges derived from `pts` dataframe. And I have turned cached properties to properties to get a proper measurement.

```py
# getitem
for i in pts.index:
    n = db[i]

this: 400 ms ± 5.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
geographs: 1.19 s ± 49.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```py
s = db.sparse

this: 17.3 ms ± 1.65 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
geographs: 42.8 ms ± 604 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

```py
for i in pts.index:
    n = db.get_neighbors(i)

this: 390 ms ± 927 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
geographs: 7.27 s ± 168 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

For anything else this shouldn't matter. 

I'll have a look at that sparse issue in minimal and then I'd suggest we merge this (it includes #555 so that can be closed by the merge commit). Then we can continue on the remaining work. I suppose that some set operations will need to be fixed for the Series-based adjacency but that can be done while writing tests for them.

@knaaptime thanks for pushing this!

knaaptime (approved):

martinfleis:
> I'll have a look at that sparse issue in minimal

Seems to be a bug in pandas 1.3 that is fixed in 1.4. bumping minimal requirements.

PR #560, opened 2023-09-06
set nopython=True in jit in alpha shapes

martinfleis:
A follow-up on #535 trying to eliminate warnings emitted by numba. Locally, this causes no issue, let's see what CI will say.


jGaboardi (approved):

jGaboardi:
Looks like CI is happy! Thanks!

PR #561, opened 2023-09-07
perimeter weights, fix kernel, set_ops

martinfleis:
- revise perimeter weights
- fix kernel weirdess
- start working on set ops

I started doing set ops but got stuck on the expected logic there that would not break our sorting and sparse I/O assumptions. Will need to discuss that later today. See the comment in `intersection` in the meantime.

### DIFF
--- a/libpysal/graph/_set_ops.py
+++ b/libpysal/graph/_set_ops.py
@@ -34,46 +34,42 @@ class _Set_Mixin:
         return symmetric_difference(self, other)
 
     def __iand__(self, other):
-        raise TypeError("weights are immutable")
+        raise TypeError("Graphs are immutable")
 
     def __ior__(self, other):
-        raise TypeError("weights are immutable")
+        raise TypeError("Graphs are immutable")
 
     def __len__(self):
         return self.n_edges
 
 
-# TODO: performance test the pandas implementation
 def intersects(left, right):
     """
-    A full table join is unnecessary here, but I'm not sure if
-    it would be faster? pandas joins are very fast, even for big tables,
-    but this can do early termination at the first intersection.
-
-    Maybe we use a heuristic to pick the fastest code path? It's also
-    very easy to do early termination in cython/numba using the row,col array.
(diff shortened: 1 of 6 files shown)


libpysal/graph/_set_ops.py
```diff
@@ -34,46 +34,42 @@ def __xor__(self, other):
...
     return False
-    return not left.adjacency.join(
-        right.adjacency, on=("focal", "neighbor"), how="inner"
-    ).empty
 
 
 def intersection(left, right):
     """
-    Keep only links that are in both left and right Graph objects.
+    Returns a binary weights object, that includes only those neighbor pairs that exist
```
jGaboardi, line 58: This should be `Graph`, correct?

```suggestion
    Returns a binary Graph object, that includes only those neighbor pairs that exist
```

martinfleis:
The failures are interesting as I don't see them locally. We may need to revise how the perimeter weighting works in relation to floating point precision.

libpysal/graph/_utils.py
```diff
@@ -167,105 +167,6 @@ def _validate_sparse_input(sparse, ids=None):
     ), "coordinates should represent a distance matrix if metric='precomputed'"
     return _sparse_to_arrays(sparse, ids)
 
-
-def lat2Graph(nrows=5, ncols=5, rook=True, id_type="int"):
```
jGaboardi, line 171: What's the reasoning behind removing `lat2Graph()`?

martinfleis: this one doesn't work :D

jGaboardi: Welp... that's a good reason.

martinfleis:
This is ready as is for now. I would merge this and then merge geographs to main, then follow in a sequence of smaller targeted PRs based on the checklist in #562.

ljwolf:
I can review in the next hour or so

jGaboardi (approved):

ljwolf (approved):

PR #566, opened 2023-09-08
finalise contiguity graph builders

martinfleis:
I think I already pushed this straight to main... ffs

Anyway, this finalises contiguity stuff on graph.


jGaboardi:
> I think I already pushed this straight to main... ffs
>

Happens to the best of us.

jGaboardi (approved):

knaaptime:
not sure if we need to merge or close, but go for it either way. Looks like this gets cliques working again?

martinfleis:
@knaaptime no, cliques were just formatted with black.

PR #569, opened 2023-09-15
BUG: fix graph kernel builder when kernel returns zero

martinfleis:
When the kernel function returns zeros, the conversion of sparse to arrays drops observations causing a mismatch in the shape of Graph.sparse and original input.

We simply cannot use `taper`. If there is a good use case for calling `eliminate_zeros()`, we may revise that but with a smarter implementation to avoid this issue. See the test for reproducible example.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -52,7 +52,7 @@
            ${{ env.RUN_TEST }}
 
        - name: codecov
-         uses: codecov/codecov-action@v4
+         uses: codecov/codecov-action@v3
          with:
            token: ${{ secrets.CODECOV_TOKEN }}
            file: ./coverage.xml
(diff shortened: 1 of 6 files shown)


martinfleis:
covecov messed up :) https://github.com/codecov/codecov-action/issues/1089 

including fix here to make CI run

martinfleis:
Yup, can do that.

jGaboardi (approved):

jGaboardi:
[Failures](https://github.com/pysal/libpysal/actions/runs/6198574767/job/16829270550?pr=569#step:4:2081) seem to be simple connection errors.

knaaptime:
what about setting a temporary sentinel value, elim zeros, then replace sentinel with 0?

PR #571, opened 2023-09-20
finalise kernel constructors in graph 

martinfleis:
- `taper` is back and working properly.
- kernel constructors are tested
- fixed issues with conincident points
- exposes sklearn metrics outside knn

One question:

What shall be the behaviour of kernel when `k` is set but the bandwidth s too small so weight is zero, eliminating the link with taper? now the affected focal has less neighbors than K without a warning. I'd say we may warn but that is about it. You can retrieve the rest using `taper=False`.

_edited_

### DIFF
--- a/libpysal/graph/_contiguity.py
+++ b/libpysal/graph/_contiguity.py
@@ -6,7 +6,7 @@ import pandas
 import geopandas
 from packaging.version import Version
 
-from ._utils import _neighbor_dict_to_edges, _validate_geometry_input
+from ._utils import _neighbor_dict_to_edges, _validate_geometry_input, _resolve_islands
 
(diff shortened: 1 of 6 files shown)


jGaboardi:
Any idea what's going on with the [failures](https://github.com/pysal/libpysal/actions/runs/6266706510/job/17018225522?pr=571)?

martinfleis:
Yes, will fix

jGaboardi (approved):

jGaboardi:
Whoops, approved too soon.

martinfleis:
@jGaboardi should be fixed now. I was too lazy to create an env without sklearn and it shows...

martinfleis:
Gonna merge and continue with triangulation. Feel free to still review this, we can fix stuff in follow-ups.

PR #572, opened 2023-09-21
DOC: add Graph to API docs

martinfleis:
Adding to API reference under Experimental and fixing some formatting issues.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -33,6 +33,11 @@ Levi John Wolf (levi.john.wolf@gmail.com)
 
 
 class Graph(_Set_Mixin):
+    """Graph class encoding spatial weights matrices
+
+    The :class:`Graph` is currently experimental and its API is incomplete and unstable.
+    """
+
     def __init__(self, adjacency, transformation="O"):
         """Weights base class based on adjacency list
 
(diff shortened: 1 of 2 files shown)


libpysal/graph/base.py
```diff
@@ -44,18 +49,20 @@ def __init__(self, adjacency, transformation="O"):
 
         Parameters
         ----------
-        adjacency : pandas.DataFrame
-            pandas.DataFrame with a an index ``"focal"`` and columns
-            ``["neighbor", "weight]`` encoding the adjacency. By convention,
+        adjacency : pandas.Series
+            pandas.Series with a MultiIndex  with ``"focal"`` and ``"neigbor"`` levels
```
jGaboardi, line 53: Not sure I understand the wording here. Should it be:

> A MultiIndexed ``pandas.Series`` with ``"focal"`` and ``"neigbor"`` levels, where adjacency is encoded as a weight value.

martinfleis: What about

> A MultiIndexed pandas.Series with ``"focal"`` and ``"neigbor"`` levels encoding adjacency, and values encoding weights.

jGaboardi: That is perfect.

jGaboardi (approved):

PR #575, opened 2023-09-22
implement set ops and to_networkx

martinfleis:
The refactor of set operations on Graph based on MultiIndex and inclusion of `to_networkx` to allow `isomorphic` check.

Operations resulting in a combination of edges from two graphs (union, symmetric_difference) are allowed only if their `unique_ids` match, otherwise we do not know the canonical order of adjacency.

I am also excluding isolates from `n_edges` as a self-loop is our internal representation of an isolate and not an edge of a graph.

I have also included all set operations as methods on Graph, following the pandas model.

Tests will come hopefully later today.

### DIFF
--- a/libpysal/graph/_set_ops.py
+++ b/libpysal/graph/_set_ops.py
@@ -149,7 +149,7 @@ class _Set_Mixin:
     def issubgraph(self, right):
         """
         Return True if every link in the left Graph also occurs in the right Graph.
-        This requires both Graph are label_equal. Isolates are ignored.
+        This requires both Graphs are labeled equally. Isolates are ignored.
         """
         join = (
             self._adjacency.drop(self.isolates)
--- a/libpysal/graph/_set_ops.py
+++ b/libpysal/graph/_set_ops.py
@@ -1,4 +1,7 @@
+import numpy as np
 import pandas
+
+from ._utils import _resolve_islands
 
 
 class _Set_Mixin:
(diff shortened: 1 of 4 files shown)


martinfleis:
Given the requirement of sorting, there's no need for `label_equals` and `identical` as two methods, so I am keeping only `identical` (no sorting happens in check) and renaming it as `equals`.

libpysal/graph/tests/test_base.py
```diff
@@ -230,9 +230,9 @@ def test_W_roundtrip(self):
         np.testing.assert_array_equal(W.id_order, self.letters[:10])
 
     def test_from_sparse(self):
-        row = np.array([0, 3, 1, 0])
-        col = np.array([1, 0, 1, 2])
-        data = np.array([0.1, 0.5, 1, 0.9])
+        row = np.array([0, 0, 1, 2, 3, 3])
+        col = np.array([1, 3, 3, 2, 1, 3])
+        data = np.array([0.1, 0.5, 0.9, 0, 0.3, 0.1])
```
martinfleis, line 235: I've changed this to ensure the data follow our sorting requirement and a requirement to ensure every focal id is present, at least as isolate.

libpysal/graph/_set_ops.py
```diff
@@ -1,232 +1,204 @@
...
+            )
         )
-    except AssertionError:
-        return False
-    return True
 
+    def issubgraph(self, right):
+        """
+        Return True if every link in the left Graph also occurs in the right Graph.
+        This requires both Graph are label_equal. Isolates are ignored.
```
jGaboardi, line 152:
```suggestion
        This requires both Graphs are labeled equally. Isolates are ignored.
```

jGaboardi (approved):

knaaptime (approved):

Issue #562, opened 2023-09-07
libpysal.graph roadmap to release


martinfleis:
I tried to outline what is missing before we could cut a release with the new graph stuff. Open to a discussion.

## testing

We have a decent coverage of base but especially constructors are not always tested for correctness. So we test the API but not that the adjacency captures what it should. So missing tests:

- [ ] test correctness of contiguity builders
- [ ] test correctness of triangulation builders
- [ ] test correctness of kernel builders
- [ ] test all kernel parameters
- [ ] test all query options (both sklearn and scipy)
- [ ] test precomputed distance matrix support

## implementation/checks

Some bits are still waiting for implementation (I have likely missed stuff here) or a double check that the current implementation works as intended. 

- [ ] isolates in kernel and knn
- [ ] co-located in knn
- [ ] knn haversine checks
- [ ] set ops implementation and tests
- [ ] fix cliques
- [ ] set operations
- [ ] draft API and raise NotImplementedError?

## discussion

A few things I'd like to discuss (during the dev call in 3 minutes?)

- [ ] default kernel - W has triangular, we now have gaussian following scipy model
- [ ] expose kernels in `transform`? kernels are essentially transformations. Shall we allow to use them post-creation?
- [ ] how should set operations work? See #561 

## follow-up

Stuff that is not implemented or started but that can likely wait for subsequent release, not the initial one. Below is the list of stuff I'd like to see soon but there's more to be done.

- [ ] lag_categorical (inside `lag()`?)
- [ ] classic I/O (we have custom Parquet now)
- [ ] plotting
- [ ] fuzzy contiguity (should be straightforward)
- [ ] other utils - any priorities?

Some of the follow-up could be piped via W for the time being if we want them available.

martinfleis:
Can you drop a note if you are planning to work on any of the topics mentioned above so we don't overlap?

ljwolf:
I am unlikely to have time until sept 27

knaaptime:
I think I’ll go back to ensuring the constructors can consume sparse next

martinfleis:
With #577, we're ready for the initial _experimental_ release of Graph. 

- cliques are not fixed, but they raise `NotImplementedError` at the moment.
- precomputed matrix is not tested explicitly but a few of our internal functions use it, so it is tested via those.
- API is not yet drafted, but that does not need to be for a release
- The default kernel is still Gaussian
[...]

PR #578, opened 2023-09-29
GHA: update release action

martinfleis:
Follow `spaghetti` example and generate changelog using Github.

### DIFF
--- a/.github/workflows/release_and_publish.yml
+++ b/.github/workflows/release_and_publish.yml
@@ -1,20 +1,26 @@
  # Release package on GitHub and publish to PyPI
- 
+
  # Important: In order to trigger this workflow for the organization
  # repo (organzation-name/repo-name vs. user-name/repo-name), a tagged
  # commit must be made to *organzation-name/repo-name*. If the tagged
  # commit is made to *user-name/repo-name*, a release will be published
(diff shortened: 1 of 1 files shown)


.github/workflows/release_and_publish.yml
```diff
@@ -1,65 +1,57 @@
...
        - name: Set up python
          uses: actions/setup-python@v4
          with:
            python-version: "3.x"
- 
+
        - name: Install Dependencies
          run: |
            python -m pip install --upgrade pip
-           pip install setuptools wheel twine jupyter urllib3 pandas pyyaml
```
martinfleis, line 34: My assumption is that `jupyter urllib3 pandas pyyaml` were required for `gitcount`

jGaboardi: Good point. Let's try without. Then can update other packages when we know for sure.

jGaboardi (approved):

jGaboardi:
xref https://github.com/pysal/libpysal/issues/562

.github/workflows/release_and_publish.yml
knaaptime: i think thats right

Issue #530, opened 2023-05-29
is `geomet` still a dependency?


jGaboardi:
* if so, `geomet` is now available [via `conda-forge`](https://anaconda.org/conda-forge/geomet) and it can be moved from [`requirements_plus_pip.txt`](https://github.com/pysal/libpysal/blob/main/requirements_plus_pip.txt) to [`requirements_plus_conda.txt `](https://github.com/pysal/libpysal/blob/main/requirements_plus_conda.txt)
* if not, `requirements_plus_pip.txt` can be entirely removed anyway
* testing suite passes without `geomet`

martinfleis:
I only found this https://github.com/pysal/libpysal/blob/8fa39a8c919f2629f4439254343670e691047f42/libpysal/io/iohandlers/db.py#L6-L9

which also indicates clearly how to replace it :)

jGaboardi:
Closed via https://github.com/pysal/libpysal/pull/579

QuLogic:
It's no longer in the requirements, or tested AFAICT, but the `try` is still in the file linked above. Should that be removed now, since it's untested?

martinfleis:
The whole `db.py` should be deprecated and removed.

PR #579, opened 2023-09-29
overhaul infrastructure

knaaptime:


requirements_plus_pip.txt
```diff
@@ -1 +0,0 @@
-geomet
```
jGaboardi, line 1: If `geomet` is no longer needed we can close out https://github.com/pysal/libpysal/issues/530

requirements_plus_conda.txt
```diff
@@ -1,14 +0,0 @@
...
-matplotlib>=1.5.1
-mplleaflet>=0.0.5
-numba
-numexpr
-networkx
-scikit-learn>=0.17.1
-seaborn>=0.7.0
-sqlalchemy
-statsmodels>=0.6.1
-xarray
```
jGaboardi, line 14: Looks like not all of these made it over to `pyproject.toml`.

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+[tool.setuptools_scm]
+
+[project]
+name = "libpysal"
+dynamic = ["version"]
+authors = [ # in alphabetical order
+    { name = "Martin Fleischmann", email = "martin@martinfleischmann.net" },
+    { name = "Eli Knaap", email = "ek@knaaptime.com" },
+    { name = "Serge Rey", email = "sjsrey@gmail.com" },
+    { name = "Levi Wolf", email = "levi.john.wolf@gmail.com" },
```
martinfleis, line 14: What is the selection logic here?

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+]
+maintainers = [{ name = "PySAL Developers" }]
+license = { text = "BSD 3-Clause" }
+description = "Core components of PySAL A library of spatial analysis functions"
```
martinfleis, line 18:
```suggestion
description = "Core components of PySAL - A library of spatial analysis functions"
```

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+keywords = ["spatial statistics", "spatial graphs"]
+readme = "README.md"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: OS Independent",
+    "Intended Audience :: Science/Research",
+    "Topic :: Scientific/Engineering :: GIS",
+]
+requires-python = ">=3.7"
```
jGaboardi, line 28: Minimum is now [3.8](https://github.com/pysal/libpysal/blob/357669184c35f6d8ce86d01e9412bb24b09b8353/.github/workflows/unittests.yml#L26), so we can bump this.

```suggestion
requires-python = ">=3.8"
```

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+dependencies = [
+    "platformdirs>=2.0.2",
+    "beautifulsoup4>=4.10",
```
martinfleis, line 31: I don't think we use `beautifulsoup4` anywhere here, do we?

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+[tool.setuptools.packages.find]
+include = ["libpysal", "libpysal.*"]
+
+[tool.black]
+line-length = 88
+
+[tool.ruff]
+line-length = 88
+select = ["E", "F", "W", "I", "UP", "N", "B", "A", "C4", "SIM", "ARG"]
+target-version = "py39"
```
jGaboardi, line 80: Should the target be `38` here since it's our minimum version?

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: OS Independent",
+    "Intended Audience :: Science/Research",
+    "Topic :: Scientific/Engineering :: GIS",
+]
+requires-python = ">=3.7"
+dependencies = [
+    "platformdirs>=2.0.2",
+    "beautifulsoup4>=4.10",
+    "jinja2>=3.0",
```
martinfleis, line 32: Same with jinja. I don't think it should be a dependency.

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+    "numpy>=1.20",
+    "pandas>=1.4",
+    "requests",
+    "scipy>=1.8",
+    "packaging"
+]
```
martinfleis, line 38: We may want to fix that before the release, but `graph` now requires also `geopandas` and `shapely`.

pyproject.toml
martinfleis, line 80: This should match `requires-python` on l. 27. One of them needs to change, I guess `requires-python`.

requirements_plus_conda.txt
knaaptime: i dont think we need them?

pyproject.toml
knaaptime: no idea. its who was in base. hoping reviewers have opinions

martinfleis: In geopandas, in `"authors"` is only Kelsey Jordahl who founded the project, no one else. Maybe that would be the solution here as well? So only @sjsrey? Just a suggestion.

pyproject.toml
knaaptime: geopandas doesnt come with shapely?

pyproject.toml
knaaptime: i copied all the deps over from requirements.txt without looking. I thouight we needed it for something in `examples`

pyproject.toml
martinfleis: it does but since we import it directly as `import shapely`, it should be explicitly listed. With `spherely` coming in future, it may no longer be the case (I don't think that will happen but it hypothetically can).

pyproject.toml
martinfleis: I can't find it by searching the repo.

pyproject.toml
knaaptime: most of the other packages have two, so i think its probably appropriate to have it be serge and levi. I'll change it to that unless they think otherwise

pyproject.toml
knaaptime: we need bs4, we dont need jinja. its used in examples.base and examples.remotes

pyproject.toml
martinfleis: Yeah, I don't know the history here.

requirements_plus_conda.txt
jGaboardi: These are need to all functionality, but they are for some. For example, [`xarray`](https://github.com/pysal/libpysal/blob/357669184c35f6d8ce86d01e9412bb24b09b8353/libpysal/weights/raster.py#L416). So they should be in a section for optional deps.

knaaptime: what do you want to call them? I'd rather discourage people from using pip[options] and encourage conda instead

pyproject.toml
knaaptime: well this is why im curious  about how folks feel about the "author" lines. I respect geopandas decision to honor Kelsey like that, but imo, founder and author aren't the same thing. At this point, Joris has written the lion's share of geopandas code, so to me that makes him the lead author and Kelsey the bdfl

knaaptime: we might alternatively have it be serge and luc, since lib is the most central layer of pysal and they're the founders of the metapackage?

martinfleis: The reason why we have Kelsey is that it solves the issue once and you never have to have a discussion _did person A did enough to be among authors?_.

martinfleis:
We need to add `pip install .` to the GHA to ensure we have access to `__version__`. That will fix the CI failure.

pyproject.toml
jGaboardi: > The reason why we have Kelsey is that it solves the issue once and you never have to have a discussion _did person A did enough to be among authors?_.

This is probably the most elegant solution for, with Serge and Lucy up in there (as you say, Eli).

requirements_plus_conda.txt
jGaboardi: I suppose it can be named anything we want. Even simply `plus`.

pyproject.toml
knaaptime: well i get the logic, but maybe its a question that should be revisited occasionally instead of solved once. Serge, Ran and I were once the authors of spopt, but now it makes more sense for the authors to be James and Xin. ...but also a lot of the code was ported over from region and originally written by juanca, serge, and aleks. So its tricky

requirements_plus_conda.txt
knaaptime: xarray is actually the only one used (sqlalchemy is technically used in an io module but we're gonna deprecate that anyway) so i dont think its worthy of its own group

jGaboardi: Fair enough. Let's move to another?

knaaptime: well actually makes me wonder why its not already in tests. Guess we're not testing the raster stuff?

jGaboardi: well that's no good...

.github/workflows/unittests.yml
```diff
@@ -1,60 +1,62 @@
...
+        os: [ubuntu-latest]
+        environment-file:
+          - ci/38-minimal.yaml
+          - ci/39.yaml
+          - ci/310.yaml
+          - ci/311.yaml
+          - ci/311-dev.yaml
+          - ci/311-no-optional.yaml
+        include:
+          - environment-file: ci/310.yaml
```
jGaboardi, line 28: Is there a s reason we aren't testing against `311` here?

.github/workflows/unittests.yml
```diff
@@ -1,60 +1,62 @@
...
-         with:
-           token: ${{ secrets.CODECOV_TOKEN }}
-           file: ./coverage.xml
-           name: libpysal-codecov
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: setup micromamba
+        uses: mamba-org/provision-with-micromamba@main
```
jGaboardi, line 40:
```suggestion
        uses: mamba-org/setup-micromamba@v1
```

pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+
+[project.urls]
+Home = "https://github.com/pysal/libpysal/"
+Repository = "https://github.com/pysal/libpysal"
+
+[project.optional-dependencies]
+dev = ["pre-commit"]
```
jGaboardi, line 44: Should we maybe add `xarray` in here?

.github/workflows/unittests.yml
knaaptime: not a strong one, but i figured its best to test the other platforms against the middle python instead of the bleeding edge python

pyproject.toml
knaaptime: if you really want :P

martinfleis (approved):
One minor nit. Otherwise looks good!

pyproject.toml
knaaptime: i'd probably characterize it as brute force rather than elegance, but sometimes any solution is preferable to nothing :P. By using founders as a replacement for authors, we're just saying our answer is to not address the question. That's fine, but it's a particular position (that assigning authorship is too hard to handle).

to throw out another few analogies, Guido is the author of Python, but at this point im sure he's the author of almost nothing in stdlib. Serge and Luc will forever be the authors of PySAL, but i think of authorship among the subpackages as additive, so the 'authors' are the respective contribution trees

i also think of pysal a bit like an edited handbook on spatial analytical methods. Serge and Luc are the editors of the handbook, and they wrote several chapters of the first edition. Many editions are pretty similar, where the same authors update their chapters in light of new evidence and techniques. But the content of the collection changes a bit over time as the discipline evolves and some of the authors rotate out, so chapters are eventually written by others. The subpackages in pysal are like the chapters, and lib is maybe the introduction to the series. 

i think theres also an argument that the four of us sat down, scoped out, wrote, and revised *this draft* of the package (the graph-based version anyway), which makes us the authors of *this version of the chapter*. Eventually somebody else will probably write a new draft, at which point they assume 'authorship' (but we will always be in the contributor tree (and picked up by the zotero citation, for whatever thats worth)

.github/workflows/unittests.yml
jGaboardi: Fair enough.

jGaboardi (approved):

pyproject.toml
jGaboardi: I think that sounds reasonable. I think it's your call at this point @knaaptime.

knaaptime: i'm the resident dev with a community planning degree, so its only my job to raise questions. Answers are for the community :P. But jokes aside i'd probably just leave it at serge and levi

martinfleis: I'm fine with any solution. Just didn't want to exclude someone if we decide to name authors based on contributions. FYI, pandas has just Pandas development team as authors.

jGaboardi:
Are we good to merge for now then, or are we waiting for something else?

martinfleis:
Two approvals are in and authorship can be discussed in the dev call and changed later if needed. Lgtm

PR #583, opened 2023-10-19
Delay example data dir creation, add fallback for unwriteable $HOME

fbdtemme:
Libpysal will try to create the data directory for storing example data at import time, even if no example objects are used.
We are running a tool in a singularity containers and have libpysal as a subdependency. We do not want to mount the user home directory in the container since this can damage reproducibility (eg python module picked up under `~/.local` etc).

libpysal causes a failure due to "read-only" OSError exceptions when creating the pysal data directory during the import when running in singularity.

This PR simply moves creation of the data dir to the moment when it is actually needed because an example dataset is used.
This avoids the issue we are having mentioned above. Additionaly this also adds a fallback in `get_data_home` to use a temporary directory instead of the users home dir if that directory cannot be created. This is the same behavior as eg. the matplotlib [MPLCONFIGDIR](https://matplotlib.org/stable/users/installing/environment_variables_faq.html#envvar-MPLCONFIGDIR) default.

### DIFF
--- a/libpysal/examples/base.py
+++ b/libpysal/examples/base.py
@@ -7,6 +7,7 @@ Base class for managing example datasets.
 
 import io
 import os
+import tempfile
 import webbrowser
 from platformdirs import user_data_dir
 import zipfile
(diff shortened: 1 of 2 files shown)


sjsrey (approved):
Looks good. Thanks!

PR #587, opened 2023-10-21
TST: fix alpha_shape failures on dev

martinfleis:


martinfleis:
It is an upstream bug https://github.com/geopandas/geopandas/pull/3047

PR #591, opened 2023-10-21
MAINT: avoid most of the warnings coming from dev CI

### LINKED ISSUE #582: GeoPandas FutureWarning in fuzzy_contiguity
Hello,

Just a heads up that `fuzzy_contiguity` produces a `FutureWarning`

Example:
```python
from libpysal.weights import fuzzy_contiguity
import geopandas as gpd

gdf = gpd.read_file("test.geojson")
gdf = gdf.set_crs(epsg=4328, allow_override=True)

fuzzy_contiguity(gdf, buffering=True, buffer=3, predicate="intersects", silence_warnings=True)
```
FutureWarning: The `query_bulk()` method is deprecated and will be removed in GeoPandas 1.0. You can use the `query()` method instead. inp, res = gdf.sindex.query_bulk(gdf.geometry, predicate=predicate)

```
print(gpd.__version__)
print(libpysal.__version__)
```
0.13.2
4.8.0

martinfleis:
Some future-proofing, a lot of silencing. Got it down from 1200+ to 5, out of which 4 will be solved in upstream (https://github.com/apache/arrow/pull/38374) and one seems to be coming from `scipy.sparse` internals, so there's not a lot we can do apart from silencing but I'd give it some time.

Closes #582


libpysal/weights/util.py
```diff
@@ -18,7 +18,7 @@
 try:
     import geopandas as gpd
 
-    GPD_08 = Version(gpd.__version__) >= Version("0.8.0")
+    GPD_013= Version(gpd.__version__) >= Version("0.13.0")
```
jGaboardi, line 21:
```suggestion
    GPD_013 = Version(gpd.__version__) >= Version("0.13.0")
```

jGaboardi (approved):
Awesome. Great work.

jGaboardi:
Approved too soon... Seeing [failures](https://github.com/pysal/libpysal/actions/runs/6599411432/job/17928404857?pr=591#step:5:6039) – `pandas._config.config.OptionError: No such keys(s): 'future.no_silent_downcasting'`

martinfleis:
Because the way pandas is deprecating the `downcast` keyword is not very friendly. You can either do that option, but that is available only on main or filter warnings. Given the behaviour in our case does not change, I will just filter them so they don't bubble up to a user.

libpysal/io/iohandlers/pyDbfIO.py
```diff
@@ -364,8 +364,8 @@ def _writeHeader(self):
         POS = self.f.tell()
         self.f.seek(0)
         ver = 3
-        now = datetime.datetime.utcfromtimestamp(
-            int(os.environ.get("SOURCE_DATE_EPOCH", time.time())),
+        now = datetime.datetime.fromtimestamp(
```
jGaboardi, line 367: 3.10 is not liking this --> https://github.com/pysal/libpysal/actions/runs/6599564231/job/17928744012?pr=591#step:5:1200

martinfleis: Let's live with the warning then :D

jGaboardi: 100% agree.

jGaboardi: 3.10 passing again. ✅

jGaboardi (approved):

PR #592, opened 2023-10-21
MAINT: dealing with `sqlalchemy` & `geomet`

### LINKED ISSUE #588: `sqlalchemy` and `geomet`
Both `sqlalchemy` and `geomet` were removed from requirements in #579, `sqlalchemy` by accident and `geomet` on purpose.

* For `sqlalchemy` we simply need to add that back into `pyproject.toml` and in the appropriate `ci/*` environment.
* For `geomet` it's a bit more complicated since we need to address some functionality and testing:
  * [libpysal/io/iohandlers/db.py](https://github.com/pysal/libpysal/blob/0ea928aaecd95343a32900acfc23dfb8a49d5fd6/libpysal/io/iohandlers/db.py#L7)
  * [libpysal/io/iohandlers/\_\_init\_\_.py](https://github.com/pysal/libpysal/blob/0ea928aaecd95343a32900acfc23dfb8a49d5fd6/libpysal/io/iohandlers/__init__.py#L22)
  * [libpysal/io/iohandlers/tests/test_db.py](https://github.com/pysal/libpysal/blob/0ea928aaecd95343a32900acfc23dfb8a49d5fd6/libpysal/io/iohandlers/tests/test_db.py#L17)

jGaboardi:
This PR resolves #588


jGaboardi:
Seems to be an [OS-specific failure](https://github.com/pysal/libpysal/actions/runs/6599595357/job/17928808831?pr=592#step:5:1590):
```
FAILED libpysal/io/iohandlers/tests/test_db.py::Test_sqlite_reader::test_deserialize - 
    PermissionError: [WinError 32] The process cannot access the file because it is 
    being used by another process: 'test.db'
```

jGaboardi:
This is actually happening in the `tearDown` portion. 
```
def tearDown(self):
>       os.remove("test.db")
E       PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'test.db'
```

Probably can simply move `os.remove("test.db")` into the actual test.

libpysal/io/iohandlers/tests/test_db.py
```diff
@@ -4,43 +4,26 @@
 import unittest as ut
 from .... import examples as pysal_examples
 
+from shapely.geometry import Point
+from shapely import wkb
```
martinfleis, line 8:
```suggestion
import shapely
```
see below

libpysal/io/iohandlers/tests/test_db.py
```diff
@@ -4,43 +4,26 @@
...
+@ut.skipIf(missing_sql, f"Missing dependency: SQLAlchemy ({missing_sql}).")
 class Test_sqlite_reader(ut.TestCase):
     def setUp(self):
         path = pysal_examples.get_path("new_haven_merged.dbf")
         if path is None:
             pysal_examples.load_example("newHaven")
             path = pysal_examples.get_path("new_haven_merged.dbf")
         df = pdio.read_files(path)
-        df["GEOMETRY"] = df["geometry"].apply(to_wkb_point)
+        df["GEOMETRY"] = df["geometry"].apply(lambda p: wkb.dumps(Point(p)))
```
martinfleis, line 26:
```suggestion
        df["GEOMETRY"] = shapely.to_wkb(shapely.points(df["geometry"]))
```
or something like that. you can completely avoid apply loop

jGaboardi: So thing here is that `df["geometry"]` is not `Point` objects yet. Let me see about a workaround.

martinfleis: I'm aware but I wasn't sure which object it is, that is why I said "something like that" :D.

jGaboardi: OK, needed some massaging. The following works:
```python
shapely.to_wkb(shapely.points(df["geometry"].values.tolist()))
```

jGaboardi:
Welp, not really sure why we are still seeing the [`PermissionError` on Windows](https://github.com/pysal/libpysal/actions/runs/6600365924/job/17930277918?pr=592#step:5:1070). Gave it go trying to resolve correctly. Unless there are other ideas, should we either skip of xfail for Windows?

martinfleis:
Skip it on win. I suppose that this code will eventually go anyway. You can probably replace it with geopandas read_sql or read SQL with GDAL via pyogrio or fiona, all of which are way more maintained than our `io` module.

jGaboardi:
> Skip it on win. I suppose that this code will eventually go anyway. You can probably replace it with geopandas read_sql or read SQL with GDAL via pyogrio or fiona, all of which are way more maintained than our `io` module.

I'll just implement the skip for now. Better to do baby steps.

jGaboardi:
We've got 100% green CI!

martinfleis (approved):
Thanks!

PR #593, opened 2023-10-21
ENH: add vectorized plotting to Graph

martinfleis:
New implementation of plotting. A fast one :).

Using `geodatasets.get_path("geoda south")`, the old `W.plot()` takes `1.76 s ± 86 ms per loop` to plot. The new `Graph.plot()` takes `15 ms ± 146 µs per loop`, both resulting the equal plot.

The API is nearly identical to `W.plot()` with some minor tweaks like the ability to control whether the nodes shall be plotted or figsize keyword when creating a new ax.

Tests will come.


martinfleis:
We have also discussed today to allow partial plots, like for a single focal or a subset of those. @sjsrey how would you like the API around that be shaped? Additional keyword `focal` used as a filter? It will be dead easy to implement.

libpysal/graph/_plotting.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Creates a ``maptlotlib`` plot based on the topology stored in the
+    Graph and spatial location defined in ``gdf``.
+
+    Parameters
+    ----------
+    G : Graph
+        Graph to be plotted
+    gdf : geopandas.GeoDataFrame
+        Geometries indexed using the same index as Graph. Geomtry types other than
```
jGaboardi, line 18:
```suggestion
        Geometries indexed using the same index as Graph. Geometry types other than
```

libpysal/graph/base.py
```diff
@@ -1197,6 +1198,71 @@ def to_networkx(self):
...
+    ):
+        """Plot edges and nodes of the Graph
+
+        Creates a ``maptlotlib`` plot based on the topology stored in the
+        Graph and spatial location defined in ``gdf``.
+
+        Parameters
+        ----------
+        gdf : geopandas.GeoDataFrame
+            Geometries indexed using the same index as Graph. Geomtry types other than
```
jGaboardi, line 1219:
```suggestion
            Geometries indexed using the same index as Graph. Geometry types other than
```

martinfleis:
This should be ready now. Allowing for filtering by focal or an array of them.

martinfleis:
A further note - this depends on index of the gdf to match graph nodes to geometries. I suppose that once we'll have a proper look ad the ID handling in the Graph, this may change. But so far it is consistent with everything else.

jGaboardi (approved):

sjsrey:
> We have also discussed today to allow partial plots, like for a single focal or a subset of those. @sjsrey how would you like the API around that be shaped? Additional keyword `focal` used as a filter? It will be dead easy to implement.

Two use cases come to mind. Let's say $n=100$ with the ids going from 0 to n-1. First use case is the user wants to see the subgraph induced by focal=[7, 94] but with the original graph shown as context. Here the extent of the window would be the extent of the original graph.

The second use case would be to "zoom in" on the induced subgraph, so the extent may/will change depending on the focal units specified.


For each of these cases, there is the issue of distinguishing the edges incident with each of the focal units. In cases where the members of focal are pairwise disjoint, a different color for the edges incident with each focal node would do the job. When a pair of focal nodes are neighbors, I'm not sure what the best way to handle that would be?

This is all assuming a static plot, in which everything has to be shown at once. With interactivity, looping over the focal units would allow the subgraph incident with each focal unit to be uniquely portrayed. But that is down the road.

Just some thoughts here. The PR looks great to me in terms of implementation and test coverage!

sjsrey (commented):
https://github.com/pysal/libpysal/pull/593#issuecomment-1776524493

martinfleis:
> Two use cases come to mind. Let's say  with the ids going from 0 to n-1. First use case is the user wants to see the subgraph induced by focal=[7, 94] but with the original graph shown as context. Here the extent of the window would be the extent of the original graph.

This can be done without distinguish the focal pairwise edges as

```py
import geopandas as gpd
import geodatasets
from libpysal import graph

df = gpd.read_file(geodatasets.get_path("geoda.nyc_neighborhoods"))
contiguity = graph.Graph.build_contiguity(df)

ax = contiguity.plot(df)
contiguity.plot(df, focal=[7, 94], color="red", ax=ax)
```
Screenshot 2023-10-24 at 10 27 16

> The second use case would be to "zoom in" on the induced subgraph, so the extent may/will change depending on the focal units specified.

For this we would need to set axis limits based on the extent of subgraph. With the upcoming commit:

```py
ax = G.plot(df)
G.plot(df, focal=[7, 94], color="red", ax=ax, limit_extent=True)
```
Screenshot 2023-10-24 at 10 41 55

> For each of these cases, there is the issue of distinguishing the edges incident with each of the focal units. In cases where the members of focal are pairwise disjoint, a different color for the edges incident with each focal node would do the job. When a pair of focal nodes are neighbors, I'm not sure what the best way to handle that would be?

This sounds like a lot of complication of the code and the API. Can we leave it for a potential follow-up if there will be an appetite to look into that?

knaaptime:
this looks sweet. One quick thing looking at the plots since  i wasnt around for the sprint is, it would be nice if you could distinguish the focal(s) from the neighbors in the subgraph(s). Different color or something

martinfleis:
I'd like to avoid to have away too many `kws` attributes to control every little aspect of the plot but distinguishing focal from neighbour is reasonable.

> i wasnt around for the sprint

neither was this topic :D

libpysal/graph/_plotting.py
```diff
@@ -0,0 +1,121 @@
...
+    Parameters
+    ----------
+    G : Graph
+        Graph to be plotted
+    gdf : geopandas.GeoDataFrame
+        Geometries indexed using the same index as Graph. Geometry types other than
+        points are converted to centroids encoding start and end point of Graph
+        edges.
+    focal : hashable | array-like[hashable] | None, optional
+        ID or an array-like of IDs of focal geometries whose weights shall be
```
knaaptime, line 32: if i'm following the plots, then focal controls whether subgraphs are highlighted, not plotted, right? the whole graph is always plotted, but if you pass a set of `focal`s, then those get highlighted as subgraphs on top? this makes it sound like you only plot the subgraphs if focal is passed

knaaptime:
> but distinguishing focal from neighbour is reasonable.

indeed. in these contiguity examples its pretty easy to intuit, but in more complicated graphs it's useful to have a way to distinguish

libpysal/graph/_plotting.py
martinfleis: No, only the the subset is plotted. Those figures I showed use two layers. Check the code. This docstring is correct. I don't want to just highlight because there are good use cases of plotting just the subgraph.

knaaptime: ah, gotcha.

martinfleis:
You can now pass optional `focal_kws` to control the aesthetics of focal nodes. It is empty by default and inherits whatever you specify in `node_kws`.

```py
ax = contiguity.plot(df)
contiguity.plot(df, focal=[7, 94], color="red", ax=ax, focal_kws=dict(color="blue", s=200), limit_extent=True)
```
8adc7bef-5a02-4ce8-b833-6a57b449f3f7

knaaptime (approved):

Issue #589, opened 2023-10-21
lint repo with `ruff`, update `pyproject.toml` and pre-commit for linting


jGaboardi:
lint repo with `ruff`, update `pyproject.toml` and pre-commit for linting.

* [x] `__init__` in all levels – #597 
* [x] top level `.py` files – #597 
* [x] `cg`
  * [x] `./cg/*.py` – #608 
  * [x] `./cg/ops/*.py` – #610
  * [x] `./cg/ops/tests/*.py` – #611 
  * [x] `./cg/tests/*.py` – #614
* [x] `examples` – #607 
* [x] `graph`
  * [x] `./graph/base.py` – #615
  * [x] `./graph/_*.py` – #618
  * [x] `./graph/tests/*.py` – #619
* [x] `io`
  * [x] `./io/*.py` – #620
  * [x] `./io/geotable/*` – #622 
  * [x] `./io/iohandlers/*` – (#630)
      * #637 
      * #638
      * #641 
      * #642 
      * #643 
  * [x] `./io/{tests,util}/*`
      * #644 
      * #645 
      * #646 
* [x] `weights`
  * https://github.com/pysal/libpysal/pull/648
  * #650 
  * #651 
  * #652 
  * #653 
  * #654 
  * #655 
  * #658 
* [x] review and final pass
  * [x] convert `# noqa <CODE>` -->   `# noqa: <CODE>` – #660 
     * Without that colon, any violation is skipped, not only the one from `<CODE>` 
  * [x] other rough edges? – #661 
* [x] update infra
  * [x] `pyproject.toml`– #661 
  * [x] `.pre-commit-config.yaml`– #661 
  * [x] enable `pre-commit` on triggered GHA – https://results.pre-commit.ci

jGaboardi:
xref #594

jGaboardi:
All files formatted and listed; pre-commit hook for CI/PRs is enabled.

PR #597, opened 2023-10-22
begin the repo linting 

jGaboardi:
This PR kicks off #589. Should we do this:
1. piece-meal one small PR at a time; or
2. do one PR, but go incrementally with small chunks + reviews

(This process is going to be a slog... between 2k-3k fixes on first inspection)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,14 +93,18 @@ ignore = [
     "B010",
     "C408",
     "E731",
-    "F401",
-    "F403",
     "N803",
     "N806",
     "N999",
     "UP007"
 ]
 exclude = ["libpysal/tests/*", "docs/*"]
+[tool.ruff.per-file-ignores]
+"*__init__.py" = [
+    "F401",  # imported but unused
+    "F403",  # star import; unable to detect undefined names
+]
+
 
 [tool.coverage.run]
 source = ["./libpysal"]
--- a/libpysal/cg/__init__.py
+++ b/libpysal/cg/__init__.py
@@ -1,11 +1,11 @@
 """
 A module for computational geometry.
 """
-from .shapes import *
-from .standalone import *
-from .locators import *
-from .kdtree import *
-from .rtree import *
-from .sphere import *
-from .voronoi import *
 from .alpha_shapes import *
+from .kdtree import *
+from .locators import *
+from .rtree import *
+from .shapes import *
+from .sphere import *
+from .standalone import *
+from .voronoi import *
--- a/libpysal/cg/ops/__init__.py
+++ b/libpysal/cg/ops/__init__.py
@@ -1,2 +1,1 @@
-from . import atomic
-from . import tabular
+from . import atomic, tabular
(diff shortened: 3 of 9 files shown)


martinfleis:
I would do one PR that contains only the automatic fixes by ruff but across the whole code base and than a series of smaller PRs with the rest. The former does not need a ton of review and may resolve quite a lot already.

pyproject.toml
```diff
@@ -93,14 +93,18 @@ ignore = [
...
     "N803",
     "N806",
     "N999",
     "UP007"
 ]
 exclude = ["libpysal/tests/*", "docs/*"]
+[tool.ruff.per-file-ignores]
+"*__init__.py" = [
+    "F401",  # imported but unused
+    "F403",  # star import; unable to detect undefined names
```
martinfleis, line 105: I'd like to eliminate star imports if possible but that can be a follow-up at some point in the future. It just may be worth opening an issue about it?

martinfleis:
That said, I am happy to merge this as is.

jGaboardi:
> I would do one PR that contains only the automatic fixes by ruff but across the whole code base and than a series of smaller PRs with the rest. The former does not need a ton of review and may resolve quite a lot already.

The problem with autofixing everything is that is sometimes breaks things. For example, in [`libpysal/weights/__init__.py`](https://github.com/pysal/libpysal/pull/597/files#diff-1639c0182a1375d17f3d4676210729c0d8e2fd15c013b355fc2e5e7f83902efdL1) the `from .weights import * # noqa I001` import location can't change or we see the following:

```
ImportError while importing test module '/Users/userlibpysal/libpysal/weights/tests/test_weights_IO.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../mambaforge/envs/py311_libpysal/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
libpysal/__init__.py:27: in <module>
    from . import cg, examples, io, weights
libpysal/examples/__init__.py:14: in <module>
    from .base import example_manager
libpysal/examples/base.py:17: in <module>
    from ..io import open as ps_open
libpysal/io/__init__.py:2: in <module>
    from .iohandlers import *
libpysal/io/iohandlers/__init__.py:6: in <module>
    from . import dat
libpysal/io/iohandlers/dat.py:2: in <module>
    from ...weights import W
libpysal/weights/__init__.py:3: in <module>
    from .gabriel import Delaunay, Gabriel, Relative_Neighborhood
libpysal/weights/gabriel.py:3: in <module>
    from libpysal.weights import W, WSP
E   ImportError: cannot import name 'W' from partially initialized module 'libpysal.weights' (most likely due to a circular import) (/Users/user/libpysal/libpysal/weights/__init__.py)
```

pyproject.toml
jGaboardi: I completely agree about star imports; and making a separate issue.

martinfleis:
> The problem with autofixing everything is that is sometimes breaks things

fun ... okay, let's go piece by piece then

martinfleis (approved):

PR #601, opened 2023-10-23
TST: convert unittests to pytest

### LINKED ISSUE #595: Convert all testing from `unittest` to `pytest`.
Convert all testing from `unittest` to `pytest`.

xref #594

martinfleis:
Closes #595

Mostly done automagically using `pytestify` with some of manual fixes. I also occasionally sorted imports with ruff and changed `Test_Rook` to `TestRook` to make ruff happy but not consistently.

I have removed all optional skips based on missing pandas, geopandas or shapely since all are hard dependencies now.

### DIFF
--- a/libpysal/cg/tests/test_ashapes.py
+++ b/libpysal/cg/tests/test_ashapes.py
@@ -18,8 +18,8 @@ this_directory = os.path.dirname(__file__)
 
 
 @skipIf(GEOPANDAS_EXTINCT, "Geopandas is missing, so test will not run.")
-class Test_Alpha_Shapes(TestCase):
-    def setUp(self):
+class Test_Alpha_Shapes:
+    def setup_method(self):
         eberly = geopandas.read_file(get_path("eberly_net.shp"))
         eberly_vertices = eberly.geometry.apply(
             lambda x: np.hstack(x.xy).reshape(2, 2).T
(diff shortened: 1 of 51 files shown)


libpysal/cg/tests/test_ashapes.py
```diff
@@ -18,8 +18,8 @@
 
 
 @skipIf(GEOPANDAS_EXTINCT, "Geopandas is missing, so test will not run.")
-class Test_Alpha_Shapes(TestCase):
-    def setUp(self):
+class Test_Alpha_Shapes:
```
jGaboardi, line 21: >  I also occasionally sorted imports with ruff and changed `Test_Rook` to `TestRook` to make ruff happy but not consistently.

So are we OK with underscored class names *for now* or *in general*? I have no problem with "for now," especially not in this PR, but probably we'll want to take care of that soon. And I can loop that in with #589.

jGaboardi (approved):
@martinfleis Absolutely amazing. Never heard of `pytestify`.

libpysal/cg/tests/test_ashapes.py
martinfleis: If change private bits like tests. We can have a discussion about public but I would probably keep those intact to avoid unnecessary deprecations.

martinfleis:
> Never heard of pytestify.

Neither did I until today 😄. It is not perfect but goes a long way.

libpysal/cg/tests/test_ashapes.py
jGaboardi: Agreed.

PR #610, opened 2023-10-26
format, lint, numpydoc – `cg/ops/*.py`

jGaboardi:
xref https://github.com/pysal/libpysal/issues/589

This PR formats, lints, & https://github.com/pysal/libpysal/pull/607#discussion_r1371335093 from `./cg/ops/*.py`.

PR #611, opened 2023-10-28
format & lint – `cg/ops/tests*.py`

jGaboardi:

This PR builds on top of #610 (should review and merge that one first) and formats & lints `cg/ops/tests*.py`

### DIFF
--- a/libpysal/cg/ops/tests/test_accessors.py
+++ b/libpysal/cg/ops/tests/test_accessors.py
@@ -211,9 +211,45 @@ class TestAccessors:
         with pytest.raises(AttributeError):
             to_test.parts(self.points)
 
-        to_test.parts(self.lines)
-        pgon_parts = to_test.parts(self.polygons)
+        line_parts = to_test.parts(self.lines)
+        line_answers = [
+            [
+                [
+                    (-0.009053924887015952, -0.25832280562918325),
+                    (0.007481157395930582, -0.2589587703323735),
+                    (0.007481157395930582, -0.2589587703323735),
+                ]
+            ],
+            [
+                [
+                    (0.10923550990637088, -0.2564149115196125),
+                    (0.12895041570526866, -0.2564149115196125),
+                ]
+            ],
+            [
+                [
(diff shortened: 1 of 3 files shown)


libpysal/cg/ops/tests/test_accessors.py
```diff
@@ -211,7 +211,7 @@ def test_parts(self):
         with pytest.raises(AttributeError):
             to_test.parts(self.points)
 
-        line_parts = to_test.parts(self.lines)
+        to_test.parts(self.lines)
```
martinfleis, line 214: I am not super excited about changes like this. `parts` return something. We then don't use it but wouldn't a better solution be `_ = to_test.parts(self.lines)`? To indicate that this indeed returns something but we intentionally ignore it?

libpysal/cg/ops/tests/test_accessors.py
```diff
@@ -275,42 +275,9 @@ def test_parts(self):
                 ]
             ],
         ]
-        line_answers = [
```
martinfleis, line 278: huh? Maybe the `line_parts` above should have been tested against this.

libpysal/cg/ops/tests/test_accessors.py
jGaboardi: Seems to be another auto fix snafu that went under the radar (including your comment below). I'll look into it.

jGaboardi: @martinfleis  So, what it seems like is the `line_parts` vs. `line_answers` was never actually being tested, which led to `ruff` determining it was unused code marked for removal. I popped in a testing chunk for `line_parts` vs. `line_answers` and all seems appropriate now.

libpysal/cg/ops/tests/test_accessors.py
jGaboardi: See [above](https://github.com/pysal/libpysal/pull/611#discussion_r1376189795).

martinfleis (approved):

PR #615, opened 2023-10-31
lint `graph/base.py`

jGaboardi:

This PR kicks off the linting of `libpysal/graph` with `base.py`.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1113,7 +1113,7 @@ class Graph(_Set_Mixin):
             for j in range(1, k):
                 wj = sp**j
                 rj, cj = wj.nonzero()
-                sj = set(zip(rj, cj), strict=True)
+                sj = set(zip(rj, cj, strict=True))
                 sk.difference_update(sj)
         if not diagonal:
             sk = {(i, j) for i, j in sk if i != j}
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -223,7 +223,7 @@ class Graph(_Set_Mixin):
             libpysal.graph.Graph
         """
         cols = dict(
-            zip(  # noqa B905
+            zip(
                 [focal_col, neighbor_col, weight_col],
                 ["focal_col", "neighbor_col", "weight_col"],
                 strict=True,
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -36,7 +36,8 @@ Levi John Wolf (levi.john.wolf@gmail.com)
 class Graph(_Set_Mixin):
     """Graph class encoding spatial weights matrices
 
-    The :class:`Graph` is currently experimental and its API is incomplete and unstable.
+    The :class:`Graph` is currently experimental
+    and its API is incomplete and unstable.
     """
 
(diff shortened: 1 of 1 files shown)


libpysal/graph/base.py
```diff
@@ -223,12 +223,12 @@ def from_adjacency(
             libpysal.graph.Graph
         """
         cols = dict(
-            zip(
+            zip(  # noqa B905
```
jGaboardi, line 226: @martinfleis I'm curious what you think about keeping vs. ignoring the `strict=True` parameter in situations like. Since we explicitly know there are going to be 3 elements in iterable there is really no need to have a `strict=True`.

martinfleis: I guess it is fine to keep it there. Better than `noqa`. We don't really have a reason to exclude them. I am not sure if there's some performance hit but I don't expect it to be noticeable.

libpysal/graph/base.py
```diff
@@ -904,7 +902,7 @@ def transform(self, transformation):
             s = self._adjacency.groupby(level=0).transform(
                 lambda group: group / math.sqrt((group**2).sum())
             )
-            nQ = self.n / s.sum()
+            nQ = self.n / s.sum()  # noqa N806
             standardized = (s * nQ).fillna(0).values  # isolate comes as NaN -> 0
```
martinfleis, line 906:
```suggestion
            n_q = self.n / s.sum()
            standardized = (s * n_q).fillna(0).values  # isolate comes as NaN -> 0
```

There is no reason not to follow Ruff's recommendation here.

libpysal/graph/base.py
```diff
@@ -1102,22 +1100,22 @@ def higher_order(self, k=2, shortest_path=True, diagonal=False, lower_order=Fals
...
         if lower_order:
-            wk = sum(map(lambda x: sp**x, range(2, k + 1)))
+            wk = sum(sp**x for x in range(2, k + 1))
             shortest_path = False
         else:
             wk = sp**k
 
         rk, ck = wk.nonzero()
-        sk = set(zip(rk, ck))
+        sk = set(zip(rk, ck))  # noqa B905
```
martinfleis, line 1109:
```suggestion
        sk = set(zip(rk, ck, strict=True))
```

libpysal/graph/base.py
```diff
@@ -1102,22 +1100,22 @@ def higher_order(self, k=2, shortest_path=True, diagonal=False, lower_order=Fals
...
 
         if shortest_path:
             for j in range(1, k):
                 wj = sp**j
                 rj, cj = wj.nonzero()
-                sj = set(zip(rj, cj))
+                sj = set(zip(rj, cj))  # noqa B905
```
martinfleis, line 1115:
```suggestion
                sj = set(zip(rj, cj), strict=True)
```

libpysal/graph/base.py
```diff
@@ -223,12 +223,13 @@ def from_adjacency(
             libpysal.graph.Graph
         """
         cols = dict(
-            zip(
+            zip(  # noqa B905
```
jGaboardi, line 226:
```suggestion
            zip(
```

martinfleis (approved):

PR #616, opened 2023-10-31
fix index check for precomputed graph distances

knaaptime:
small tweak, but the existing code will fail when trying to instantiate a kernel graph from a precomputed (sparse) distance matrix (because it has no len)

when passing a precomputed distance matrix its also handy to be able to pass ids as well. I think that would be a small change in this function too, but not sure if we disallowed that for some reason

### DIFF
--- a/libpysal/graph/tests/test_builders.py
+++ b/libpysal/graph/tests/test_builders.py
@@ -4,7 +4,7 @@ import numpy as np
 import pandas as pd
 import pytest
 from numpy.testing import assert_array_almost_equal
-from scipy.sparse.csr import csr_matrix
+from scipy.sparse import csr_matrix
 from shapely import get_coordinates
 
 from libpysal import graph
--- a/libpysal/graph/tests/test_builders.py
+++ b/libpysal/graph/tests/test_builders.py
@@ -4,8 +4,8 @@ import numpy as np
 import pandas as pd
 import pytest
 from numpy.testing import assert_array_almost_equal
-from shapely import get_coordinates
 from scipy.sparse.csr import csr_matrix
+from shapely import get_coordinates
 
 from libpysal import graph
 
@@ -202,7 +202,9 @@ class TestKernel:
         sklearn = pytest.importorskip("sklearn")
         df = gpd.read_file(geodatasets.get_path("nybb"))
         df = df.to_crs(df.estimate_utm_crs())
-        distmat = csr_matrix(sklearn.metrics.pairwise.euclidean_distances(get_coordinates(df.centroid)))
+        distmat = csr_matrix(
+            sklearn.metrics.pairwise.euclidean_distances(get_coordinates(df.centroid))
+        )
         G = graph.Graph.build_kernel(distmat, metric="precomputed")
         expected = np.array(
             [
--- a/libpysal/graph/tests/test_builders.py
+++ b/libpysal/graph/tests/test_builders.py
@@ -6,7 +6,6 @@ import pytest
 from numpy.testing import assert_array_almost_equal
 from shapely import get_coordinates
 from scipy.sparse.csr import csr_matrix
-from sklearn.metrics.pairwise import euclidean_distances
 
 from libpysal import graph
 
--- a/libpysal/graph/tests/test_builders.py
+++ b/libpysal/graph/tests/test_builders.py
@@ -5,6 +5,7 @@ import pandas as pd
 import pytest
 from numpy.testing import assert_array_almost_equal
 from shapely import get_coordinates
+from scipy.sparse.csr import csr_matrix
 from sklearn.metrics.pairwise import euclidean_distances
 
 from libpysal import graph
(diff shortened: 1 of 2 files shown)


martinfleis:
Can we include a quick test?

knaaptime:
Yes, but wanted to gauge opinions on allowing ids in the _builders. I’d vote for that option which makes it easier to work with precomputed distmats, but not a dealbreaker

martinfleis:
>  opinions on allowing ids in the _builders

It is easy enough to allow that but it is a rabbit hole. What do we do in `plot` and `explore` then? Those assume the relationship between adjacency and original data via the index. Do we expose `ids` there as well? There may be other use cases like this.

I am fine with adding it but am slightly afraid about the impact this will have on the rest of the codebase, if we want to allow separation of the ids used in adjacency and the index of input dataframe / position of geometry or coords in an array.

knaaptime:
agree. I see all the issues and would rather keep things simple and avoid the can of worms. The only thing with the builder functions is it can make sense to do something like 

```
from sklearn.metrics.pairwise import euclidean_distances

gdf = gdf.set_index('meaningful_index')
distmat = euclidean_distances(get_coordinates(gdf.centroid))
g = graph.Graph.build_kernel(
        distmataph,
        metric="precomputed",
        bandwidth=2000,
        kernel=kernel,)

```

(where you have a more meaningful distmat there) so the alignment is correct with the gdf, but there's no way to tell the graph you want to use the meaningful index. You can reset the index on the gdf, or do something like Graph.from_adjacency(g.adjacency[['focal', 'neighbor']].replace(id_map)) or whatever

we could make it so ids is only considered when metric is precomputed? that's not my favorite, but could be a compromise

martinfleis:
Let's pick it up on Thursday during the call

knaaptime:
im fine passing on the ids arg (especially until there's more discussion) but until then, i'd consider this ready. the 3.12 failure is unrelated and the other is failing because the test requires sklearn

libpysal/graph/tests/test_builders.py
```diff
@@ -194,6 +198,38 @@ def setup_method(self):
         )
         self.gdf_str = self.gdf.set_index("placeid")
 
+    def test_kernel_precompute(self):
+        df = gpd.read_file(geodatasets.get_path("nybb"))
+        df = df.to_crs(df.estimate_utm_crs())
+        distmat = euclidean_distances(get_coordinates(df.centroid))
```
martinfleis, line 204: This does not test the change you have made. This test passes with the original code. I think that the only case where we need to check the shape instead of len is a sparse array or matrix, so we should test against that. `distmat` is a dense array which has `__len__`.

libpysal/graph/tests/test_builders.py
```diff
@@ -1,8 +1,12 @@
...
-
+import geodatasets
 import geopandas as gpd
+import numpy as np
 import pandas as pd
-import geodatasets
+import pytest
+from numpy.testing import assert_array_almost_equal
+from shapely import get_coordinates
+from sklearn.metrics.pairwise import euclidean_distances
```
martinfleis, line 8:
```suggestion
```
remove from here to make the no deps CI run

libpysal/graph/tests/test_builders.py
```diff
@@ -194,6 +198,38 @@ def setup_method(self):
         )
         self.gdf_str = self.gdf.set_index("placeid")
 
+    def test_kernel_precompute(self):
```
martinfleis, line 201:
```suggestion
    def test_kernel_precompute(self):
        sklearn = pytest.importorskip("sklearn")
```
Import via importorskip

libpysal/graph/tests/test_builders.py
martinfleis, line 204:
```suggestion
        distmat = sklearn.metrics.pairwise.euclidean_distances(get_coordinates(df.centroid))
```

knaaptime: lol oops

libpysal/graph/tests/test_builders.py
knaaptime: i did not know about importorskip. thats awfully handy

libpysal/graph/tests/test_builders.py
```diff
@@ -1,8 +1,12 @@
-import pytest
-
+import geodatasets
 import geopandas as gpd
+import numpy as np
 import pandas as pd
-import geodatasets
+import pytest
+from numpy.testing import assert_array_almost_equal
+from scipy.sparse.csr import csr_matrix
```
martinfleis, line 7:
```suggestion
from scipy.sparse import csr_matrix
```
Otherwise it raises ```DeprecationWarning: Please import `csr_matrix` from the `scipy.sparse` namespace; the `scipy.sparse.csr` namespace is deprecated and will be removed in SciPy 2.0.0.``` in CI.

martinfleis (approved):

jGaboardi (approved):

PR #617, opened 2023-10-31
[WIP] explore method for graph

knaaptime:
Screenshot 2023-10-31 at 12 22 10 AM

(the graph is actually based on network distance so that water observation is a known snapping issue :P)

### DIFF
--- a/libpysal/graph/_plotting.py
+++ b/libpysal/graph/_plotting.py
@@ -186,6 +186,7 @@ def _explore_graph(
     gdf = gdf.copy()
     gdf["id"] = gdf.index.values
     gdf = gdf.set_geometry(gdf.centroid)
+    gdf = gdf[["id", "geometry"]]
 
     if node_kws is not None:
         if "color" not in node_kws:
@@ -217,21 +218,19 @@ def _explore_graph(
         shapely.linestrings(np.hstack([origins, destinations]).reshape(-1, 2, 2)),
         crs=gdf.crs,
     )
-    edges = gpd.GeoDataFrame(adj, geometry=lines)
+    edges = gpd.GeoDataFrame(adj, geometry=lines)[
+        ["focal", "neighbor", "weight", "geometry"]
+    ]
 
-    m = (
-        edges[["focal", "neighbor", "weight", "geometry"]].explore(m=m, **edge_kws)
-        if m is not None
-        else edges[["focal", "neighbor", "weight", "geometry"]].explore(**edge_kws)
-    )
+    m = edges.explore(m=m, **edge_kws) if m is not None else edges.explore(**edge_kws)
 
     if nodes is True:
         if focal is not None:
             dests = gdf.loc[adj.neighbor].index.values
-            m = gdf[gdf.id.isin(dests)][["id", "geometry"]].explore(m=m, **node_kws)
+            m = gdf[gdf.id.isin(dests)].explore(m=m, **node_kws)
             if focal_kws is None:
                 focal_kws = {}
-            m = gdf[gdf["id"].isin(focal)][["id", "geometry"]].explore(m=m, **dict(node_kws, **focal_kws))
+            m = gdf[gdf["id"].isin(focal)].explore(m=m, **dict(node_kws, **focal_kws))
         else:
-            m = gdf[["id", "geometry"]].explore(m=m, **node_kws)
+            m = gdf.explore(m=m, **node_kws)
     return m
--- a/libpysal/graph/_plotting.py
+++ b/libpysal/graph/_plotting.py
@@ -229,7 +229,9 @@ def _explore_graph(
         if focal is not None:
             dests = gdf.loc[adj.neighbor].index.values
             m = gdf[gdf.id.isin(dests)][["id", "geometry"]].explore(m=m, **node_kws)
-            m = gdf[gdf["id"].isin(focal)][["id", "geometry"]].explore(m=m, **focal_kws)
+            if focal_kws is None:
+                focal_kws = {}
+            m = gdf[gdf["id"].isin(focal)][["id", "geometry"]].explore(m=m, **dict(node_kws, **focal_kws))
         else:
             m = gdf[["id", "geometry"]].explore(m=m, **node_kws)
     return m
--- a/libpysal/graph/_plotting.py
+++ b/libpysal/graph/_plotting.py
@@ -156,15 +156,18 @@ def _explore_graph(g, gdf, **kwargs):
         folium map
     """
     gdf = gdf.copy()
-    gdf['id']=gdf.index.values
-    m = gdf.reset_index()[['id', 'geometry']].set_geometry(gdf.centroid).explore(**kwargs)
+    gdf["id"] = gdf.index.values
+    gdf = gdf.set_geometry(gdf.centroid)
+    m = gdf[["id", "geometry"]].explore(**kwargs)
 
     adj = g.adjacency.reset_index()
 
-    origins = gdf.loc[adj.focal].centroid.geometry
-    destinations = gdf.loc[adj.neighbor].centroid.geometry
-    ods = zip(origins, destinations)
-    lines = gpd.GeoSeries([shapely.geometry.LineString(od) for od in ods], crs=gdf.crs)
+    origins = gdf.loc[adj.focal].get_coordinates().values
+    destinations = gdf.loc[adj.neighbor].get_coordinates().values
+    lines = gpd.GeoSeries(
+        shapely.linestrings(np.hstack([origins, destinations]).reshape(-1, 2, 2)),
+        crs=gdf.crs,
+    )
     edges = gpd.GeoDataFrame(adj, geometry=lines)
     edges.explore(m=m, **kwargs)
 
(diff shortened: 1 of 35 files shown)


martinfleis:
Shall we aim for API parity with static plots here?

libpysal/graph/_plotting.py
```diff
@@ -135,3 +136,36 @@ def _plot(
...
+    kwargs: additional keyword arguments passed to the geopandas explore
+        method. For example, to make the  nodes larger pass `marker_kwds={'radius':8}`
+
+    Returns
+    -------
+    folium.Map
+        folium map
+    """
+    gdf = gdf.copy()
+    gdf['id']=gdf.index.values
```
jGaboardi, line 159: Let's run `ruff` and `black`.

martinfleis:
@knaaptime a side note as I noticed the code in the screenshot above - for explore, CRS makes no difference. Map is always EPSG:3857 and is reprojected under the hood.

knaaptime:
> @knaaptime a side note as I noticed the code in the screenshot above - for explore, CRS makes no difference. Map is always EPSG:3857 and is reprojected under the hood.

Ah, good call. I get it mixed up with contextily where I need to set it

knaaptime:
> Shall we aim for API parity with static plots here?

Not sure where folks wanna take the api. Don’t have much of an opinion but figured this is enough to get us started

libpysal/graph/_plotting.py
```diff
@@ -135,3 +136,36 @@ def _plot(
...
+    m = gdf.reset_index()[['id', 'geometry']].set_geometry(gdf.centroid).explore(**kwargs)
```
martinfleis, line 160: Why do we need to assign index as a column (that could overwrite existing) and then reset? To have known names for subsetting?

libpysal/graph/_plotting.py
```diff
@@ -135,3 +136,36 @@ def _plot(
...
+
+    adj = g.adjacency.reset_index()
+
+    origins = gdf.loc[adj.focal].centroid.geometry
+    destinations = gdf.loc[adj.neighbor].centroid.geometry
```
martinfleis, line 165:
```suggestion
    origins = gdf.loc[adj.focal].centroid
    destinations = gdf.loc[adj.neighbor].centroid
```
I know this may be obsolete but centroid is always a GeoSeries.

libpysal/graph/_plotting.py
```diff
@@ -135,3 +136,36 @@ def _plot(
...
+    ods = zip(origins, destinations)
+    lines = gpd.GeoSeries([shapely.geometry.LineString(od) for od in ods], crs=gdf.crs)
```
martinfleis, line 167: Try avoiding this loop. This shall work

```py
origins = gdf.loc[adj.focal].centroid.get_coordinates().values
destinations = gdf.loc[adj.neighbor].centroid.get_coordinates().values
lines = gpd.GeoSeries(shapely.linestrings(np.hstack([origins, destinations]).reshape(-1, 2, 2)), crs=gdf.crs)
```

libpysal/graph/_plotting.py
```diff
@@ -135,3 +136,36 @@ def _plot(
...
+    folium.Map
+        folium map
+    """
+    gdf = gdf.copy()
+    gdf['id']=gdf.index.values
+    m = gdf.reset_index()[['id', 'geometry']].set_geometry(gdf.centroid).explore(**kwargs)
+
+    adj = g.adjacency.reset_index()
+
+    origins = gdf.loc[adj.focal].centroid.geometry
```
martinfleis, line 164: Can we assign `gdf.centroid` to a variable and reuse it? Now we call it three times which is actually even more per geometry. I know it is not expensive, but indexing is cheaper :).

libpysal/graph/_plotting.py
knaaptime: probably nonsense... it was late and i added 'id' back in later so you could see it in the popup :)

knaaptime:
Screenshot 2023-11-01 at 3 23 36 PM

knaaptime:
here's a simpler example. It's a touch verbose, i think that just is what it is. Should match the plot api now, i think

Screenshot 2023-11-01 at 4 38 52 PM

libpysal/graph/_plotting.py
```diff
@@ -136,3 +137,99 @@ def _plot(
...
+        edges[["focal", "neighbor", "weight", "geometry"]].explore(m=m, **edge_kws)
+        if m is not None
+        else edges[["focal", "neighbor", "weight", "geometry"]].explore(**edge_kws)
+    )
+
+    if nodes is True:
+        if focal is not None:
+            dests = gdf.loc[adj.neighbor].index.values
+            m = gdf[gdf.id.isin(dests)][["id", "geometry"]].explore(m=m, **node_kws)
+            m = gdf[gdf["id"].isin(focal)][["id", "geometry"]].explore(m=m, **focal_kws)
```
martinfleis, line 232:
```suggestion
            if focal_kws is None:
                focal_kws = {}
            m = gdf[gdf["id"].isin(focal)][["id", "geometry"]].explore(m=m, **dict(node_kws, **focal_kws))
```

This is to ensure that generic `node_kws` are applied to all nodes and `focal_kws` only adapt that. If both define the same, `focal_kws` will override `node_kws`. It is the same behaviour we have in `plot`.

martinfleis:
This is great and apart from that one suggestion works as expected. Nice! Do you want to give a go to tests (ideally mirroring those from `plot`) or shall I do that?

knaaptime:
@martinfleis i'll take the first pass, but would probably be speediest if we tag team again. I'll try and do a thorough first draft. Would be interested to cut RCs for lib and esda today for kicking the tires

knaaptime:
related, are there common patterns like [this one](https://github.com/pysal/esda/blob/3d405c43648cfe084dd7cc4302343930828b430d/esda/tests/test_moran.py#L369) that are becoming common enough to centralize in a testing module of lib? (a la numpy?). We now have explore in esda, pointpats, and here... its a tiny function so nbd, but there are probably other patterns used across the meta

martinfleis:
The original is from folium :D. I think it makes sense to centralise stuff as `assert_series_equal` that does multiple things but utils like this can easily live in each. But I don't mind if we move it to libpysal.testing and re-use it across pysal

knaaptime:
@martinfleis there's a first cut, but its nowhere near as robust as the plot tests. Not sure what else you want to test coming out of that string, but im satisfied...

knaaptime:
looks like this will also require a minimum pin for a recent geopandas

martinfleis:
We'll need to include a version of this as we currently plot both ij and ji. I will take back to this later today and expand CI a bit to ensure similar coverage we have in `plot`.

https://github.com/pysal/libpysal/blob/3976997887b5f0eabf1ba52d615e195c557f3776/libpysal/graph/_plotting.py#L106-L107

martinfleis:
Did a bit of refactor based on the code from `plot` but as a result we get the focal IDs for free so it simplifies the code a bit as well (no need to create that `adj` dataframe.

I think it is ready. It is smooooth!

knaaptime (commented):
sweet!

knaaptime:
i need one of you two to approve and merge though, :P

then lets do an RC?

libpysal/graph/tests/test_plotting.py
```diff
@@ -5,12 +5,13 @@
...
 
 from libpysal import graph
-
-matplotlib = pytest.importorskip("matplotlib")
+from libpysal.graph.tests.test_utils import fetch_map_string
 
 
 class TestPlotting:
     def setup_method(self):
+        _ = pytest.importorskip("matplotlib")
```
jGaboardi, line 13: Do we need to declare this underscore here? Why like in L262 below?

martinfleis (approved):

jGaboardi (approved):
All looks great. One minor comment.

libpysal/graph/tests/test_plotting.py
martinfleis: Sorry! Not really but it does not matter much I suppose. This does return a symlink.

martinfleis:
> then lets do an RC?

Do we need RC? Can we just do 4.9.0?

knaaptime:
we can just do 49 if you like

jGaboardi:
My opinion is that it's never a bad idea to do an RC just to make sure things are working properly. On the other hand, we can delete tags & releases if something is broken.

martinfleis:
I mean, what do you expect to test by RC in this case?

knaaptime:
the only thing with an "unnecessary" RC is that the changelog tooling techninically fails and you need to copy over the changelog

jGaboardi:
Making sure the actions (doc build & release) still work as they should. But I won't be stickler on that. Let's just try for the `v4.9.0`

martinfleis:
Hold on with that, there's a bug in this PR

knaaptime:
aw crap

knaaptime:
looks like we're doing a 491 today 🙃

martinfleis:
it'll be 4.9.1 then :D

jGaboardi:
I done told yall! 🤣

martinfleis:
Sorry @jGaboardi, we should listen to you more :D

PR #618, opened 2023-10-31
lint `graph/_*.py`

jGaboardi:

This PR:
* formats and lints `libpysal/graph/_*.py`
* rename `_set_ops._Set_Mixin` to `_set_ops.SetOpsMixin`

### DIFF
--- a/libpysal/graph/_kernel.py
+++ b/libpysal/graph/_kernel.py
@@ -1,17 +1,16 @@
 import numpy
-from scipy import sparse, optimize, spatial, stats
+from scipy import optimize, sparse, spatial, stats
 
 from ._utils import (
-    _validate_geometry_input,
     _build_coincidence_lookup,
-    _induce_cliques,
     _jitter_geoms,
-    _sparse_to_arrays,
     _resolve_islands,
+    _sparse_to_arrays,
+    _validate_geometry_input,
 )
 
 try:
-    from sklearn import neighbors, metrics
+    from sklearn import metrics, neighbors
 
     HAS_SKLEARN = True
 except ImportError:
@@ -50,7 +49,7 @@ def _boxcar(distances, bandwidth):
     return r
 
 
-def _identity(distances, bandwidth):
+def _identity(distances, _):
     return distances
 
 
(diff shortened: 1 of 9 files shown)


martinfleis:
> I can add the remaining 2 files (_{contiguity, _kernel}.py) in here

include those.

> What do we think about renaming _set_ops._Set_Mixin --> _set_ops._SetMixin? Or even something like _set_ops .SetOpsMixin to conform with naming conventions? 

Whatever, this is fully private. Happy to stick to strict CamelCase if it lives in the private module anyway

knaaptime:
swap it james. I dont like ignoring pep8 anyway. i'll be psyched once we finally get rid of those underscores in esda

libpysal/graph/_kernel.py
```diff
@@ -300,7 +306,7 @@ def _prepare_tree_query(coordinates, metric, p=2):
         return tree(coordinates, metric=metric, **dist_kwds).query
     else:
         if metric in ("euclidean", "manhattan", "cityblock", "minkowski"):
-            from scipy.spatial import KDTree as tree
+            from scipy.spatial import KDTree as tree  # noqa N813
 
             tree_ = tree(coordinates)
```
martinfleis, line 311:
```suggestion
            tree_ = spatial.KDTree(coordinates)
```

libpysal/graph/_parquet.py
```diff
@@ -1,8 +1,9 @@
-import libpysal
 import json
 
+import libpysal
+
 
-def _to_parquet(G, destination, **kwargs):
+def _to_parquet(G, destination, **kwargs):  # noqa N803
```
martinfleis, line 6: We can use something else than `G` here, to comply.

libpysal/graph/_plotting.py
```diff
@@ -4,7 +4,7 @@
 
 
 def _plot(
-    G,
+    G,  # noqa N803
```
martinfleis, line 7: As above.

libpysal/graph/_utils.py
```diff
@@ -137,7 +137,7 @@ def _build_coincidence_lookup(geoms):
     """
     Identify coincident points and create a look-up table for the coincident geometries.
     """
-    valid_coincident_geom_types = set(("Point",))
+    valid_coincident_geom_types = set(("Point",))  # noqa C405
```
martinfleis, line 140: I remember some issue with this but why do we add `noqa` instead of `{("Point",)}`? The latter does not work?

libpysal/graph/_utils.py
```diff
@@ -224,7 +224,7 @@ def _validate_sparse_input(sparse, ids=None):
     return _sparse_to_arrays(sparse, ids)
 
 
-def _vec_euclidean_distances(X, Y):
+def _vec_euclidean_distances(X, Y):  # noqa N803
```
martinfleis, line 227: Can we just use lowercase x and y?

martinfleis (commented):
Two more notes, sorry

libpysal/graph/_utils.py
jGaboardi: I tried `{("Point",)}` and it broke CI. But let me try again to double check.

libpysal/graph/_utils.py
jGaboardi: I'll go with `x_vec, y_vec`. I think easier to read as the difference between scalar and vector.

libpysal/graph/_utils.py
jGaboardi: OK, se we get a ton of these: `ValueError: coindicence checks are only well-defined for geom_types: {('Point',)}`

For example:
```
libpysal/graph/tests/test_set_ops.py::Test_Set_Ops::test___len__ - ValueError: coindicence checks are only well-defined for geom_types: {('Point',)}
```

And then we also get several: `AssertionError: Regex pattern did not match.`

For example:
```
libpysal/graph/tests/test_triangulation.py::Test_Coincident::test_voronoi_error - AssertionError: Regex pattern did not match.
```

jGaboardi: Where the `valid_geometry_types` come from [`_VALID_GEOMETRY_TYPES`](https://github.com/pysal/libpysal/blob/ed98cbd797beaa03fcb479e2f4fa915266b99281/libpysal/graph/_triangulation.py#L32)

martinfleis: Okay, leave it.

martinfleis (approved):

PR #619, opened 2023-10-31
lint & format `graph/tests/*.py`

jGaboardi:

This PR formats and lints `libpysal/graph/tests/*.py`


libpysal/graph/tests/test_utils.py
```diff
@@ -112,10 +107,17 @@ def test_validate_raises(
     contiguity_types=contiguity_types,
     triang_types=triang_types,
 ):
+    # kernels
     with pytest.raises(ValueError):  # no lines for kernels
         _validate_geometry_input(rivers, valid_geometry_types=kernel_types)
     with pytest.raises(ValueError):  # no polygons for kernels
         _validate_geometry_input(columbus, valid_geometry_types=kernel_types)
+    # triangulation
```
jGaboardi, line 115: @knaaptime @martinfleis 

Before I got too deep in the weeds, thought I'd get your opinions on this – `ruff` was pitching a fit because `triang_types` wasn't actually being used here. Since `triang_types` are currently the same as `kernel_types`, this maybe isn't a big deal. However, to be thorough I thought it best to add explicit tests for `triang_types`. Thoughts on keeping the explicit testing or removing `triang_types` from this file altogether?

martinfleis: It would be nice to have additional tests for this rather than removing it.

jGaboardi:
@martinfleis This one's a bit bigger diff than ideal, but the vast majority are upper->lowercase variable renaming in tests.

libpysal/graph/tests/test_set_ops.py
```diff
@@ -96,10 +97,10 @@ def test___gt__(self):
 
     def test___eq__(self):
         assert self.distance2500 == self.distance2500.copy()
-        assert not self.distance2500 == self.distance2500_id
+        assert self.distance2500 != self.distance2500_id
 
     def test___ne__(self):
-        assert not self.distance2500 != self.distance2500.copy()
+        assert self.distance2500 == self.distance2500.copy()
         assert self.distance2500 != self.distance2500_id
```
martinfleis, line 104: Can you revert these changes? This intentionally tests implementation of `==` and `!=`.

martinfleis (commented):
Just one note. Thanks!

libpysal/graph/tests/test_set_ops.py
jGaboardi: Ah ha. OK, will do.

martinfleis (approved):

PR #621, opened 2023-11-01
TST: ignore pandas dev blockmanager warnings

martinfleis:
This should fix the failures on dev that are to be resolved in upstream.


jGaboardi (approved):
:tada:

PR #625, opened 2023-11-02
fix fileio regression

### LINKED ISSUE #624: regression in libpysal.io.fileio.FileIO
#620 [introduced a regression](https://github.com/pysal/libpysal/pull/620/files#diff-63caaa2e22b46da7ffb19263d6483704698216a98ab2ec087ea6ba01f19f9ea1L93) (for example [in `spaghetti`](https://github.com/pysal/spaghetti/actions/runs/6726774671/job/18283603971#step:7:296)), however, I'm not sure if it's something we should fix in `libpysal` or in affected downstream packages. The change was prompted by `ruff` whereby:

```python
return open(dataPath, mode)
``` 

was replaced by:

```python
with open(dataPath, mode) as file:
    return file
```

for proper context handling.

This broke no tests in `libpysal` and so far I can only find that it has broken tests in `spaghetti`.

jGaboardi:
The PR resolves #624

### DIFF
--- a/libpysal/io/fileio.py
+++ b/libpysal/io/fileio.py
@@ -14,7 +14,7 @@ Readers and Writers will mimic python file objects.
 """
 
 
-# ruff: noqa: ARG002, N801, N802, N803, N806
+# ruff: noqa: ARG002, N801, N802, N803, N806, SIM115
 
 __author__ = "Charles R Schmidt <schmidtc@gmail.com>"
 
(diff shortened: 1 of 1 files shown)


martinfleis:
Have you tested it against `spaghetti`?

jGaboardi:
> Have you tested it against `spaghetti`?

Yes, last night. But let me triple check again right now.

jGaboardi:
Yep, can confirm passing in `spaghetti` with the `jGaboardi:io_regresion` branch of `libpysal`

martinfleis (approved):

PR #627, opened 2023-11-03
pass kwargs to explore

martinfleis:
Follow-up on #617

We were not able to change anything about the map as we forgot to pass kwargs though like we did in `esda`. But stuff like `tiles`  or `prefer_canvas` shall work out of the box.

### DIFF
--- a/libpysal/graph/_plotting.py
+++ b/libpysal/graph/_plotting.py
@@ -149,6 +149,7 @@ def _explore_graph(
     node_kws=None,
     focal_kws=None,
     m=None,
+    **kwargs,
 ):
     """Plot graph as an interactive Folium Map
 
@@ -177,6 +178,10 @@ def _explore_graph(
         passing a subset of nodes with the `focal` argument
     m : Folilum.Map, optional
         folium map objecto to plot on top of, by default None
+    **kwargs : dict, optional
+        additional keyword arguments are passed directly to geopandas.explore, when
+        ``m=None`` by default None
+
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


knaaptime:
cool. The way i'd handled that before was to just cheat and inculude extra args like tiles in the kwargs for one of the earlier maps (like stick them in the edge_kws). This is definitely preferable

martinfleis:
A side note - I also realised that we can directly plot weight. Neat.
Screenshot 2023-11-03 at 23 08 10

knaaptime:
oh dope!

martinfleis:
> The way i'd handled that before was to just cheat and inculude extra args like tiles in the kwargs for one of the earlier maps 

Yeah but that is very unintuitive and requires you to know the implementation.

knaaptime:
>Yeah but that is very unintuitive and requires you to know the implementation.
💯

>A side note - I also realised that we can directly plot weight. Neat.

does that mean that works for the static plots too?

knaaptime (approved):

martinfleis:
> does that mean that works for the static plots too?

No, because we don't reuse `GeoDataFrame.plot` there but built collections directly. Going via GeoPandas would cause significant performance hit. We could potentially map colours ourselves though...

jGaboardi (approved):

martinfleis:
@knaaptime would you like to cut 4.9.1? :)

knaaptime:
As soon as I’m out of the Getis lecture at sdsu

--

Elijah Knaap, Ph.D.
Senior Research Scientist & Associate Director
Center for Open Geographical Science
San Diego State University
knaaptime.com | @knaaptime

martinfleis:
@knaaptime I suggest we first figure out what caused this conda-forge error https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=817319&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3

cc @jGaboardi 

It is midnight here so that is not a task for me today :)

knaaptime:
so, i dunno why thats failing. The error seems to come from xarray not being available when weights is imported, but afaict xarray is never imported in the weights module except inside a function-level 'try' block.

...that said, it looks like we *do* define xarray as a hard dependency in pyproject.toml, but not in the conda forge recipe. A quick fix would be to include xarray in conda, but do we need it? Also not clear why the error is actually materializing here

PR #634, opened 2023-11-09
ENH: Graph.eliminate_zeros, refactor isolates

### LINKED ISSUE #633: include a method to remove non-isolate zero-weight edges
I have a use case of subsetting the Graph but to ensure all focals are preserved, the only safe option is to set weight of unwanted edges to 0. It would be nice to be able to simply clean the adjacency and remove zero-weight edges to save memory. I can roundtrip via sparse but I need to know this trick exists and it is not necessarily the cheapest thing to do.

martinfleis:
Closes #633

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -972,9 +972,13 @@ class Graph(SetOpsMixin):
         pandas.Index
             Index with a subset of observations that do not have any neighbor
         """
-        nulls = self._adjacency[self._adjacency == 0].reset_index(level=1)
+        nulls = self._adjacency[self._adjacency == 0]
         # since not all zeros are necessarily isolates, do the focal == neighbor check
-        return nulls[nulls.index == nulls.neighbor].index.unique()
+        return (
+            nulls[nulls.index.codes[0] == nulls.index.codes[1]]
+            .index.get_level_values(0)
+            .unique()
+        )
 
     @cached_property
     def unique_ids(self):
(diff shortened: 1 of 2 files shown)


libpysal/graph/base.py
```diff
@@ -972,9 +972,13 @@ def isolates(self):
         pandas.Index
             Index with a subset of observations that do not have any neighbor
         """
-        nulls = self._adjacency[self._adjacency == 0].reset_index(level=1)
+        nulls = self._adjacency[self._adjacency == 0]
         # since not all zeros are necessarily isolates, do the focal == neighbor check
-        return nulls[nulls.index == nulls.neighbor].index.unique()
+        return (
+            nulls[nulls.index.codes[0] == nulls.index.codes[1]]
```
martinfleis, line 978: This is indeed faster than the original solution. On exploded `geoda tampa1` dataset `84.8 µs ± 5.65 µs` vs `253 µs ± 6.75 µs`.

jGaboardi (approved):

libpysal/graph/base.py
```diff
@@ -1337,6 +1341,24 @@ def explore(
...
+        Returns
+        -------
+        Graph
+            subset of Graph with zero-weight edges eliminated
+        """
+        # get a mask for isolates
+        isolates = self._adjacency.index.codes[0] == self._adjacency.index.codes[1]
+        # substract isolates from mask of zeros
+        zeros = (self._adjacency == 0) != isolates
+        return Graph(self._adjacency[~zeros])
```
martinfleis, line 1360: This should get `is_sorted=True` once #635 is in. So if we can merge that one first, I'll update this.

sjsrey (approved):

PR #635, opened 2023-11-09
fast-path for initialisation of Graph for adjacency that is known to be sorted

### LINKED ISSUE #632: allow quick pass through the Graph constructor
Graph constructor currently always ensures sorting.
https://github.com/pysal/libpysal/blob/44204b599f9e6d04540f76dd7abcfd3c187f93ec/libpysal/graph/base.py#L95-L96

While needed, it is relatively costly operation, esp. on larger graphs (I am now working on one with 120mil edges and it quite noticeable). When doing stuff like transformation, we know that the order does not change, only the weight. We should implement a fast path that skips sorting for adjacencies that are known to be sorted.

martinfleis:
Closes #632 

When we know that the adjacency series is sorted as it is supposed to, we can skip sorting in `__init__`. This is particularly noticeable with large graphs of tens of millions edges. 

There may be some cases that we know are sorted and could use this. I thought that everything that goes through `_resolve_islands` is sorted but if I pass `is_sorted=True` everywhere, I get some strange failures. So better be safe than sorry in this case. Sorting is sensitive here.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -271,7 +271,7 @@ class Graph(SetOpsMixin):
             libpysal.graph.Graph based on sparse
         """
 
-        return cls.from_arrays(*_sparse_to_arrays(sparse, ids), is_sorted=True)
+        return cls.from_arrays(*_sparse_to_arrays(sparse, ids))
 
     @classmethod
     def from_arrays(cls, focal_ids, neighbor_ids, weight, **kwargs):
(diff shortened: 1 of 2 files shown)


libpysal/graph/base.py
```diff
@@ -261,7 +274,7 @@ def from_sparse(cls, sparse, ids=None):
         return cls.from_arrays(*_sparse_to_arrays(sparse, ids))
 
     @classmethod
-    def from_arrays(cls, focal_ids, neighbor_ids, weight):
+    def from_arrays(cls, focal_ids, neighbor_ids, weight, **kwargs):
```
martinfleis, line 277: This is not used at the moment but allows us to pass `is_sorted=True` to any constructor using `from_arrays`. So nearly all of them. But we first need to be sure that the result is, in fact, always sorted. That seems to include how pandas encodes ids within `codes` in MultiIndex. So unless we are certain, it is better to sort in `__init__`.

jGaboardi (approved):
lgtm

knaaptime:
makes sense to me

PR #630, opened 2023-11-05
lint & format – `io/iohandlers/*`

jGaboardi:
@martinfleis small fix left here inside `libpysal/graph/tests/test_utils.py` from https://github.com/pysal/libpysal/commit/dd4b7d228f976e734abf432c34d94b19a026cfd5 last week. Merge here or open a different PR?

***EDIT***

Broken out into parts:
* #637 
* #638 
* #641 
* #642 
* #643 

I'll close this PR once completed in broken out sections.

---------------------

xref https://github.com/pysal/libpysal/issues/589

This PR formats and lints `libpysal/io/iohandlers/*`

***EDIT***

@martinfleis This one ended up with a much larger diff that I had thought it would. I can split into 2 or 3 separate PRs if that would be better for you. Let me know.

PR #637, opened 2023-11-09
lint & format `io/iohandlers{a-d}.py`

jGaboardi:

This PR
* breaks out #630 (part 1 of $n$)
* formats and lints `io/iohandlers{a-d}.py`

### DIFF
--- a/libpysal/io/iohandlers/csvWrapper.py
+++ b/libpysal/io/iohandlers/csvWrapper.py
@@ -1,7 +1,9 @@
-from .. import tables
+# ruff: noqa: N801, N802, N806, N999, SIM115
+
+
 import csv
 
-from typing import Union
+from .. import tables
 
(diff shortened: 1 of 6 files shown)


libpysal/io/iohandlers/csvWrapper.py
```diff
@@ -1,7 +1,9 @@
-from .. import tables
+# ruff: noqa: N801, N802, N806, N999, SIM115
```
martinfleis, line 1: N999 for which reason?

jGaboardi: File name is `csvWrapper.py`, which triggers that. I am on the fence about renaming files... may introduce cascading unknowns...

martinfleis: That is a public API, we need to keep it... for now.

martinfleis (approved):

PR #638, opened 2023-11-09
lint & format `io/iohandlers{g-m}`.py

jGaboardi:

This PR:

* breaks out https://github.com/pysal/libpysal/pull/630 (part 2 of $n$)
* formats and lints `io/iohandlers{g-m}.py`

### DIFF
--- a/libpysal/io/iohandlers/geoda_txt.py
+++ b/libpysal/io/iohandlers/geoda_txt.py
@@ -1,31 +1,31 @@
+# ruff: noqa: A003, N802, N806, SIM115
+
 from .. import tables
 
 __author__ = "Charles R Schmidt <schmidtc@gmail.com>"
 __all__ = ["GeoDaTxtReader"]
 
-from typing import Union
-
 
 class GeoDaTxtReader(tables.DataTable):
     """GeoDa Text File Export Format.
-    
+
     Examples
     --------
-    
+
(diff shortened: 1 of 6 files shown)


libpysal/io/iohandlers/geoda_txt.py
```diff
@@ -1,31 +1,31 @@
+# ruff: noqa: A003, N802, N806, SIM115
```
martinfleis, line 1: A003 is not optimal to silence and for sure not globally. I suppose we can't refactor the stuff, right? Can we at least to inline `noqa`?

jGaboardi: `A003` is being triggered [here](https://github.com/pysal/libpysal/blob/79c4f821354c84b32883dc9b1c5061753d43f91a/libpysal/io/iohandlers/geoda_txt.py#L34) for `__doc__`. 2 choices here:
1. add an inline `noqa`
2. remove the offending declaration and add `GeoDaTxtReader.__doc__ = tables.DataTable.__doc__` after `GeoDaTxtReader` is declared.

Probably (2) is best, no?

martinfleis: Yes, 2 would be ideal.

martinfleis (approved):

Issue #631, opened 2023-11-06
Release coordination


martinfleis:
Recent changes that made it to 4.9.2 caused a lot of trouble at many places.

- the change broke spreg (https://github.com/pysal/spreg/pull/129) which wrongly did star imports from libpysal.common and used numpy imported from there instead of their own `import numpy as np` code. While we noticed this immediately, we released 4.9.2 without keeping track of these issues which then [broke production environments](https://github.com/martinfleis/sds/actions/runs/6768436793/job/18392870043).
- we have not updated requirements on conda-forge (https://github.com/conda-forge/libpysal-feedstock/pull/23) which resulted in 4.9.2 being installed in Python 3.9 and 3.8. environments while being incompatible with those (https://github.com/pysal/momepy/actions/runs/6764819962/job/18392907589, https://github.com/pysal/pointpats/actions/runs/6764866388/job/18383734872 and probably the rest of the ecosystem). I have updated the recipe (https://github.com/conda-forge/libpysal-feedstock/pull/24) and added a bot checking this but 4.9.2 build 0 is still installable. That is being patched (https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/585).
- The release of 4.9 was also a bit bumpy, that is why there is 4.9.2 already.

Considering all of this, I believe that we need to put in place additional checks before making a release of libpysal. A subset of that could be also applied elsewhere, to ensure that we don't break our own ecosystem.

knaaptime:
as a first cut, we should probably make sure that all the subpackages test against libpysal@main (like [esda](https://github.com/pysal/esda/blob/3d405c43648cfe084dd7cc4302343930828b430d/ci/311-DEV.yaml#L30)), and that downstream tests pass before a new libpysal is released

knaaptime:
> we have not updated requirements on conda-forge (https://github.com/conda-forge/libpysal-feedstock/pull/23) which resulted in 4.9.2 being installed in Python 3.9 and 3.8. environments while being incompatible with those 

i think the other way to handle this is to [turn off automerge](https://github.com/conda-forge/libpysal-feedstock/pull/24/files). While i like automerge, there's no way to control something like the minimum dependency change we just encountered (unless you catch the recipe PR before the bot merges). When we issue a new release, we the CI can guarantee the dependencies are correct in the source code, but theres no way to force the conda-forge recipe to use those deps

martinfleis:
Automerge wasn't on.

jGaboardi:
[...]

PR #639, opened 2023-11-10
GHA: include reverse dependency testing

martinfleis:
This is very much WIP because to make this work we need to ensure that the state of CI in downstream PySAL packages is as it should be. 

Once this works, it ensures that libpysal release does not break anything in the rest of the federation (xref #631). I'll see if there's some more work needed on the side of the action (I created the beta version during the Scientific Python summit, so have admin rights over it). But it will likely require changes in downstream (e.g. you can see that `tobler` does not properly skip H3 tests etc.).

The trigger is temporary to make sure we can properly test it here. We can then discuss when this should run but it should run before any release for sure and in the optimal situation should be required to pass for a release.

### DIFF
--- a/.github/workflows/reverse.yml
+++ b/.github/workflows/reverse.yml
@@ -34,11 +34,12 @@ jobs:
               hdbscan
               pandana
               astropy
+              geodatasets
             install_pip: >-
               opencv-contrib-python
               KDEpy
             installation_command: >-
-              pip install .; python -c 'import libpysal; libpysal.examples.fetch_all()'
+              pip install .; python -c 'import libpysal; libpysal.examples.builtin_datasets.keys()'
             fail_on_failure: false
             verbose: true
             parallel: false
--- a/.github/workflows/reverse.yml
+++ b/.github/workflows/reverse.yml
@@ -27,6 +27,9 @@ jobs:
               h3-py
               hdbscan
               pandana
+            install_pip: >-
+              opencv-contrib-python
+              KDEpy
             installation_command: >-
               pip install .; python -c 'import libpysal; libpysal.examples.fetch_all()'
             fail_on_failure: false
--- a/.github/workflows/reverse.yml
+++ b/.github/workflows/reverse.yml
@@ -27,6 +27,7 @@ jobs:
               h3-py
               hdbscan
               pandana
+              astropy
             install_pip: >-
               opencv-contrib-python
               KDEpy
@@ -34,3 +35,4 @@ jobs:
               pip install .; python -c 'import libpysal; libpysal.examples.fetch_all()'
             fail_on_failure: false
             verbose: false
+            parallel: false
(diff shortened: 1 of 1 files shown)


martinfleis:
So far, this goes well 😂

```sh
======================= reverse dependency tests summary =======================
PASSED: 
FAILED: esda, geosnap, giddy, inequality, mapclassify, mgwr, momepy, pointpats, region, segregation, spaghetti, spglm, spint, splot, spopt, spreg, spvcm, tobler
```

martinfleis:
It seems that built-in datasets in `libpysal.examples` are not available for some reason I am not able to figure out. Any guesses?

jGaboardi (approved):
Amazing work, @martinfleis!



As a followup, would it be prudent to migrate [`martinfleis/reverse-dependency-testing`](https://github.com/martinfleis/reverse-dependency-testing/) into the `pysal` org?

martinfleis:
> would it be prudent to migrate [martinfleis/reverse-dependency-testing](https://github.com/martinfleis/reverse-dependency-testing/) into the pysal org?

That is just a fork so I can work on it. It lives here https://github.com/scientific-python/reverse-dependency-testing so there will be no moving. Just a merge of my branch to main once I fix the parallelisation settings that doesn't seem to work now.

jGaboardi:
Should we merge this prior to merging #638 ?

martinfleis:
No, let me merge the changes of the action before that, so we link to scientific-python repo. But we should merge it before the next release to ensure any of the downstream test failures are not caused by us.

We should also go through them and report to downstream if they need to be fixed.

There is also a number of tests failing because the test files are not available in the installed version of the packages. We can either fetch those over wire or decide on a federation-wide pytest.mark to be able to skip them here.

martinfleis:
Questions here:

Do we want the verbose output or not? Maybe yes until we resolve downstream issues?
When should this run?

jGaboardi:
> Do we want the verbose output or not? Maybe yes until we resolve downstream issues?

I agree. At least verbose for now, but possibly in the future? Perhaps logging that provides a detailed report if failures occur? No idea if that's possible or not.

> When should this run?

Do you mean when should it be triggered/scheduled? This is good question since it takes quite a while to run... It would be nice to have on each PR, but that may not be feasible. I'd say at least 2x a month though for a schedule?

martinfleis:
> Perhaps logging that provides a detailed report if failures occur? No idea if that's possible or not.

No idea how would I do that. Pytest can produce some JSON reports but how to create an artefact from those is beyond my current knowledge. I'd skip that for now.

> Do you mean when should it be triggered/scheduled? This is good question since it takes quite a while to run... It would be nice to have on each PR, but that may not be feasible. I'd say at least 2x a month though for a schedule?

What about every push to main and then together with the scheduled CI? I would not run it on every push to a branch/PR but main might be good.

jGaboardi:
> No idea how would I do that. ... I'd skip that for now.

Same and agreed.

> What about every push to main and then together with the scheduled CI?

This sounds good to me.

martinfleis:
Okay. Gonna wait for https://github.com/scientific-python/reverse-dependency-testing/issues/1 and then update the action and merge.

PR #640, opened 2023-11-10
ENH: add subgraph method to Graph to get subsets

martinfleis:
I have found myself in a need to create subsets of a large graph covering only specific portions of my geometries. Hence I have developed a method to create a subgraph since making a subset of adjacency and passing that to a constructor discards isolates.

### DIFF
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -246,7 +246,7 @@ def _resolve_islands(heads, tails, ids, weights):
     Induce self-loops for a collection of ids and links describing a
     contiguity graph. Induced self-loops will have zero weight.
     """
-    islands = np.setdiff1d(ids, heads)
+    islands = pd.Index(ids).difference(pd.Index(heads))
     if islands.shape != (0,):
         heads = np.hstack((heads, islands))
         tails = np.hstack((tails, islands))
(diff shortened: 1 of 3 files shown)


libpysal/graph/_utils.py
```diff
@@ -246,7 +246,7 @@ def _resolve_islands(heads, tails, ids, weights):
     Induce self-loops for a collection of ids and links describing a
     contiguity graph. Induced self-loops will have zero weight.
     """
-    islands = np.setdiff1d(ids, heads)
+    islands = pd.Index(ids).difference(pd.Index(heads))
```
martinfleis, line 249: This turned out to be several orders of magnitude faster for large string indices.

jGaboardi (approved):

knaaptime:
cool. I wonder if we can use this to enhance https://github.com/pysal/esda/pull/259 as discussed over there (e.g by computing the largest range first, then successively cutting down the graph). Seems like you'd still need a tree query to get the indices though?

martinfleis:
> I wonder if we can use this to enhance https://github.com/pysal/esda/pull/259 as discussed over there (e.g by computing the largest range first, then successively cutting down the graph). Seems like you'd still need a tree query to get the indices though?

I don't think so. This aims to create a subgraph based on a subset of focals. In correlogram, you need to keep the same focals but cut their neighbors. 

What you could doe once #635 is in is something like this:

```py
for i in distances:
    adj = graph_w_distance.adjacency.copy()
    adj[adj > i] = 0
    smaller_graph = graph.Graph(adj, is_sorted=True).transform("r")
    # compute stats using smaller_graph
```
This assumes that `graph_w_distance` has weight == distance. Right now, there is the sorting bottleneck in this code but after #635 that will be gone.

ljwolf (approved):

PR #642, opened 2023-11-11
lint & format `io/iohandlers{t-w}`.py

jGaboardi:

This PR:

* breaks out https://github.com/pysal/libpysal/pull/630 (part 4 of $n$)
* formats and lints `io/iohandlers{t-w}.py`

@martinfleis I made in the note in the original PR before breaking out --> [see here](https://github.com/pysal/libpysal/pull/630#discussion_r1382624998). I think we can at least comment out that section if we don't remove entirely.


martinfleis:
Try commenting it out. The rest looks fine.

martinfleis (approved):

Issue #573, opened 2023-09-22
BUG: relative neighbourhood depends on the order of observations


martinfleis:
The graph coming from the relative neighbourhood depends on the order of observations in the df. I guess that should not happen.

```py
import geodatasets
import geopandas

from libpysal.graph._triangulation import _relative_neighborhood

stores = geopandas.read_file(geodatasets.get_path("geoda liquor_stores")).explode(
    index_parts=False
)
stores_unique = stores.drop_duplicates(subset="geometry")

for _ in range(5):
    data = stores_unique.sample(frac=1)
    head, tail, weight = _relative_neighborhood(data)
    print(head.shape)
```
```
(3360,)
(3338,)
(3360,)
(3346,)
(3342,)
```

This issue is also present in the implementation in `weights`, not only the new one.

```py
for _ in range(5):
    data = stores_unique.sample(frac=1)
    W = weights.Relative_Neighborhood.from_dataframe(data)
    print(W.to_adjlist().shape)
```
```
(3352, 3)
(3350, 3)
(3340, 3)
(3344, 3)
(3352, 3)
```

ljwolf:
Wow indeed should not happen. 

I will get a look next week.

ljwolf:
I think the issue is that the dkmax search has to iterate over all pairs, rather than terminating early.

martinfleis:
Closed by https://github.com/pysal/libpysal/pull/647

PR #647, opened 2023-11-13
support cliques as a resolver of coincident points in Graph

knaaptime:
i think we were just missing a couple parens


martinfleis:
The issue, as far as I remember it, was that `_induce_cliques` was not working properly. That is why I added `NotImplementedError`. I'd be happy to allow it again but the PR needs to come with extensive tests to ensure it all works as it is supposed to :).

ljwolf:
I think there was an [issue after a bad rebase.](https://github.com/pysal/libpysal/pull/554#issuecomment-1695680478)

I can get some attention on it this week to fix.

ljwolf:
I'm pretty certain that the issue was induced when we [switched from focal-indexed adjacency to non-indexed adjacency.](https://github.com/knaaptime/libpysal/pull/40). @knaaptime, I've sent a PR to your clique branch with the change.

The dual merge in [lines 76-82](https://github.com/knaaptime/libpysal/blob/cliques/libpysal/graph/_utils.py#L76) was supposed to match up the deduped ids back onto the clique ids, and then explode the cliques into their own links. This only works when the `adjtable` is indexed on `focal`. 

With this fix, tests still fail, but I think [the test answers are incorrect](https://github.com/knaaptime/libpysal/blob/cliques/libpysal/graph/tests/test_triangulation.py#L351)? In the tests, 0 and 4 are coincident. If 0 and 4 are coincident, then every link and backlink containing 0 *should also* contain 4. So, for example, link `(5,0)` is missing, as is link `(2,4)`.

So, the correct answer, I think, is: 

```python
exp_heads = numpy.array([0, 0, 0, 0, 1, 1, 1, 1, 1, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, 5, 5,
       5, 5])
exp_tails = numpy.array([1, 3, 4, 5, 0, 2, 3, 4, 5, 1, 3, 5, 0, 1, 2, 4, 0, 1, 3, 5, 0, 1,
       2, 4])
```

The expand cliques correctly recovers this.

ljwolf:
OK, after some further testing, I think I've confirmed the answers need to change, but I had to make sure to re-set the working values (like, `coordinates, ids, geoms, heads, tails, weights`) after expanding the cliques. 

I will change the tests, but the correct edge table is in the comment above.

ljwolf:
This now also addresses #573. The fix was as described, but also exposed a few other minor issues in the implementation that have been fixed. 

example_output

ljwolf:
Triangulation is done. I'm now verifying the usage in `_knn()`/`_kernel()`, so still WIP.

ljwolf:
`_knn`/`_kernel` is done. Currently, I'm getting local failures in the round-tripping from sparse? The adjacency tables are the same, but comparing the graphs by equality is not. I'm not sure how/where this would have been introduced by this PR, and it's currently failing on main?

ljwolf:
This is ready for review.

martinfleis:
> Currently, I'm getting local failures in the round-tripping from sparse? The adjacency tables are the same, but comparing the graphs by equality is not. I'm not sure how/where this would have been introduced by this PR, and it's currently failing on main?

Could it be some issue with your local env? It seems to be work as expected on CI.

We have some issue with coincident points in the env with the minimum supported versions of dependencies for some reason...

ljwolf:
then it must be something in my local environment. I will check out the failures in py3.10. I think that's likely a numba issue.

ljwolf:
Not a numba issue, but rather needed a bit more than [4091467](https://github.com/pysal/libpysal/pull/647/commits/4091467aaa8bbcf29d3164bed359b58a6a1b4489) to fully fix #573.

ljwolf:
Tests are passing, ready for review!

ljwolf (approved):

ljwolf:
@knaaptime I can't assign you as a reviewer, but you probably should be one of the reviewers anyway, having motivated me to revisit this 😄

knaaptime (commented):
you rule. But i also cant approve, apparently, since it's technically "my" pr 

so, approved. :P

martinfleis (approved):
I can approve :)

PR #648, opened 2023-11-13
lint & format `weights/*.py part 1

jGaboardi:
This PR:
* formats and lints `weights/*.py` –  part 1 of $n$

### DIFF
--- a/libpysal/weights/adjtools.py
+++ b/libpysal/weights/adjtools.py
@@ -159,8 +159,8 @@ def adjlist_map(
     funcs=(np.subtract,),
     W=None,
     alist=None,
-    # focal_col="focal",
-    # neighbor_col="neighbor",
+    focal_col="focal",  # noqa ARG001
+    neighbor_col="neighbor",  # noqa ARG001
     to_adjlist_kws=dict(drop_islands=None),
 ):
     """
--- a/libpysal/weights/adjtools.py
+++ b/libpysal/weights/adjtools.py
@@ -1,3 +1,5 @@
+# ruff: noqa: B006, C408, N802, N803, N806
+
 import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


libpysal/weights/adjtools.py
```diff
@@ -152,8 +159,8 @@ def adjlist_map(
     funcs=(np.subtract,),
     W=None,
     alist=None,
-    focal_col="focal",
-    neighbor_col="neighbor",
+    # focal_col="focal",
```
jGaboardi, line 162: What do we think here? Unused keywords. Passes without.

libpysal/weights/adjtools.py
```diff
@@ -128,7 +134,8 @@ def _get_W_and_alist(W, alist, to_adjlist_kws, skip_verify=False):
     2. adjacencylist from a W
     3. raise ValueError if neither are provided,
     4. raise AssertionError if both W and adjlist are provided and don't match.
-    If this completes successfully, the W/adjlist will both be returned and are checked for equality.
+    If this completes successfully, the W/adjlist will both be
```
jGaboardi, line 137: Try to comment on Line 130 & 145 --

`N806` specifically for this case of `arg` W then variable W. Wanted to keep refactoring to a min.

libpysal/weights/adjtools.py
martinfleis: Need to keep them. It is public API, this would require deprecation.

libpysal/weights/adjtools.py
martinfleis: We don't import `libpysal.weights.W` here so I'd say let's leave it but it can be a bit confusing.

martinfleis (approved):

PR #654, opened 2023-11-15
lint & format `weights/*.py` part 6

jGaboardi:
This PR:
* formats and lints `weights/*.py` –  part 6 of $n$

### DIFF
--- a/libpysal/weights/weights.py
+++ b/libpysal/weights/weights.py
@@ -856,7 +856,7 @@ class W:
                 list(self.cardinalities.values()),
                 list(range(self.min_neighbors, self.max_neighbors + 2)),
             )
-            self._histogram = list(zip(bin_, ct, strict=False))
+            self._histogram = list(zip(bin_[:-1], ct, strict=True))
             self._cache["histogram"] = self._histogram
         return self._histogram
 
--- a/libpysal/weights/weights.py
+++ b/libpysal/weights/weights.py
@@ -1197,7 +1197,7 @@ class W:
                     wijs = self.weights[i]
                     q[i] = math.sqrt(sum([wij * wij for wij in wijs]))
                     s[i] = [wij / q[i] for wij in wijs]
-                    Q += sum([si for si in s[i]])  # noqa C416
+                    Q += sum(s[i])
                 nQ = self.n / Q
                 for i in self.weights:
                     weights[i] = [w * nQ for w in s[i]]
--- a/libpysal/weights/weights.py
+++ b/libpysal/weights/weights.py
@@ -1,27 +1,31 @@
 """
 Weights.
 """
+
+# ruff: noqa: N802
+
 __author__ = "Sergio J. Rey <srey@asu.edu>"
 
 import copy
-from os.path import basename as BASENAME
 import math
 import warnings
+from collections import defaultdict
+from os.path import basename
+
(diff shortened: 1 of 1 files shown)


libpysal/weights/weights.py
```diff
@@ -849,11 +852,11 @@ def histogram(self):
         value is the number of neighbors for that unit.
         """
         if "histogram" not in self._cache:
-            ct, bin = np.histogram(
+            ct, bin_ = np.histogram(
                 list(self.cardinalities.values()),
                 list(range(self.min_neighbors, self.max_neighbors + 2)),
             )
-            self._histogram = list(zip(bin, ct))
+            self._histogram = list(zip(bin_, ct, strict=False))
```
jGaboardi, line 859: When re-writing this line with `strict=True` the following failures occur:

```
FAILED libpysal/weights/tests/test_weights.py::TestW::test_histogram - ValueError: zip() argument 2 is shorter than argument 1
FAILED libpysal/weights/tests/test_weights.py::Test_WSP_Back_To_W::test_histogram - ValueError: zip() argument 2 is shorter than argument 1
```

libpysal/weights/weights.py
```diff
@@ -1197,15 +1190,14 @@ def set_transform(self, value="B"):
...
-                k = self.cardinalities
                 s = {}
                 Q = 0.0
                 self.weights = self.transformations["O"]
                 for i in self.weights:
                     wijs = self.weights[i]
                     q[i] = math.sqrt(sum([wij * wij for wij in wijs]))
                     s[i] = [wij / q[i] for wij in wijs]
-                    Q += sum([si for si in s[i]])
+                    Q += sum([si for si in s[i]])  # noqa C416
```
jGaboardi, line 1200: [This line](https://github.com/pysal/libpysal/blob/27087efe17010433bf981a199bed29af8402c468/libpysal/graph/tests/test_base.py#L753) in `libpysal/graph/tests/test_base.py` fails when this line is re-written as `ruff` suggests. With:

```python
Q += sum([si for si in s[i]])
```

```
`libpysal/weights/weights.py:1202:30: C416 [*] Unnecessary `list` comprehension (rewrite using `list()`)
```


Re-written to:

```python
Q += sum(list(s))
```

```
FAILED libpysal/graph/tests/test_base.py:753: AssertionError

libpysal/graph/tests/test_base.py::TestBase::test_transform_v - assert <libpysal.gra...t 0x179d0b010> == <libpysal.gra...t 0x179f21790>
```

libpysal/weights/weights.py
martinfleis, line 859:
```suggestion
            self._histogram = list(zip(bin_[:-1], ct, strict=True))
```
It is relying on `zip` omitting the last bin edge. We should do that explicitly.

libpysal/weights/weights.py
martinfleis, line 1200:
```suggestion
                    Q += sum(s[i])
```
It should look like this no? or
```suggestion
                    Q += sum(list(s[I]))
```

need to be tested but both should be equal imho

martinfleis: I was quite proud when I completely re-wrote this using that nice groupby we have in `graph` :)

martinfleis (approved):

PR #655, opened 2023-11-15
lint & format weights/*.py part 7

jGaboardi:
This PR:
* formats and lints `weights/*.py` –  part 7 of $n$

This is a bit more than I had thought it would be. Can split into 2 PRs if desired, @martinfleis


libpysal/weights/raster.py
```diff
@@ -563,7 +565,7 @@ def _index2da(data, index, attrs, coords):
             else:
                 min_data = np.min(data)
                 fill_value = min_data - 1 if min_data < 0 else -1
-                attrs["nodatavals"] = tuple([fill_value])
+                attrs["nodatavals"] = fill_value
```
martinfleis, line 568: This changes the type, doesn't it? Can you test that?

jGaboardi: So after looking into this a bit, there's a bit of strangeness going on with `pytest-cov`. The report [claims those lines are covered](https://github.com/pysal/libpysal/actions/runs/6883267128/job/18723462976?pr=655#step:7:1926), but after some tinkering locally it appears they are actually not. Going to investigate further.

jGaboardi: So (unless I am missing something exceedingly obvious), indeed those lines were not being tested... and they fail with the change... ***BUT*** *they also fail on `main` when actually tested*... This adds another level of anxiety if we can't trust the coverage reports from `pytest`...

```python
from libpysal.weights import raster
pytest.importorskip("xarray")
da2_missing_in = raster.testDataArray((1, 4, 4), missing_vals=True)
w2_missing = raster.da2W(da2_missing_in, "rook", n_jobs=-1)
da2_missing_out = raster.w2da(
    da2_missing_in.data.flatten(), w2_missing, da2_missing_in.attrs, None
)
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 6
      4 da2_missing_in = raster.testDataArray((1, 4, 4), missing_vals=True)
      5 w2_missing = raster.da2W(da2_missing_in, "rook", n_jobs=-1)
----> 6 da2_missing_out = raster.w2da(
      7     da2_missing_in.data.flatten(), w2_missing, da2_missing_in.attrs, None
      8 )

File ~/libpysal/libpysal/weights/raster.py:343, in w2da(data, w, attrs, coords)
    341     raise TypeError("w must be an instance of weights.W")
    342 if hasattr(w, "index"):
--> 343     da = _index2da(data, w.index, attrs, coords)
    344 else:
    345     raise AttributeError(
    346         "This method requires `w` object to include `index` "
    347         "attribute that is built as a `pandas.MultiIndex` object."
    348     )

File ~/libpysal/libpysal/weights/raster.py:570, in _index2da(data, index, attrs, coords)
    568 else:
    569     data_complete = np.empty(shape, data.dtype)
--> 570 data_complete[indexer] = data
    571 coords = {}
    572 for dim, lev in zip(dims, idx.levels, strict=True):

ValueError: shape mismatch: value array of shape (16,) could not be broadcast to indexing result of shape (6,)
```

jGaboardi: So I suppose I'll revert and create an issue for this?

martinfleis: Yes, that sounds the best

martinfleis: just this line and add `noqa,` the rest is fine

jGaboardi: Done

martinfleis (approved):

PR #658, opened 2023-11-17
lint & format `weights/*.py` part 8

jGaboardi:
This PR:
* formats and lints `weights/*.py` –  and should be part 8 of 8
* nearly completes the linting of `libpysal` -> there seem to be some loose ends in `cg` that I will take care of in a follow PR.

### DIFF
--- a/libpysal/weights/tests/test_util.py
+++ b/libpysal/weights/tests/test_util.py
@@ -162,7 +162,7 @@ class Testutil:
         util.higher_order(wsparse, 2)
         util.higher_order(ww, 2)
         ww.transform = "r"
-        wsparse_notbinary = wrook.sparse  # noqa F841
+        _ = wrook.sparse
         util.higher_order(wsparse, 2)
         with pytest.raises(ValueError):
             util.higher_order(ww, 3)
--- a/libpysal/weights/tests/test_util.py
+++ b/libpysal/weights/tests/test_util.py
@@ -4,7 +4,7 @@ import numpy as np
 import pytest
 
 from ... import examples
-from ...io.fileio import FileIO as psopen
+from ...io.fileio import FileIO
 from .. import util
 from ..contiguity import Queen, Rook
 from ..distance import KNN, DistanceBand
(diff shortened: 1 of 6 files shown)


libpysal/weights/tests/test_util.py
```diff
@@ -162,9 +162,9 @@ def test_higher_order_classes(self):
         util.higher_order(wsparse, 2)
         util.higher_order(ww, 2)
         ww.transform = "r"
-        wsparse_notbinary = wrook.sparse
+        wsparse_notbinary = wrook.sparse  # noqa F841
```
jGaboardi, line 165: After `ruff` notified that this line is not being used, I looked a bit deeper and found that here are merely testing things don't fail (except the last one that I move out). Not sure if we want to do the same for the `wsparse_notbinary = wrook.sparse` or just remove it? Maybe outside the scope of this PR?

martinfleis: We can keep it but don't assign to a variable if the result is not tested.

libpysal/weights/tests/test_util.py
```diff
@@ -162,9 +162,9 @@ def test_higher_order_classes(self):
...
+        util.higher_order(wsparse, 2)
```
martinfleis, line 166: Is this correctly moved outside of the pytest.raises?

libpysal/weights/tests/test_util.py
```diff
@@ -200,7 +200,6 @@ def test_full2_w(self):
         ids = ["myID0", "myID1", "myID2", "myID3"]
         w = util.full2W(a, ids=ids)
         np.testing.assert_array_equal(w.full()[0], a)
-        w.full()[0] == a
```
martinfleis, line 203: Why is this removed?

libpysal/weights/tests/test_util.py
jGaboardi: It does not raise. If it was there because it previously raised, it no longer does. But my guess is that it simply was indented by mistake.

libpysal/weights/tests/test_util.py
jGaboardi: The line above it actually does the testing. That line is a "pointless expression" according to `ruff` since it merely calls a comparison but does't actually assert anything.

martinfleis (approved):

PR #660, opened 2023-11-18
correct in-line `ruff` syntax

jGaboardi:
xref #589 

This PR corrects in-line `ruff` syntax in `common.py`.

`# noqa <CODE>` to  `# noqa: <CODE>`


martinfleis:
I just did global find and replace to tackle it across the codebase. It is all the same change, there's no need to split it by file or folder.

martinfleis (approved):

PR #661, opened 2023-11-22
final lint/format run through & update pre-commit

jGaboardi:
This PR:
* cleans up several lingering formatting & linting spots
* updates `.pre-commit` for for `black` and `ruff`
* xref #589

To decide:
* Within `pyproject.toml`, I am more leaning towards the mindset of omitting [`[tool.ruff.ignore]`](https://github.com/pysal/libpysal/blob/9e459608ef1b34e41ef63711da8e86c5ca085f7c/pyproject.toml#L89) in `pyproject.toml` altogether and handling 99.9% of cases within files themselves (everything in the PR passes that way). However, I know others may think this is too strict. So yay or nay for keeping `[tool.ruff.ignore]`?

Once the PR is reviewed and merged the last step in #589 will be to [enable `pre-commit` on triggered GHA](https://results.pre-commit.ci)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,15 @@
-repos: 
-- repo: https://github.com/psf/black
-  rev: "23.10.0"
-  hooks:
-  - id: black
-    language_version: python3
+files: "libpysal\/"
+repos:
+  - repo: https://github.com/psf/black
+    rev: "23.10.1"
+    hooks:
+      - id: black
+        language_version: python3
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: "v0.1.6"
+    hooks:
+      - id: ruff
+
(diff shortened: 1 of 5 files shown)


martinfleis:
> I am more leaning towards the mindset of omitting

If there are no global rules we want to disable then there's no reason to keep that section. But there is some minor compatibility issue with black I think. Some rules go against black, that is why they're disallowed. Otherwise you get to a dance between ruff and black with one making a change and the other reverting it.

jGaboardi:
This PR passes both `black` and `ruff` with no global rules (confirmed locally). But yes, I have experience in the past where `black` and `ruff` seem to countermand each other. Though recently, no.

martinfleis:
Okay. Let's give it a go without.

martinfleis (approved):

PR #662, opened 2023-11-25
Formatting via `ruff`

jGaboardi:
This PR:
* follows the example in [`xarray-contrib/xvec`](https://github.com/xarray-contrib/xvec/pull/54/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R7) for swapping out `black` for `ruff format` in formatting
* updates `pyproject.toml` of this
* makes several minor linting and formatting corrections
* updates contributing guidelines and `docs/install` for supported Python versions 

xref #589

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -3,6 +3,8 @@ review the [contributing procedures](https://github.com/pysal/pysal/wiki/GitHub-
 
 ## Style and format
 
-1. Python 3.6, 3.7, and 3.8 are the officially supported versions.
-2. This project follows the formatting conventions of [`black`](https://black.readthedocs.io/en/stable/) and utilizes [`pre-commit`](https://pre-commit.com) to format commits prior to pull requests being made. 
-    * LJ Miranda provides an [excellent, concise guide](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/) on setting up and implementing a `pre-commit` hook for `black`.
+1. At the time of this writing, Python 3.10, 3.11, and 3.12 are the officially supported versions.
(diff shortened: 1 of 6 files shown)


martinfleis:
Do we want to switch to ruff-format? I think it deserves some discussion. 

I am all for it. When doing that in `xvec` I was just curious but the ruff extension in VS Code is so good that I actually noticed a difference compared to black. And it is one less tool to deal with.

jGaboardi:
> Do we want to switch to ruff-format? I think it deserves some discussion.
> 
> I am all for it. When doing that in `xvec` I was just curious but the ruff extension in VS Code is so good that I actually noticed a difference compared to black. And it is one less tool to deal with.

I am also all for it after realizing how simple the switch was, but also agree that we might want a broader discussion and approval. Shall we open an actual issue for this (and keep this PR open), or have more of a discussion in here?

martinfleis:
I think it is fine to have it here.

martinfleis (approved):
Giving a formal approval to show my support.

knaaptime:
interesting, hadnt seen xvec. We started doing something like that when geosnap started, but it became clear quickly that a long-form geodataframe was far better in our case (since vectors often have different geoms in different time slices...) anyway, another tangent :)

fine with the switch. I use the ruff extension in vscode locally

gegen07 (approved):
Looks good to me! Simplifying the environment is always a good enhancement.

jGaboardi:
> fine with the switch. I use the ruff extension in vscode locally

@knaaptime Can you hit approve to make it more "formal?" 🤵‍♂️

knaaptime (approved):

jGaboardi:
Nows this PR has 3 approvals. @gegen07 @knaaptime @martinfleis Shall we wait another week or so to see if anyone else has opinions? I am perfectly fine taking even longer to make sure we aren't being too hasty.

martinfleis:
There's no rush. I'd wait for the next dev call next week.

jGaboardi:
The dev team in attendance at the 2023-12-7 meeting [voted in favor of this implementation](https://github.com/orgs/pysal/discussions/1278#discussioncomment-7791889). Shall we go ahead and merge now or would it be better to wait longer?

Issue #665, opened 2023-12-08
Interest in Optimal Spatial Matching? 


ljwolf:
For a different causal inference project, I'm writing a few spatial/feature matching algorithms.

One that I have not seen before mentioned anywhere in the literature is like an [optimal spatial matching graph](https://gist.github.com/ljwolf/1529b5cf5bfac2341fd8f103b53f9ad6)? 

Formally, this is a graph where each observation is connected to at least k neighbors in a way that minimises the total edge distance in the graph. So, let distance d_{ij} be the length of a possible edge connecting i to j, and let m_{ij} be a binary variable that is 1 when i is connected to j and 0 otherwise. To build a graph within a single point pattern, we force m_{ij} = m_{ji} so that links are symmetric. But, you can relax this to build a "cross-matching" from one point pattern to another point pattern that minimises the distance of match links.

Letting k be our number of neighbors per observation, the *spatial matching graph* arises from solving the following integer program. 

```
minimize \sum_i^n \sum_j^n  d_{ij}m_{ij} 
subject to
    \sum_j^n m_{ij} >= k \forall i
    m_{ij} \in {0,1} forall ij
```

This gives rise to the following graphs:

![match_knn](https://github.com/pysal/libpysal/assets/2250995/828c4624-cb73-4995-9a25-ce9caf552efe)

The IOU score at the bottom is the intersection over union score with respect to the standard KNN graph. 

In some cases (like odd k) you can't solve this exactly for a dataset with odd numbers. So, depending on how this is specified, you can allow for either:
1. **K-Match**: a single observation with `k+1` neighbors, keeping inequality for the k-neighbors constraint. 
2. **K-Partial Match**: a set of "unmatchable" observations with `k+1` neighbors, one pair each which has half-weight (plotted in light blue above).

ljwolf:
I've looked into it, and I can't express this in a form for `scipy.optimize.milp()`, since the decision variable is not one-dimensional. Thus, this would require a soft dependency on `pulp`.

ljwolf:
I should note, too, that the fact that `k=2` causes a tour in the above data doesn't mean that, in general, `k=2` solves a tour.

martinfleis:
Closed by #666

PR #666, opened 2023-12-08
Matching

ljwolf:
justification in point 5. 

2. [x] This pull request is directed to the `pysal/main` branch.
4. [x] You have [assigned a
5. [x] The justification for this PR is contained in #665 

# todo:

- [x] add `pulp` to our testing matrix. 
- [x] write tests using the toy example in [the gist](https://gist.github.com/ljwolf/1529b5cf5bfac2341fd8f103b53f9ad6)
- [x] write an example notebook


libpysal/graph/_matching.py
```diff
@@ -0,0 +1,164 @@
...
+    return_mip : bool (default: False)
+        whether or not to return the instance of the pulp.LpProblem. By
+        default, the problem is not returned to the user.
+    allow_partial_match : bool (default: False)
+        whether to allow for partial matching. A partial match may have
+        a weight between zero and one, while a "full" match (by default)
+        must have a weight of either zero or one. A partial matching may
+        have a shorter total distance, but will result in a weighted
+        graph.
+    p : int (default: 2)
```
ljwolf, line 62: This isn't being sent to metric at the moment.

libpysal/graph/_matching.py
```diff
@@ -0,0 +1,164 @@
...
+                    match_vars[j, i]
+                    for i in range(n_sources)
+                ]
+            )
+            sense = 1
+        else:
+            summand = pulp.lpSum(
+                [
+                    # in the "upper" triangle, or "lower" triangle
+                    match_vars.get((i, j), match_vars.get((j, i)))
```
ljwolf, line 133: The problem is much more efficient when `x=y`, and we can implement this trick to halve the number of decision variables. If we recursively solved `_spatial_matching(x=x, y=x)`, we'd lose this optimisation. I *think* this optimisation also would work if `x.shape==y.shape`.

ljwolf: No, the only reason why the upper triangle works is because the rows are exchangeable with the columns. If you had a 3 x 5 matching matrix, neither triangle can express the full set of possible matches.

All cases where x != y need the full assignment matrix, and we can only use the upper/lower triangle if y=x.

ljwolf:
Tests are now added, but pulp installation fails on windows through `conda-forge`. Is there a better way to do this?

jGaboardi:
> Tests are now added, but pulp installation fails on windows through `conda-forge`. Is there a better way to do this?

In `spopt` we install `pulp` via `pip` --> https://github.com/pysal/spopt/blob/4760bd9e03f9b31c003bfa7f5d785a2a81b174bd/ci/312-latest.yaml#L32

jGaboardi:
Looks like `numpy` is never being imported --> https://github.com/pysal/libpysal/actions/runs/7159612453/job/19493044782?pr=666#step:7:2098

jGaboardi:
Is [this](https://github.com/pysal/libpysal/actions/runs/7159753204/job/19493357426?pr=666#step:7:1796) passing locally for you?

ljwolf:
Yes, but fixing that, I still have fails locally. Will fix.

ljwolf:
One question I had for maintainers: 

I built the spatial matching framework to be able to do cross-pattern queries. So, something like: 

> For all points in pattern X, find at least k matched points in pattern Y such that the total distance of X->Y matches are minimised. 

I think this would be good to expose in `cg`, since it's a general purpose spatial matching function for computational geometry applications. I have something even more general that will probably make it into a separate package for spatial causal inference in python, but the spatial implementation can live here. 

So, would it be OK to grab `_spatial_matching()` from `libpysal.graph._matching` and expose it directly in `libpysal.cg`? I could alternatively define a dispatching function that *requires* the second pattern (`y` argument) to be provided.

libpysal/graph/_matching.py
```diff
@@ -0,0 +1,170 @@
...
+                ]
+            )
+            sense = int(not allow_partial_match)
+        mp += pulp.LpConstraint(summand, sense=sense, rhs=n_matches)
+    if match_between:  # but, we may choose to ignore some sources
+        for i in range(n_sources):
+            summand = pulp.lpSum([match_vars[j, i] for j in range(n_targets)])
+            mp += pulp.LpConstraint(summand, sense=-1, rhs=n_matches)
+    if solver is None:
+        solver = pulp.COIN(msg=False)
```
jGaboardi, line 151: I'm wondering if there is way to simply choose the default installed solver without needing to specific (or is this what's happening here with `pulp.COIN()`?). Implementing that logic would go a long way over in `spopt` >> https://github.com/pysal/spopt/issues/348

cc @gegen07

ljwolf: I think we need to do this, since `pulp.PULP_CBC_CMD()` I think is different from `pulp.COIN_CMD()`, and some students could run `pulp.GUROBI()` but not `pulp.GUROBI_CMD()`. 

I think we should really leave it to pulp to deal with this, rather than specifying COIN.

jGaboardi: Yeah, seems like it would make a nice improvement in `spopt` to have the ability to pass in no solver, and then the default one installed is used [within the `solve()` methods](https://github.com/pysal/spopt/blob/ec36c133fbef6f1cde1b3974d34e69b703eed03a/spopt/locate/p_dispersion.py#L315).

ljwolf:
The code is ready to review. I'm not sure why the ruff check is failing... I've run the ruff auto formatter myself locally on the submission (2c04acf, 215022d, 273e6ea)

jGaboardi:
> The code is ready to review. I'm not sure why the ruff check is failing... I've run the ruff auto formatter myself locally on the submission ([2c04acf](https://github.com/pysal/libpysal/commit/2c04acf749ef12c10536e164af95437fbc7e61d4), [215022d](https://github.com/pysal/libpysal/commit/215022db802949c4f2dcadca53d858f2bd594529), [273e6ea](https://github.com/pysal/libpysal/commit/273e6ea27000b8540da604272fb75e34de4cdc7f))

Do you have the most recent `ruff` installed locally? I think there was a fresh version released last night. Could be playing a part.

ljwolf:
Yes, not quite sure what's up here: 

Screenshot 2023-12-18 at 11 42 40

0.1.8 is the most recent version of ruff, and it fixes nothing both in the CI and locally?

martinfleis:
It is `ruff-format` that is failing, not `ruff` linter.

ljwolf:
OK! Wasn't clear on the differences between ruff check and format. Sorry! 

I think this is complete.

libpysal/graph/_matching.py
```diff
@@ -0,0 +1,168 @@
...
+        for i in range(n_sources):
+            summand = pulp.lpSum([match_vars[j, i] for j in range(n_targets)])
+            mp += pulp.LpConstraint(summand, sense=-1, rhs=n_matches)
+
+    status = mp.solve(solver)
+
+    if (status != 1) & (not allow_partial_match):
+        warnings.warn(
+            f"Problem is {pulp.LpStatus[status]}, so edge weights may be non-integer!",
+            stacklevel=1,
```
jGaboardi, line 155: This should probably be `stacklevel=2`, if I'm not mistaken.

libpysal/graph/tests/test_matching.py
```diff
@@ -0,0 +1,110 @@
...
+    index_parts=False
+)
+stores_unique = stores.drop_duplicates(subset="geometry")
+
+np.random.seed(85711)
+simple = np.random.random(size=(5, 2))
+
+
+def test_correctness_k1():
+    pulp = pytest.importorskip("pulp")
```
jGaboardi, line 29: Let's move this above and skip all tests in the file if `pulp` isn't installed.

libpysal/graph/tests/test_matching.py
```diff
@@ -0,0 +1,110 @@
...
+    pytest.importorskip("sklearn")
+    default_solver = pulp.listSolvers(onlyAvailable=True)
+    if len(default_solver) == 0:
```
jGaboardi, line 32: Same for fetching and defining the default solver. This only needs to happen once.

libpysal/graph/tests/test_matching.py
jGaboardi: And same with `sklearn`

jGaboardi (changes requested):
Looking very cool.

ci/310-oldest.yaml
```diff
@@ -29,6 +29,7 @@ dependencies:
   - sqlalchemy=2.0
   - zstd
   - xarray=2022.3
-  - pip
+  - pulp
```
martinfleis, line 32:
```suggestion
  - pip
```

ci/312-no-optional.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   # testing
   - codecov
   - matplotlib
+  - scikit-learn
```
martinfleis, line 17:
```suggestion
```
No optional dependencies shall be in this env

environment.yml
```diff
@@ -26,3 +26,7 @@ dependencies:
   - sqlalchemy
   - xarray
   - zstd
+  - coin-or-cbc
```
martinfleis, line 29: Why do we have this here and not in CI envs?

libpysal/graph/_matching.py
```diff
@@ -0,0 +1,168 @@
+import warnings
+
+import numpy
+from sklearn.metrics import pairwise_distances
```
martinfleis, line 4: So far, sklearn has been a soft dependency for libpysal. Do we want to change that?

martinfleis (commented):
The large diff in base is just moving stuff around to get alphabetic order of builders?

Can you also add a test for `Graph.build_spatial_matches` ?

ci/312-no-optional.yaml
jGaboardi: Good catch @martinfleis

jGaboardi:
> The large diff in base is just moving stuff around to get alphabetic order of builders?
> 
> Can you also add a test for `Graph.build_spatial_matches` ?

Yes.

libpysal/graph/_matching.py
knaaptime: yes, we decided to change

ci/312-no-optional.yaml
knaaptime: made it a hard dep

ljwolf: So, I will keep `scikit-learn` in this env?

environment.yml
ljwolf: This was part of my attempt to get pulp installed correctly. Fine to remove.

ci/312-no-optional.yaml
martinfleis: I suppose so but it needs to make its way to requirements in pyproject.toml if it is becoming hard dep now.

ljwolf:
I believe I've addressed all comments!

knaaptime (approved):

jGaboardi (approved):

ljwolf:
Hi! could one of the approvers please merge this?

ljwolf:
thank you!

PR #670, opened 2023-12-22
ENH: include assign_self_weight (formerly fill_diagonal) in Graph

### LINKED ISSUE #668: fill diagonal in graph
is still not yet implemented. The laziest way i can do this now is 

```python
g = Graph.build_kernel(gdf, bandwidth=1000)
sp = g.sparse
sp = sp.tolil()
sp.setdiag(1)
g = Graph.from_sparse(sp, ids=gdf.index)
```

which is what we do currently for W. Not sure if there's something more performant than roundtripping through sparse, given the new format? maybe some pandas-fu or something?

martinfleis:
Closes #668

@knaaptime this includes pandas-fu version as well as roundtrip via sparse for now. Timing below. 

```py
import geopandas as gpd
from geodatasets import get_path
from libpysal import graph

df = gpd.read_file(get_path("geoda.ncovr"))
contig  = graph.Graph.build_contiguity(df)

%%timeit
contig.fill_diagonal()
5.31 ms ± 108 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
contig.fill_diagonal_sparse()
16.3 ms ± 92.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Here, the sparse version is about 3x slower. This ratio seems to be more or less consistent with any size of the graph.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1429,10 +1429,11 @@ class Graph(SetOpsMixin):
         Parameters
         ----------
         weight : float | array-like
-            Defines the value(s) to which the weight of self-loops should be set. If a
-            constant is passed then each self-loop will get this value (default is 1).
-            An array of length ``Graph.n`` can be passed to set explicit values to each
-            self-loop (assumed to be in the same order as original data).
+            Defines the value(s) to which the weight representing the relationship with
+            itself should be set. If a constant is passed then each self-weight will get
+            this value (default is 1). An array of length ``Graph.n`` can be passed to
+            set explicit values to each self-weight (assumed to be in the same order as
+            original data).
 
         Returns
         -------
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -162,7 +162,7 @@ class Graph(SetOpsMixin):
         pandas.Series
             Underlying adjacency list
         """
-        return self._adjacency.copy()
+        return self._adjacency.copy(deep=True)
 
     @classmethod
     def from_W(cls, w):  # noqa: N802
@@ -1418,6 +1418,25 @@ class Graph(SetOpsMixin):
         # substract isolates from mask of zeros
         zeros = (self._adjacency == 0) != isolates
         return Graph(self._adjacency[~zeros], is_sorted=True)
+
+    def fill_diagonal(self):
+        no_isolates = self.unique_ids.difference(self.isolates)
+        addition = pd.Series(
+            1,
+            index=pd.MultiIndex.from_arrays(
+                [no_isolates, no_isolates], names=["focal", "neighbor"]
+            ),
+            name="weight",
+        )
+        adj = pd.concat([self._adjacency, addition])
+        adj.loc[self.isolates] = 1
+        return Graph(adj, is_sorted=False)
+
+    def fill_diagonal_sparse(self):
+        sp = self.sparse
+        sp = sp.tolil()
+        sp.setdiag(1)
+        return Graph.from_sparse(sp, ids=self.unique_ids)
 
 
 def _arrange_arrays(heads, tails, weights, ids=None):
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1421,26 +1421,26 @@ class Graph(SetOpsMixin):
         zeros = (self._adjacency == 0) != isolates
         return Graph(self._adjacency[~zeros], is_sorted=True)
 
-    def set_self_weights(self, val=1):
-        """Set self-weights in the Graph to a value.
+    def add_self_loops(self, weight=1):
+        """Set add self-loopswith a set weight to the Graph.
 
-        Value for each ``focal == neighbor`` location in the graph is set to ``val``.
+        Value for each ``focal == neighbor`` location in the graph is set to ``weight``.
 
         Parameters
         ----------
-        val : float | array-like
-            Defines the value(s) to which the self-weights should be set. If a constant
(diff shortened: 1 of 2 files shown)


libpysal/graph/base.py
```diff
@@ -1419,6 +1421,45 @@ def eliminate_zeros(self):
...
+
+        Value for each ``focal == neighbor`` location in the graph is set to ``val``.
+
+        Parameters
+        ----------
+        val : float | array-like
+            Defines the value(s) to which the weights matrix diagonal should
+            be set. If a constant is passed then each element along the
+            diagonal will get this value (default is 1). An array of length
+            Graph.n can be passed to set explicit values to each element along
```
jGaboardi, line 1435:
```suggestion
            ``Graph.n`` can be passed to set explicit values to each element along
```

jGaboardi (commented):
Looking good.

jGaboardi (approved):

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,45 @@ def eliminate_zeros(self):
...
+            name="weight",
+        )
+        adj = (
+            pd.concat([self.adjacency.drop(self.isolates), addition])
+            .reindex(self.unique_ids, level=0)
+            .reindex(self.unique_ids, level=1)
+        )
+        return Graph(adj, is_sorted=True)
+
+    def fill_diagonal_sparse(self):
```
knaaptime, line 1457: nice! not sure if we are keeping both? but this needs to accept a `val`

martinfleis: I kept it here for an easy comparison. I will remove it :). Can't think of a reason to keep both.

knaaptime: thats what i figured 🚀

sjsrey:
https://github.com/pysal/libpysal/issues/668#issuecomment-1868027857

jGaboardi (commented):
Will the terminology in this PR be affected by the discussion in #668?

martinfleis:
> Will the terminology in this PR be affected by the discussion in #668?

Yes. I'll update this PR as soon we have an agreement.

martinfleis:
Updated as per #668. Check the docstring please if it makes sense.

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,38 @@ def eliminate_zeros(self):
         zeros = (self._adjacency == 0) != isolates
         return Graph(self._adjacency[~zeros], is_sorted=True)
 
+    def assign_self_weight(self, weight=1):
+        """Assign values to edges represeting self-weight.
```
jGaboardi, line 1425:
```suggestion
        """Assign values to edges representing self-weight.
```

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,38 @@ def eliminate_zeros(self):
...
+
+        Value for each ``focal == neighbor`` location in the graph is set to ``weight``.
```
jGaboardi, line 1427:
```suggestion
        The value for each ``focal == neighbor`` location in the graph is set to ``weight``.
```

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,38 @@ def eliminate_zeros(self):
...
+
+        Parameters
+        ----------
+        weight : float | array-like
+            Defines the value(s) to which the weight of self-loops should be set. If a
```
jGaboardi, line 1432: Are we sticking with "self-loop" here or going with "self-weight?"

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,38 @@ def eliminate_zeros(self):
...
+            constant is passed then each self-loop will get this value (default is 1).
```
jGaboardi, line 1433: Same here.

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,38 @@ def eliminate_zeros(self):
...
+            An array of length ``Graph.n`` can be passed to set explicit values to each
+            self-lopp (assumed to be in the same order as original data).
```
jGaboardi, line 1435:
```suggestion
            self-loop (assumed to be in the same order as original data).
```

Same here.

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,38 @@ def eliminate_zeros(self):
...
+
+        Returns
+        -------
+        Graph
+            A new Graph with added self-weights
```
jGaboardi, line 1440:
```suggestion
            A new ``Graph`` with added self-weights.
```

knaaptime (approved):

libpysal/graph/base.py
```diff
@@ -1419,6 +1421,39 @@ def eliminate_zeros(self):
         zeros = (self._adjacency == 0) != isolates
         return Graph(self._adjacency[~zeros], is_sorted=True)
 
+    def assign_self_weight(self, weight=1):
+        """Assign values to edges representing self-weight.
+
+        The value for each ``focal == neighbor`` location in the graph is set to ``weight``.
```
jGaboardi, line 1427: Looks like `pre-commit` is complaining that this line is too long.

```suggestion
        The value for each ``focal == neighbor`` location in
        the graph is set to ``weight``.
```

jGaboardi (approved):

jGaboardi:
Shall we update the title of the issue to reflect changes in terminology?

sjsrey (approved):

PR #676, opened 2024-01-15
ENH: add Graph.apply, Graph.aggregate and allow callable as transformation in transform

### LINKED ISSUE #675: Interest in Graph.apply?
Within `momepy`, I often need to apply a custom function over an array of values or geometries of neighbours to any focal. I think that a function like `apply` shown below would go a long way in there. Shall we add it to Graph in this form? Or using different name?

```py
def apply(graph, y, func, **kwargs):
    grouper = y.take(
        graph.adjacency.index.codes[1]
        ).groupby(graph.adjacency.index.codes[0])
    result = grouper.agg(func, **kwargs)
    result.index = graph.unique_ids
    return result
```

This allows customised lags like getting median or mode or if ´y´is a GeoSeries apply a geo `func`.

```py
# median
apply(queen, df.area, np.median)

# IQR
apply(queen, df.area, scipy.stats.iqr)

# area of convex hull around neighbors
apply(queen, df.geometry, lambda x: x.unary_union.area)
```

_I love how simple this is on top of Graph_

martinfleis:
Closes #675

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1455,6 +1455,35 @@ class Graph(SetOpsMixin):
         )
         return Graph(adj, is_sorted=True)
 
+    def apply(self, y, func, **kwargs):
+        """Apply a reduction across the neighbor sets
+
+        Applies ``func`` over groups of ``y`` defined by neighbors for each focal.
+
+        Parameters
+        ----------
+        y : array_like
+            array of values to be grouped. Can be 1-D or 2-D and will be coerced to a
+            pandas object
+        func : function, str, list, dict or None
+            Function to use for aggregating the data passed to pandas ``GroupBy.apply``.
+
+        Returns
+        -------
+        Series | DataFrame
+            pandas object indexed by unique_ids
+        """
+        if not isinstance(y, pd.Series | pd.DataFrame):
+            y = pd.DataFrame(y) if hasattr(y, "ndim") and y.ndim == 2 else pd.Series(y)
+        grouper = y.take(self._adjacency.index.codes[1]).groupby(
+            self._adjacency.index.codes[0]
+        )
+        result = grouper.apply(func, **kwargs)
+        result.index = self.unique_ids
+        if isinstance(result, pd.Series):
+            result.name = None
+        return result
+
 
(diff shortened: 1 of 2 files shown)


ljwolf:
great implementation! two small thoughts
1. I think that it might be good to support applying `func` over the weights. So, something like `graph.apply('median')` gives you the median weight in each neighbor set. I don't mind if this looks like `graph.apply(None, 'median', **kwargs)`... 
2.  ...but this probably instead should look like `graph.apply('median', values=None, **kwargs)`, which matches the `func`-first way that `grouper.apply()` and `functools.reduce()` is written?

martinfleis:
So apply `func` over weights if `values` are not set and otherwise ignore `weights` and apply `func` over `values`? I am on the verge whether it can be potentially confusing or not, meaning if it should not be two methods rather than one.

jGaboardi:
> I am on the verge whether it can be potentially confusing or not, meaning if it should not be two methods rather than one.

I agree on the "potentially confusing" part and think 2 distinct methods might be a good idea.

knaaptime:
nice. two quick reactions

1. i think the really nice thing about the graph, especially compared to the old implementation, is how easy it is to do these kind of manipulations manually because of the way the pandas structure works underneath. An apply func is nice, but i think it would actually be more useful to demo these kinds of things in a tutorial so folks know how to actually manipulate a graph at will. You couldnt do that kinda thing with a W

2. couldnt this just be `g._adjacency.groupby(level=0).transform(func)`? like [i initially sketched](https://github.com/pysal/libpysal/pull/513#issuecomment-1407586099) for the lags?

martinfleis:
Actually not because `y` does not have the MultiIndex assigned.

ljwolf:
Fine with that!

martinfleis:
Added `aggregate` and allowed callable as transformation in `transform`.

martinfleis:
It will need some examples but that applies to the rest of Graph as well :).

ljwolf (approved):
looks good to me!

knaaptime (approved):
sweet

jGaboardi (approved):

Issue #681, opened 2024-01-29
Reverse dependency issues


martinfleis:
Tracking currently failing checks in reverse dependency testing.

- [ ] geosnap (https://github.com/reiinakano/scikit-plot/issues/112)
- [ ] mgwr - test files are not shipped with `mgwr` (which is a good thing). Need to find a way of skipping those, probably via a custom pytest.mark within mgwr. To be discussed.
- [ ] pointpats - ModuleNotFoundError: OpenCV2 is required to use this function. Please install it with `pip install opencv-contrib-python` - it should be installed, not sure why it is not picked up. In any case, this test should likely be conditionally skipped in pointpats.
- [ ] segregation - some issue with `quilt3` loading the data from remote. might be a multithreading issue?
- [ ] spvcm - actual numpy and statsmodels errors. need to report or discuss the maintenance status of `spvcm`
- [ ] tobler - `dask_geopandas` missing (easy), same quilt3 issue as with segregation, and one numeric difference [here](https://github.com/pysal/libpysal/actions/runs/7689480168/job/20951993268#step:4:9050)

jGaboardi:
Excellent job @martinfleis for the push on this.

PR #682, opened 2024-01-29
CI: partially resolve reverse dep testing issues

martinfleis:
Remove parallel execution to resolve those `quilt3` issues, add some missing dependencies.

Couldn't make opencv work but we should resolve that in pointpats by a) skipping the test then cv2 cannot be imported, b) refactoring the minimum_bounding_rectangle on top of shapely instead.

See the log of the latest run with this config [here](https://github.com/martinfleis/libpysal/actions/runs/7703159298)

mgwr, pointpats, spvcm, tobler remain failing, all of which need to be solved in upstream.

xref #681

### DIFF
--- a/.github/workflows/reverse.yml
+++ b/.github/workflows/reverse.yml
@@ -49,6 +49,8 @@ jobs:
               pulp
               dask-geopandas
               kdepy
+              matplotlib
+              statsmodels
             installation_command: >-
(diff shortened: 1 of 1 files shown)


knaaptime:
cool! tobler's having an issue with h3 at the moment im not sure how to resolve. The update started when they updated the c library recently, so its probably best to just update the tests--but windows is still passing

jGaboardi (approved):

PR #691, opened 2024-03-08
REF: refactor Graph.to_W to avoid perf bottleneck

### LINKED ISSUE #672: Graph.to_W is slow
For a 150k buildings, it is way faster to create Queen contiguity using old constructor than converting the Graph based on queen to W. We should avoid double groupby and maybe look at other improvements there.

martinfleis:
Closes #672 

For adjacency of 1,404,080 edges, the conversion time is 42.7s on main and 4.5s in this PR.

@jGaboardi not sure how to properly label this...

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -189,34 +189,18 @@ class Graph(SetOpsMixin):
         libpysal.weights.W
             representation of graph as a weights.W object
         """
-        ids, labels = pd.factorize(
-            self._adjacency.index.get_level_values("focal"), sort=False
-        )
-        neighbors = (
-            self._adjacency.reset_index(level=1)
-            .groupby(ids)
-            .apply(
-                lambda group: list(
-                    group[
-                        ~((group.index == group.neighbor) & (group.weight == 0))
-                    ].neighbor
-                )
-            )
-        )
-        neighbors.index = labels[neighbors.index]
-        weights = (
-            self._adjacency.reset_index(level=1)
-            .groupby(ids)
-            .apply(
-                lambda group: list(
-                    group[
-                        ~((group.index == group.neighbor) & (group.weight == 0))
-                    ].weight
-                )
-            )
-        )
-        weights.index = labels[weights.index]
-        return W(neighbors.to_dict(), weights.to_dict(), id_order=labels.tolist())
+        grouper = self._adjacency.groupby(level=0)
+        neighbors = {}
+        weights = {}
+        for ix, chunk in grouper:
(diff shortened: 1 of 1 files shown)


jGaboardi:
I'd say this is an "enhancement" due to the performance improvement.

martinfleis:
Thinking about this more, I think that we shall use the same logic in `weights` and `neighbors` properties that are unbearably slow for large graphs. Hard to catch these things when developing on small graphs...

knaaptime:
i got a question about the performance of the weights builders in the workshop yesterday, and i had to proudly scoff that they're the fastest you can find... 'Martin uses these routinely on datasets in the millions' :D

knaaptime (approved):
looks nice. The idea is the groupby already gives you all the info you need, so just bypass the apply and build the lists cheaply? 
Agreed, lets adopt this same style elsewhere

jGaboardi:
Do we know what's causing the failures on [macOS](https://github.com/pysal/libpysal/actions/runs/8204865013/job/22442404100?pr=691) & [ubuntu-dev](https://github.com/pysal/libpysal/actions/runs/8204865013/job/22442403518?pr=691)?

martinfleis:
Builders are fast but some compatibility bits seem to be worse :). I'm noticing it only now when I regularly use Graph on large data. There's a reason it is still experimental :)

ljwolf:
This looks fine to me! I think we may want to consider some benchmarks with [asv](https://docs.scipy.org/doc/scipy/dev/contributor/benchmarking.html) for these kinds of things... I think that construction, serialization, conversion, lag, and standardization are the big targets?

ljwolf (approved):

jGaboardi (approved):

jGaboardi:
Looks like [all failures](https://github.com/pysal/libpysal/actions/runs/8204865013/job/22448677505?pr=691#step:7:14795) will be fixed by #692. Going ahead with that merge now and see if we can get this green.

jGaboardi:
All green (for now --> the [macOS failure just before](https://github.com/pysal/libpysal/actions/runs/8211557855/job/22460651714#step:6:43) had to do with a connection issue for `geodatasets`).

Not sure if we wanted to see about implementing @ljwolf's idea for `asv` benchmarking here, or that's something from the future. Thinking probably it's future, but I'll wait on merging until confirmed.

martinfleis:
Not going to do `asv` here. I'm generally a bit skeptic about it given we have it in geopandas and no one is running it or anything.

PR #693, opened 2024-03-25
Work around GEOS issue in voronoi_frames

martinfleis:
While using the new `voronoi_frames` on a fairly large dataset, I noticed this issue libgeos/geos#1062 plus another where some polygons coming from GEOS were not actually valid, likely due to the same reason. This is a workaround that solves all I faced.

### DIFF
--- a/libpysal/cg/voronoi.py
+++ b/libpysal/cg/voronoi.py
@@ -425,7 +425,7 @@ def voronoi_frames(
         shapely.GeometryCollection(objects.values), extend_to=limit
     )
     # Get individual polygons out of the collection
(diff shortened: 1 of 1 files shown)


jGaboardi (approved):

jGaboardi:
Is this something that might be fixed in the future upstream? If so, should we make a ticket here to check the behavior in the future? Or will the workaround continue to be functional after the upstream issue is resolved?

martinfleis:
I suppose it will be fixed in upstream. The if clause will be just skipped in that case, like it is in most cases where the issue does not happen right now. This is a very corner case but it was causing an issue when running enclosed_tessellation in momepy.

PR #694, opened 2024-03-26
ENH: add Graph.build_h3

### LINKED ISSUE #667: Build Graph from H3
I have an implementation of building a (higher-order) contiguity Graph based on H3 index. I'd like to expose it as a builder on `Graph, so dumping the code below before I'll forget.

 ```py
import h3

from libpysal import graph

data = ... # dataframe with H3 IDs as an index, like those coming from tobler h3fy
order = 5

neighbors = {}
weights = {}
for ix in data.index:
    rings = h3.hex_range_distances(ix, order)
    for i, ring in enumerate(rings):
        if i == 0:
            neighbors[ix] = []
            weights[ix] = []
        else:
            neighbors[ix].extend(list(ring))
            weights[ix].extend([i] * len(ring))

matrix = graph.Graph.from_dicts(neighbors, weights)
```

It encodes the order as weight now, I was casting to a binary graph and used topological inverse weighting based on that (`inverse = graph.Graph(1 / matrix.adjacency)`). We can figure out what the API should be once the PR is alive?

I also think that the latest beta release of h3 completely changed the API so the h3 calls will need to change a bit but it's been beta for a year. 

based on https://github.com/Urban-Analytics-Technology-Platform/demoland-project/blob/main/code/04_generalisation/create_spatial_weights.ipynb

### PR DESCRIPTION
**Pull request by martinfleis**
Closes #667 

I am open to API suggestions, especially regarding `weight`. Example usage:

```py
import geopandas as gpd
from geodatasets import get_path
from tobler.util import h3fy

from libpysal import graph

gdf = gpd.read_file(get_path("geoda guerry"))
h3 = h3fy(gdf, resolution=4)
G = graph.Graph.build_h3(h3.index)
```

In theory, you can use this to do spatial stats without having any geometry, just the indices.

@ljwolf I am aware of you suggestion to support other hierarchical indices as well but I am not sure if the overall goal with that would be to have a single builder method (or of it is wise to have only one) or individual, so I went ahead with H3-specific one.


knaaptime:
would it be better to move h3fy to lib?

martinfleis:
> would it be better to move h3fy to lib?

I got no opinion there, either way is fine with me.

jGaboardi (approved):

PR #704, opened 2024-04-10
Added examples to from_dicts and build_knn.

Kryndlea:
Added examples to `from_dicts` and `build_knn`.


jGaboardi (approved):

libpysal/graph/base.py
```diff
@@ -336,6 +336,59 @@ def from_dicts(cls, neighbors, weights=None):
...
+        on the length of the shared border):
+
+        >>> neighbors = {
+        ...     'Africa': ['Asia'],
+        ...     'Asia': ['Africa', 'Europe'],
+        ...     'Australia': [],
+        ...     'Europe': ['Asia'],
+        ...     'North America': ['South America'],
+        ...     'South America': ['North America'],
+        ... }
```
martinfleis, line 372: Given it is already defined above, there's probably no need to do that again.
```suggestion
```

libpysal/graph/base.py
```diff
@@ -685,6 +738,53 @@ def build_knn(cls, data, k, metric="euclidean", p=2, coincident="raise"):
...
+
+        Examples
+        --------
+        >>> import geopandas as gpd
+        >>> from geodatasets import get_path
+        >>> nybb = gpd.read_file(get_path('nybb')).set_index('BoroName')
+        >>> nybb
+                       BoroCode  ...                                           geometry
+        BoroName                 ...
+
```
martinfleis, line 750:
```suggestion
```

I don't think this line is supposed to be there.

libpysal/graph/base.py
```diff
@@ -685,6 +738,53 @@ def build_knn(cls, data, k, metric="euclidean", p=2, coincident="raise"):
...
+        Staten Island         5  ...  MULTIPOLYGON (((970217.022 145643.332, 970227....
+        Queens                4  ...  MULTIPOLYGON (((1029606.077 156073.814, 102957...
+        Brooklyn              3  ...  MULTIPOLYGON (((1021176.479 151374.797, 102100...
+        Manhattan             1  ...  MULTIPOLYGON (((981219.056 188655.316, 980940....
+        Bronx                 2  ...  MULTIPOLYGON (((1012821.806 229228.265, 101278...
+
+        >>> knn1 = graph.Graph.build_knn(nybb.centroid, k=1)
```
martinfleis, line 757: I would maybe switch the order of these two examples. Start with `k=3` and then say something like "Specifying k=1 identifies the nearest neighbor (note that this can be asymmetrical)".

martinfleis (commented):
Few notes.

PR #711, opened 2024-05-27
ENH: implement Graph.__repr__

martinfleis:
I quite like when the `__repr__` actually says something about the object, so here's a draft of `Graph.__repr__`. Happy to iterate over if you have ideas what and how should be shown. The info I picked now is very cheap to compute.

```py
In [1]: import numpy as np
   ...: from libpysal import graph
   ...: import geopandas as gpd
   ...: from geodatasets import get_path

In [2]: coords = np.random.random((100_000, 2))

In [5]: delaunay = graph.Graph.build_triangulation(coords)

In [6]: %time r = delaunay.__repr__()
CPU times: user 27 ms, sys: 10.9 ms, total: 37.9 ms
Wall time: 36.9 ms

In [7]: delaunay
Out[7]: 
<Graph of 100000 nodes and 599936 nonzero edges indexed by
 [0, 1, 2, 3, 4, ...]>

In [8]: nybb = gpd.read_file(get_path('nybb')).set_index('BoroName')

In [9]: contig = graph.Graph.build_contiguity(nybb)
   ...: contig
Out[9]: 
<Graph of 5 nodes and 10 nonzero edges indexed by
 ['Staten Island', 'Queens', 'Brooklyn', 'Manhattan', 'Bronx']>
```

_udpated to match the latest formatting_


libpysal/graph/base.py
jGaboardi, line 1: I'm +1 for this. Similar info as to that in [`networkx.Graph.__str__`](https://github.com/networkx/networkx/blob/729bec23b10e89b5f6d37175e30484f2278107c4/networkx/classes/graph.py#L413). Should we also add a similar `__str__` method here for our `Graph` object?

martinfleis: No need. In the absence of `__str__`, Python falls back to `__repr__`.

```py
In [3]: print(str(delaunay))
<Graph with 100000 nodes and 599920 nonzero edges>
 Indexed by: [0, 1, 2, 3, 4, ...]
```

libpysal/graph/base.py
```diff
@@ -149,6 +149,16 @@ def __getitem__(self, item):
             )
         return self._adjacency.loc[item]
 
+    def __repr__(self):
+        if len(self.unique_ids) > 5:
+            unique_ids = f"{str(self.unique_ids[:5].tolist())[:-1]}, ...]"
+        else:
+            unique_ids = self.unique_ids.tolist()
+        return (
+            f"<Graph with {self.n} nodes and {self.nonzero} nonzero edges>\n "
```
jGaboardi, line 158: Is there a particular reason for the angle brackets only on this first line? Perhaps either also adding for the second line?

```
<Graph with ....>
<Indexed by: [...]>
```

Or one set containing all info:

```
<Graph with ....
 Indexed by: [...]>
```

martinfleis: I started off from scipy sparse, which does the latter of your examples but didn't like how it looks :D. I am not sure if there's any convention about this. Happy to go with whatever we agree on, don't have a strong preference.

jGaboardi: Maybe replace the angle brackets with emojis? Kidding of course.

Another option might be:

```
<
 info line 1 ...
 info line 2 ...
>
```

jGaboardi: And as far as the precise terminology we can also consider:
* "Graph which has $x$ nodes and ... "
* "Graph of $x$ nodes and ... "

martinfleis: > [...] a string of the form <...some useful description...> should be returned 

From [docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__).

So 
```
<Graph with ....
 Indexed by: [...]>
```
is probably the best option.

martinfleis: What about

```
<Graph of 5 nodes and 10 nonzero edges indexed by
    ['Staten Island', 'Queens', 'Brooklyn', 'Manhattan', 'Bronx']>
```

```
<Graph of 1000 nodes and 5956 nonzero edges indexed by
    [0, 1, 2, 3, 4, ...]>
```

It is somewhat similar to 
```
<5x5 sparse array of type '<class 'numpy.float64'>'
	with 11 stored elements in Compressed Sparse Row format>
```

jGaboardi: I'm not against, but I prefer the flushness of:

```
<stuff...
 things...>
```

jGaboardi (approved):

PR #714, opened 2024-05-29
Refactor handling of coincident points in triangulation

### LINKED ISSUE #713: Co-location issues with Graph triangulation builders
I have encountered an issue with triangulation where the precision used by shapely does not match the precision used by QHull, leading to inability of our tooling to filter out points seen as equal. Repr:

```py
s = gpd.GeoSeries.from_wkt(
    ['POINT (4071747.2387195 3117751.359062438)',
 'POINT (4071747.2396468357 3117751.359621293)',
 'POINT (4071799.634601838 3117828.4939858434)',
 'POINT (4071770.345504768 3117986.514968087)',
 'POINT (4071827.410512167 3117723.096411029)']
)
Graph.build_triangulation(s)
```

The first two points are not seen as duplicated by shapely but are seen as duplicated by QHull.

I think that the best way around this is to refactor our handling of co-located points and rely on the output of `Delaunay.coplanar` instead of trying to filter this before triangulation. At the same time, we can let QHull to deal with jittering as well, by passing the `QJ` option to QHull, to simplify our codebase. 

We would then only need custom treatment in Voronoi, but only for jittering as shapely returns duplicated faces for duplicated points, so a clique is created automatically.

martinfleis:
Closes #713 

Had to touch more stuff than intended but we now rely on QHull to identify coincident points and provide a lookup from which we can build cliques. This should ensure that we won't hit the precision issues. 

Some minor cleanup happened alongside, i.e. to avoid solving isolates twice in a row.

### DIFF
--- a/libpysal/graph/_kernel.py
+++ b/libpysal/graph/_kernel.py
@@ -3,6 +3,7 @@ import pandas
 from scipy import optimize, sparse, spatial, stats
 
 from ._utils import (
+    CoplanarError,
     _build_coplanarity_lookup,
     _induce_cliques,
     _jitter_geoms,
@@ -239,7 +240,7 @@ def _knn(coordinates, metric="euclidean", k=1, p=2, coplanar="raise"):
         coplanar_lut = _build_coplanarity_lookup(geoms)
         max_at_one_site = coplanar_lut.input_index.str.len().max()
         if coplanar == "raise":
-            raise ValueError(
+            raise CoplanarError(
                 f"There are {len(coplanar_lut)} unique locations in the dataset, "
                 f"but {len(coordinates)} observations. At least one of these sites "
                 f"has {max_at_one_site} points, more than the {k} nearest neighbors "
--- a/libpysal/graph/_kernel.py
+++ b/libpysal/graph/_kernel.py
@@ -7,7 +7,6 @@ from ._utils import (
     _induce_cliques,
     _jitter_geoms,
     _reorder_adjtable_by_ids,
-    # _resolve_islands,
     _sparse_to_arrays,
     _validate_geometry_input,
 )
@@ -125,6 +124,8 @@ def _kernel(
         parameter for minkowski metric, ignored if metric != "minkowski".
     taper : bool (default: True)
         remove links with a weight equal to zero
+    resolve_isolates : bool
+        Try to resolve isolates. Can be disabled if we are dealing with cliques later.
     """
     if metric != "precomputed":
         coordinates, ids, _ = _validate_geometry_input(
@@ -202,11 +203,7 @@ def _kernel(
     if taper:
         d.eliminate_zeros()
 
-    heads, tails, weights = _sparse_to_arrays(
-        d, ids=ids, resolve_isolates=resolve_isolates
-    )
-
-    return heads, tails, weights
+    return _sparse_to_arrays(d, ids=ids, resolve_isolates=resolve_isolates)
 
 
 def _knn(coordinates, metric="euclidean", k=1, p=2, coincident="raise"):
--- a/libpysal/graph/_kernel.py
+++ b/libpysal/graph/_kernel.py
@@ -7,7 +7,7 @@ from ._utils import (
     _induce_cliques,
     _jitter_geoms,
     _reorder_adjtable_by_ids,
-    _resolve_islands,
+    # _resolve_islands,
     _sparse_to_arrays,
     _validate_geometry_input,
 )
@@ -79,6 +79,7 @@ def _kernel(
     p=2,
     taper=True,
     coincident="raise",
+    resolve_isolates=True,
 ):
     """
     Compute a kernel function over a distance matrix.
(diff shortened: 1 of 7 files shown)


libpysal/graph/_triangulation.py
```diff
@@ -52,46 +50,23 @@ def tri_with_validation(
...
-                    coincident_lut.geometry,
-                    ids=coincident_lut.index,
-                    valid_geometry_types=_VALID_GEOMETRY_TYPES,
-                )
-            else:
-                raise ValueError(
-                    f"Recieved option coincident='{coincident}', but only options "
-                    "'raise','clique','jitter' are suppported."
-                )
+        if coincident == "jitter":
```
ljwolf, line 63: As I read the code, I think this has to be implemented this way---the coplanar check happens in QHull, we can't know whether we have to use our jittering until we run the triangulation.  

This means we have a tradeoff: 
- have "two" jitter functions (one for KNN and one for triangulated weights) 
- **or** always jitter inputs when coincident='jitter', regardless of if it's needed. I'd recommend doing it here and in KNN, to keep consistent. 

I think between the two, I'd prefer option 1. What do you think?

libpysal/graph/_triangulation.py
```diff
@@ -126,11 +102,32 @@ def tri_with_validation(
                 fill_value = _kernel_functions[kernel](
                     numpy.array([0]), bandwidth
                 ).item()
-            adjtable = _induce_cliques(adjtable, coincident_lut, fill_value=fill_value)
```
ljwolf, line 129: Fine to use a different implementation of clique expansion, but can we keep this factored into a function that can be called from [`knn`](https://github.com/pysal/libpysal/blob/main/libpysal/graph/_kernel.py#L267) too?

ljwolf (changes requested):

libpysal/graph/_triangulation.py
martinfleis: By two jitter functions you mean using the QJ QHull option? We can easily use the same one we have in knn (as done here). 

I would prefer doing the latter and always jitter, even in KNN than having different behaviour in different places solving the same. We still have as we rely on different precision here but it is consistent from user perspective.

libpysal/graph/_triangulation.py
```diff
@@ -543,13 +551,24 @@ def _filter_relativehood(edges, coordinates, return_dkmax=False):
     return out, r
 
 
-def _voronoi_edges(coordinates):
+def _voronoi_edges(coordinates, coincident):
     dt = spatial.Delaunay(coordinates)
+
+    if dt.coplanar.shape[0] > 0 and coincident == "raise":
+        raise ValueError(
```
jGaboardi, line 558: Perhaps it might be wise to have our own custom error this, something like `CoplanarError`? Or would that be too involved?

Since similar ones use used in [other methods](https://github.com/pysal/libpysal/blob/181784671da99af1f211cc5a3cbd1f2c4a553ce7/libpysal/graph/_kernel.py#L239), maybe not a bad idea?

ljwolf: Also +1 on using "coplanar" throughout, rather than "coincident" and "coplanar"

jGaboardi: Good point on that. I was about ask if there was a subtle difference between the two that I wasn't aware of.

libpysal/graph/_kernel.py
```diff
@@ -201,20 +204,21 @@ def _kernel(
     if taper:
         d.eliminate_zeros()
 
-    heads, tails, weights = _sparse_to_arrays(d, ids=ids)
+    return _sparse_to_arrays(d, ids=ids, resolve_isolates=resolve_isolates)
 
-    return _resolve_islands(heads, tails, ids, weights)
```
martinfleis, line 206: `_resolve_islands` is called within `_sparse_to_arrays`, no need to call it twice

libpysal/graph/_kernel.py
```diff
@@ -201,20 +204,21 @@ def _kernel(
...
-
-def _knn(coordinates, metric="euclidean", k=1, p=2, coincident="raise"):
+def _knn(coordinates, metric="euclidean", k=1, p=2, coplanar="raise"):
     """internal function called only within _kernel, never directly to build KNN"""
     coordinates, ids, geoms = _validate_geometry_input(
         coordinates, ids=None, valid_geometry_types=_VALID_GEOMETRY_TYPES
     )
-    n_coincident = geoms.geometry.duplicated().sum()
+    if coplanar == "jitter":
+        coordinates, geoms = _jitter_geoms(coordinates, geoms=geoms)
```
martinfleis, line 216: Always jitter if `coplanar == "jitter"`

martinfleis:
Okay, this is now significantly updated.

- switch terminology from `coincident` to `coplanar` as suggested by @ljwolf  (cc @knaaptime in case you are using "coincident" somewhere in your materials, this is technically a breaking change but we're still in experimental mode)
- knn uses the same coplanar resolver as triangulation. that also required new function to build coplanar lookup to match the one coming from Delaunay
- raising CoplanarError
- always jitter when `coplanar="jitter"`, no matter if there is a coplanar point or not

The performance is about the same as it was before.

ljwolf (approved):

jGaboardi (approved):

Issue #687, opened 2024-02-28
Ensure `Graph.sparse` is robust enough


martinfleis:
I've been recently reading some adjacencies with `fastparquet` instead of `pyarrow` and noticed that the resulting sparse array is different even though the Series are considered equal. The reason is that the MultiIndex is created differently, leading to different underlying `MultiIndex.codes`, which, I suppose, somehow changes the conversion to sparse. We need to revise this and ensure that the way pandas creates sparse is robust enough and always yields properly sorted array or deploy our own way of generating sparse (which we had at some point as I wrote it so it will be in the commit history).

this is mostly a note to myself to check this before Graph leaves experimental zone.

ljwolf:
Is there a reproducible example we can use to test against? Round-tripping was a core design goal...

martinfleis:
Yes, here:

```py
In [1]: import pandas as pd
   ...: from libpysal.graph import Graph

In [2]: path = "https://github.com/Urban-Analytics-Technology-Platform/demoland-engine/raw/3fba672d08d7d09bdb692c35594064dbcdd69662/data/tyne_and_wear/matrix.parquet"

In [3]: adjacency_pyarrow = pd.read_parquet(path, engine="pyarrow")["weight"]
   ...: adjacency_fastparquet = pd.read_parquet(path, engine="fastparquet")["weight"]

In [4]: pd.testing.assert_series_equal(adjacency_pyarrow, adjacency_fastparquet)

In [5]: graph_pyarrow = Graph(adjacency_pyarrow, "r", is_sorted=True)
   ...: graph_fastparquet = Graph(adjacency_fastparquet, "r", is_sorted=True)

In [6]: graph_pyarrow.equals(graph_fastparquet)
Out[6]: True

In [7]: sparse_pyarrow = graph_pyarrow.sparse
   ...: sparse_fastparquet = graph_fastparquet.sparse

In [8]: (sparse_pyarrow == sparse_fastparquet).todense().all()
Out[8]: False

In [9]: sparse_pyarrow.todense()[0][0] # <- this is correct
Out[9]: 0.0

In [10]: sparse_fastparquet.todense()[0][0] # <- this is wrong
Out[10]: 0.007518796992481203
```

I haven't looked into pandas source code, but my guess is that it is because:
```py
In [11]: (adjacency_pyarrow.index.codes[1] == adjacency_fastparquet.index.codes[1]).all()
Out[11]: False

In [12]: adjacency_fastparquet.index.codes
[...]

PR #715, opened 2024-05-30
PERF: sorting-related improvements in Graph

martinfleis:
I've done some profiling which led to these changes.

- avoiding re-creation of adjacency and re-sorting already sorting arrays in `_sparse_to_arrays` - cuts time to about half
- faster `_reorder_adjtable_by_ids` (no apply) (10x faster than original)
- `_validate_sparse_input` is removed as it is not used anywhere

We still do sorting multiple times in many places but if I short-circuit that, we hit #687 and issues with sparse round tripping. The final sorting either needs to happen in `__init__` as now or would need to happen in `.sparse`. Keeping it in the constructor is probably better idea as we know that a Graph is always correct as is.

### DIFF
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -16,7 +16,7 @@ class CoplanarError(ValueError):
     pass
 
 
-def _sparse_to_arrays(sparray, ids=None, resolve_isolates=True):
+def _sparse_to_arrays(sparray, ids=None, resolve_isolates=True, return_adjacency=False):
     """Convert sparse array to arrays of adjacency
 
     When we know we are dealing with cliques, we don't want to resolve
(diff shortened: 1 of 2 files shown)


libpysal/graph/_utils.py
```diff
@@ -214,13 +223,6 @@ def _validate_geometry_input(geoms, ids=None, valid_geometry_types=None):
     )
 
 
-def _validate_sparse_input(sparse, ids=None):
```
jGaboardi, line 217: Was `_validate_sparse_input()` simply not being used anywhere?

martinfleis: Yes, this is directly in https://github.com/pysal/libpysal/blob/6dba8b68a695b0165d7361d47e8fc3f2cc0eecae/libpysal/graph/_kernel.py#L136-L140

Since it is one liner and single use, I don't think we need to have a function for it anyway.

jGaboardi (approved):

PR #716, opened 2024-05-31
Categorical spatial lag using the Graph

sjsrey:
This has two changes from the creation of categorical spatial lags using W

1. `categorical` is now an option in `_lag_spatial`, to unify treatment of the lag.
2. This will raise an exception in the case where `categorical=True` , `ties=raise`, and ties are encountered.
"Explicit is better than implicit."

### DIFF
--- a/libpysal/graph/_spatial_lag.py
+++ b/libpysal/graph/_spatial_lag.py
@@ -1,8 +1,12 @@
-def _lag_spatial(graph, y):
+import numpy as np
+import pandas as pd
+
+
+def _lag_spatial(graph, y, categorical=False, ties='raise'):
     """Spatial lag operator
 
-    If w is row standardized, returns the average of each observation's neighbors;
-    if not, returns the weighted sum of each observation's neighbors.
+    Constructs spatial lag based on neighbor relations of the graph.
+
 
     Parameters
     ----------
@@ -10,11 +14,52 @@ def _lag_spatial(graph, y):
         libpysal.graph.Graph
     y : array
         numpy array with dimensionality conforming to w
+    categorical : bool
+        True if y is categorical, False if y is continuous.
+    ties : {'raise', 'random', 'tryself'}, optional
+        Policy on how to break ties when a focal unit has multiple
+        modes for a categorical lag.
+        - 'raise': This will raise an exception if ties are
+          encountered to alert the user (Default).
+        - 'random': Will break ties randomly.
+        - 'tryself': Add a self-weight to attempt to break the tie
+          with the focal label. If the self-weight does not break a
+          tie, or the self-weight induces a tie, the tie will be be
+          broken randomly.
+
 
     Returns
     -------
     numpy.array
-        array of numeric values for the spatial lag
+        array of numeric|categorical values for the spatial lag
+
+
+    Examples
+    --------
+    >>> from libpysal.graph._spatial_lag import _lag_spatial
+    >>> import numpy as np
+    >>> from libpysal.weights.util import lat2W
+    >>> from libpysal.graph import Graph
+    >>> graph = Graph.from_W(lat2W(3,3))
+    >>> y = np.arange(9)
+    >>> _lag_spatial(graph, y)
+    array([ 4.,  6.,  6., 10., 16., 14., 10., 18., 12.])
+
+    Categorical Lag (no ties)
+    >>> y = np.array([*'ababcbcbc'])
+    >>> _lag_spatial(graph, y, categorical=True)
+    array(['b', 'a', 'b', 'c', 'b', 'c', 'b', 'c', 'b'], dtype=object)
+
+    Handling ties
+    >>> y[3] = 'a'
+    >>> np.random.seed(12345)
+    >>> _lag_spatial(graph, y, categorical=True, ties='random')
+    array(['a', 'a', 'b', 'c', 'b', 'c', 'b', 'c', 'b'], dtype=object)
+    >>> _lag_spatial(graph, y, categorical=True, ties='random')
+    array(['b', 'a', 'b', 'c', 'b', 'c', 'b', 'c', 'b'], dtype=object)
+    >>> _lag_spatial(graph, y, categorical=True, ties='tryself')
+    array(['a', 'a', 'b', 'c', 'b', 'c', 'b', 'c', 'b'], dtype=object)
+
     """
     sp = graph.sparse
     if len(y) != sp.shape[0]:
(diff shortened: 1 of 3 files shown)


libpysal/graph/_spatial_lag.py
```diff
@@ -1,25 +1,145 @@
...
+                "issue, consider setting `ties='tryself'` "
+                "or `ties='random'` or consult the documentation "
+                "about ties and the categorical spatial lag."
+            )
+        elif ties in ('random', 'tryself', 'raise'):
+            return gb.apply(_get_categorical_lag).values
+
+        else:
+            raise ValueError(
+                f"Recieved option ties='{ties}', but only options "
```
jGaboardi, line 103:
```suggestion
                f"Received option ties='{ties}', but only options "
```

jGaboardi (commented):
[Three `ruff` errors](https://results.pre-commit.ci/run/github/81501824/1717157113.dVzLQnFFRC2MHeouslkyaw) (2 linting & 1 formatting) need to to fixed locally and re-pushed.

```
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

libpysal/graph/tests/test_spatial_lag.py:1:1: I001 [*] Import block is un-sorted or un-formatted
libpysal/graph/tests/test_spatial_lag.py:62:13: F841 Local variable `yl` is assigned to but never used
Found 2 errors.
[*] 1 fixable with the `--fix` option (1 hidden fix can be enabled with the `--unsafe-fixes` option).

ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

2 files reformatted, 151 files left unchanged
```

libpysal/graph/_spatial_lag.py
```diff
@@ -1,25 +1,145 @@
...
     sp = graph.sparse
     if len(y) != sp.shape[0]:
         raise ValueError(
             "The length of `y` needs to match the number of observations "
             f"in Graph. Expected {sp.shape[0]}, got {len(y)}."
         )
+
+    if categorical:
+        if ties == 'tryself':
+            graph = graph.assign_self_weight()
```
martinfleis, line 83: I am not sure if we want to do `tryself` in this way. This can result in a situation when without self weights, there's no tie but with it there is and we resolve it randomly to a category that is not dominant without a self weight.

libpysal/graph/_spatial_lag.py
```diff
@@ -1,25 +1,145 @@
...
+    array(['a', 'a', 'b', 'a', 'b', 'c', 'b', 'c', 'b'], dtype=object)
+
     """
     sp = graph.sparse
     if len(y) != sp.shape[0]:
         raise ValueError(
             "The length of `y` needs to match the number of observations "
             f"in Graph. Expected {sp.shape[0]}, got {len(y)}."
         )
+
```
martinfleis, line 80:
```suggestion
   if ( isinstance(y.dtype, CategoricalDtype)
        or pd.api.types.is_object_dtype(y.dtype)
        or pd.api.types.is_bool_dtype(y.dtype)
        or pd.api.types.is_string_dtype(y.dtype)
    ):
        categorical = True
```

This way we can automatically detect categorical data when it is clear. A user would then need to manually specify only if they have int labels for categories.

martinfleis (commented):
This needs to be propagated to `base.py`

https://github.com/pysal/libpysal/blob/5bda115749e6c8a57323fe86d74487dff3878a4c/libpysal/graph/base.py#L1618

libpysal/graph/_spatial_lag.py
sjsrey: Agreed.

Refactoring to avoid inducing ties with self-weights

libpysal/graph/tests/test_spatial_lag.py
```diff
@@ -29,3 +34,34 @@ def test_lag_spatial(self):
         yl = _lag_spatial(g_row, y)
         ylc = np.array([2.0, 2.0, 3.0, 3.33333333, 4.0, 4.66666667, 5.0, 6.0, 6.0])
         np.testing.assert_array_almost_equal(yl, ylc)
+
+    def test_lag_spatial_categorical(self):
+        yl = _lag_spatial(self.gc, self.yc, categorical=True)
```
martinfleis, line 39:
```suggestion
        yl = _lag_spatial(self.gc, self.yc)
```
Let's leave `categorical=True` out at least once to check the automatic detection

libpysal/graph/tests/test_spatial_lag.py
```diff
@@ -29,3 +34,34 @@ def test_lag_spatial(self):
...
+        gc = graph.Graph.from_dicts(neighbors)
+        self.yc[3] = "b"
+        yls = _lag_spatial(gc, self.yc, categorical=True, ties="tryself")
+        assert yls[0] in ["b", "a"]
+        self.yc[3] = "a"
+        yls = _lag_spatial(gc, self.yc, categorical=True, ties="tryself")
+        assert yls[0] == "a"
+
+    def test_ties_raise(self):
+        with pytest.raises(ValueError):
```
martinfleis, line 65:
```suggestion
        with pytest.raises(ValueError, match="There are 2 ties that must be broken"):
```

martinfleis (commented):
Two minor suggestions in tests.

As mentioned before, this needs to propagate to https://github.com/pysal/libpysal/blob/5bda115749e6c8a57323fe86d74487dff3878a4c/libpysal/graph/base.py#L1618-L1635

ljwolf:
What would you think about dispatching to categorical lag if the type is [categorical](https://pandas.pydata.org/docs/reference/api/pandas.Categorical.html)?

I can appreciate that this might make the lag type implicit, but if someone has cast the variable to a categorical already, I think we can infer that they need the appropriate lag function?

martinfleis:
@ljwolf that is what we do on l. 87, no?

libpysal/graph/base.py
```diff
@@ -1613,22 +1613,37 @@ def higher_order(self, k=2, shortest_path=True, diagonal=False, lower_order=Fals
...
+
 
         Returns
         -------
-        numpy.ndarray
-            array of numeric values for the spatial lag
+        numpy.array
+            array of numeric|categorical values for the spatial lag
         """
         return _lag_spatial(self, y)
```
martinfleis, line 1648:
```suggestion
        return _lag_spatial(self, y, categorical=categorical, ties=ties)
```

libpysal/graph/_spatial_lag.py
```diff
@@ -1,25 +1,181 @@
...
+           4.66666667, 5.        , 6.        , 6.        ])
+
+
+    Categorical Lag (no ties)
+    >>> y = np.array([*'ababcbcbc'])
+    >>> _lag_spatial(graph, y, categorical=True)
+    array(['b', 'a', 'b', 'c', 'b', 'c', 'b', 'c', 'b'], dtype=object)
+
+    Handling ties
+    >>> y[3] = 'a'
```
pedrovma, line 65:
```suggestion
>>> y[3] = 'b'

File: libpysal/graph/base.py [hunk 8]
```diff
@@ -1613,24 +1613,39 @@ def higher_order(self, k=2, shortest_path=True, diagonal=False, lower_order=Fals
...
-        of each observation's neighbors.
+        Constructs spatial lag based on neighbor relations of the graph.
+
 
         Parameters
         ----------
-        y : array-like
-            array-like (N,) shape where N is equal to number of observations in self.
+        graph : Graph
+            libpysal.graph.Graph
```
martinfleis, line 1625:
```suggestion
```
that is `self`

libpysal/graph/base.py
```diff
@@ -1613,24 +1613,39 @@ def higher_order(self, k=2, shortest_path=True, diagonal=False, lower_order=Fals
...
+            tie will be be broken randomly. If the focal unit has a
+            self-weight, focal label is not used to break any tie,
+            rather any tie will be broken randomly.
+
 
         Returns
         -------
-        numpy.ndarray
-            array of numeric values for the spatial lag
+        numpy.array
```
martinfleis, line 1645:
```suggestion
        numpy.ndarray
```
[`numpy.array`](https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array) is a function constructing the class [`numpy.ndarray`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html). So the original was correct.

martinfleis (approved):
Just two minor nits on the docstring. Also, ruff complains at the moment. But once that is solved, happy to merge.

martinfleis:
pre-commit autofix

jGaboardi (approved):

PR #717, opened 2024-06-04
ENH: include Graph.describe() to describe neighbourhood values

u3ks:
This PR adds a method to the graph api which takes an array of values and calculates descriptive statistics within each neighborhood. 
Optionally, some neighbors can be filtered out based on the percentiles of the passed values.
The supported stats are - "count", "mean", "median", "std", "min", "max", "sum", "nunique" and "mode".

The method similar to .apply, but all values are calculated in one grouping operation and all functions are jitted.

### DIFF
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -304,7 +304,7 @@ def _limit_range(values, index, low, high):  # noqa: ARG001
     nan_tracker = np.isnan(values)
 
     if (not nan_tracker.all()) & (len(values[~nan_tracker]) > 2):
-        lower, higher = np.percentile(values, (low, high))
+        lower, higher = np.nanpercentile(values, (low, high))
     else:
         return ~nan_tracker
 
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -298,6 +298,7 @@ def _mode(values, index):  # noqa: ARG001
     counts = np.diff(idx)
     return unique[np.argmax(counts)]
 
+
 @njit
 def _limit_range(values, index, low, high):  # noqa: ARG001
     nan_tracker = np.isnan(values)
@@ -309,7 +310,8 @@ def _limit_range(values, index, low, high):  # noqa: ARG001
 
     return (lower <= values) & (values <= higher)
 
-def _compute_stats(grouper, to_compute:list[str]|None=None):
+
+def _compute_stats(grouper, to_compute: list[str] | None = None):
     """Fast compute of "count", "mean", "median", "std", "min", "max", \\
     "sum", "nunique" and "mode" within a grouper object. Using numba.
 
@@ -334,13 +336,22 @@ def _compute_stats(grouper, to_compute:list[str]|None=None):
         )
 
     if to_compute is None:
-        to_compute = ["count", "mean", "median",
-                    "std", "min", "max", "sum", "nunique", "mode"]
-    agg_to_compute = [f for f in to_compute if f != 'mode']
+        to_compute = [
+            "count",
+            "mean",
+            "median",
+            "std",
+            "min",
+            "max",
+            "sum",
+            "nunique",
+            "mode",
+        ]
+    agg_to_compute = [f for f in to_compute if f != "mode"]
     stat_ = grouper.agg(agg_to_compute)
-    if 'mode' in to_compute:
+    if "mode" in to_compute:
         if HAS_NUMBA:
-            stat_["mode"] = grouper.agg(_mode, engine='numba')
+            stat_["mode"] = grouper.agg(_mode, engine="numba")
         else:
             stat_["mode"] = grouper.agg(lambda x: _mode(x.values, x.index))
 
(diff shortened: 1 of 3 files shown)


martinfleis:
Just to add some context to this. As we are refactoring momepy, we realised that we rely very often on this internal function, which is fairly generic and shall be tied directly to Graph.

The idea behind the `q` limiting the range is coming from morphology. We often want to get some sort of a spatial average but given the high likelihood of outliers (think of a church in the middle of a neighborhood), we can't include all the values within each neighborhood.

ljwolf:
I think, for generality, this should be called a [truncated](https://en.wikipedia.org/wiki/Truncation_(statistics)) or [trimmed](https://en.wikipedia.org/wiki/Least_trimmed_squares) reduction?

This is very useful generally... @weikang9009 and I have been working on related concepts recently, so it'd be very nice to have something core here!

libpysal/graph/tests/test_base.py
```diff
@@ -1129,3 +1130,46 @@ def test_aggregate(self):
             contig.aggregate(lambda x: np.exp(np.sum(x))),
             expected,
         )
+
+    def test_describe(self):
```
martinfleis, line 1134: We should also include a tiny test case (nybb?) with an isolate. Could be just a simple mean, max.

martinfleis:
> I think, for generality, this should be called a [truncated](https://en.wikipedia.org/wiki/Truncation_(statistics)) or [trimmed](https://en.wikipedia.org/wiki/Least_trimmed_squares) reduction/lag?

Only if `q` is not None. Otherwise it is just a generic lag. I am also not sure what can be called a lag (nunique?). The `describe` terminology comes from pandas. It felt close enough to what we're doing here.

martinfleis (approved):

libpysal/graph/_utils.py
```diff
@@ -274,3 +283,107 @@ def _reorder_adjtable_by_ids(adjtable, ids):
...
+
+def _compute_stats(grouper, to_compute: list[str] | None = None):
+    """Fast compute of "count", "mean", "median", "std", "min", "max", \\
+    "sum", "nunique" and "mode" within a grouper object. Using numba.
+
+    Parameters
+    ----------
+    grouper : pandas.GroupBy
+        Groupby Object which specifies the aggregations to be performed.
+    to_compute : List[str]
```
jGaboardi, line 322:
```suggestion
    to_compute : list[str]
```

libpysal/graph/_utils.py
```diff
@@ -274,3 +283,107 @@ def _reorder_adjtable_by_ids(adjtable, ids):
...
+    return (lower <= values) & (values <= higher)
+
+
+def _compute_stats(grouper, to_compute: list[str] | None = None):
+    """Fast compute of "count", "mean", "median", "std", "min", "max", \\
+    "sum", "nunique" and "mode" within a grouper object. Using numba.
+
+    Parameters
+    ----------
+    grouper : pandas.GroupBy
```
jGaboardi, line 320: Should this be ``pandas.Grouper``?

libpysal/graph/_utils.py
```diff
@@ -274,3 +283,107 @@ def _reorder_adjtable_by_ids(adjtable, ids):
...
+    -------
+    DataFrame
+    """
+
+    if not HAS_NUMBA:
+        warnings.warn(
+            "The numba package is used extensively in this module"
+            " to accelerate the computation of graphs. Without numba,"
+            " these computations may become unduly slow on large data.",
+            stacklevel=3,
```
jGaboardi, line 335: Are we sure about `stacklevel=3` here? Or should it be ``stacklevel=2``?

libpysal/graph/base.py
```diff
@@ -1993,6 +1995,69 @@ def aggregate(self, func):
...
+        neighbourhood of each node. Optionally, the values can be limited to a certain
+        quantile range before computing the statistics.
+
+        Notes
+        -----
+        The index of ``values`` must match the index of the graph.
+
+        Weight values do not affect the calculations, only adjacency does.
+
+        Returns nan for all isolates.
```
jGaboardi, line 2016: Maybe 'nan' is OK here, but also maybe `NaN` or `numpy.nan` (or something else?)

Probably not a big deal either way.

libpysal/graph/base.py
```diff
@@ -1993,6 +1995,69 @@ def aggregate(self, func):
...
+        Parameters
+        ----------
+        y : NDArray[np.float_] | Series
+            An 1D array of numeric values to be described.
+        q : tuple[float, float] | None, optional
+            Tuple of percentages for the percentiles to compute.
+            Values must be between 0 and 100 inclusive. When set, values below and above
+            the percentiles will be discarded before computation of the statistics.
+            The percentiles are computed for each neighborhood. By default None.
+        statistics : List[str] | None
```
jGaboardi, line 2031:
```suggestion
        statistics : list[str] | None
```

libpysal/graph/_utils.py
martinfleis: It's called from within `describe`, so it is 3.

libpysal/graph/_utils.py
u3ks: I think the pandas.Grouper is another type of object, that is used for filtering columns , i used the name grouper since its used in other functions and the type is groupby since, pandas groupy returns a groupby object

libpysal/graph/base.py
u3ks: changed to numpy.nan

libpysal/graph/base.py
```diff
@@ -2040,7 +2040,7 @@ def describe(
         """
 
         if not isinstance(y, pd.Series):
-            y = pd.Series(y)
+            y = pd.Series(y, index=self.unique_ids)
```
martinfleis, line 2043: Looking at this, we may want to check that the y.index matches self.unique_ids in case of a custom Series is passed. I suppose that non-matching index may break this.

jGaboardi (approved):

knaaptime:
i think i would call this `describe_cardinalities` or something because "Graph.describe() to describe neighbourhood values" implies we're looking at the neighbor values

martinfleis:
But this is not describing cardinalities, no? Where cardinality is a number of elements in a set. It is describing distribution of values within a neighbourhood.

knaaptime:
oh i see. It was this note on line 2014 that tripped me up:

'Weight values do not affect the calculations, only adjacency does.'

Issue #695, opened 2024-03-28
import libpysal stuck on loading remote examples


TonyPod:
Hi, there! When I'm importing libpysal, it runs function "poll_remotes" (examples/remotes.py) which has a line like this:

`page = requests.get(url)`

In offline machines, it get stuck forever. Could you add a timeout or else?

jGaboardi:
Closed via #718

PR #718, opened 2024-06-07
Add timeout in request to handle off-line use cases in examples

sjsrey:
#695

### DIFF
--- a/libpysal/examples/remotes.py
+++ b/libpysal/examples/remotes.py
@@ -643,7 +643,7 @@ def _remote_data():
 
     url = "https://geodacenter.github.io/data-and-lab//"
     try:
-        page = requests.get(url)
+        page = requests.get(url, timeout=10)
     except:  # noqa: E722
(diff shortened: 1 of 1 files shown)


libpysal/examples/remotes.py
```diff
@@ -643,7 +643,7 @@ def _remote_data():
 
     url = "https://geodacenter.github.io/data-and-lab//"
     try:
-        page = requests.get(url)
+        page = requests.get(url, timeout=10)
```
jGaboardi, line 646: Curious if there is a specific reason for 10 seconds?

jGaboardi (approved):

libpysal/examples/remotes.py
martinfleis, line 646:
```suggestion
        page = requests.get(url, timeout=(10, None))
```

Let's set only the connect timeout and leave the read timeout without a limit. Some examples are quite large and 10s may be too short.

martinfleis (approved):

PR #726, opened 2024-06-26
Use fixtures for test data that uses the network

QuLogic:
2. [x] This pull request is directed to the `pysal/master` branch.  **This should say `main` now.**
4. [n/a] You have [assigned a
Currently, if you have no network, the marked tests will fail. However, much worse than that, because some datasets are loaded at the top level, pytest collection will ERROR out. This makes it impossible to even run the tests that don't use the network.

Moving these datasets from the top level to fixtures means that pytest collection works, and those tests can be skipped by deselecting the `network` marker.


martinfleis (approved):
Thanks! All good with me. Are you aware of an easy way of testing this on CI? I can't promise that we don't introduce a new test requiring network without CI shouting at me.

QuLogic:
On Linux, you can use `unshare` or `systemd-run`: https://unix.stackexchange.com/a/454767

martinfleis:
@jGaboardi do we want to add another action testing this? It needs to be an action, not a job to avoid caching.

jGaboardi:
> @jGaboardi do we want to add another action testing this? It needs to be an action, not a job to avoid caching.

I'm not against that at all. Is it something we need in this PR? If not let merge. If so, I can try to put it on the To Do list for next month.

martinfleis:
>  Is it something we need in this PR? 


not at all. I am also not sure if we need it at all.

jGaboardi (approved):

Issue #712, opened 2024-05-28
Migration guide W->Graph


sjsrey:
Two teaching materials where one is derived from the other and updated to use Graph instead of W.

- Dani's W - https://darribas.org/gds_course/content/bE/lab_E.html
- Martin's Graph https://martinfleischmann.net/sds/chapter_04/hands_on.html

Apart from the different datasets, it is nearly 1:1 and can be used as a basis of the migration guide.

_edited by @martinfleis_

knaaptime:
following the discussion from the [June dev meeting](https://github.com/orgs/pysal/discussions/1278), we want to open a discussion about Graph nomenclature (specifically when you need to work with multiple graphs simultaneously e.g. from libpysal and networkx at once) it might be easier to rename `Graph` to `GeoGraph` or similar. Opening here for discussion so we can vote in the next dev meeting

martinfleis:
I am tentatively for the renaming as it is going to be clearer to the user what is required where and avoids some confusion between different graph representations. But I also don't feel super strongly about this. It's also going to break some of my research code and teaching materials but I can deal with that easily.

jGaboardi:
xref: https://github.com/pysal/momepy/issues/579

I have no strong opinions either way, so probably not a bad to go with the more distinctive `GeoGraph`. Wasn't `GeoGraph` the original name that was under consideration when the refactor first started out?

martinfleis:
Existence of this https://geograph.readthedocs.io/en/latest/ may affect the decision here...

jGaboardi:
Are there any other alternative names being thrown around?

knaaptime:
at the steering committee meeting this morning we voted to keep `Graph` for the time being

jGaboardi:
complete via #733

PR #733, opened 2024-06-30
W to g guide

sjsrey:
This is a start towards addressing #712.
Seeking feedback and discussion.

It will add a note to the index page:

image

This also adds a dynamically generated rst file that will report on the members for W and Graph, which is the target of the note.

image

In addition, I will add a notebook that builds on the suggestions in #712 to have more of a user facing orientation.

This is intended to be dynamic and assist us as we fully build out the `Graph`  moving forward.

Once we are done with the transition, we can remove this dynamically generated rst file, if so desired.


knaaptime:
sweet

docs/Makefile
```diff
@@ -12,23 +12,18 @@ BUILDDIR      = _build
...
+
+# Custom target to generate rst files
+generate-rst:
+	@echo "Generating custom rst files..."
+	@python ../scripts/w_g_migration.py
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
-	@rsync -r --exclude '.ipynb_checkpoints/' ../notebooks/ ./notebooks/
```
jGaboardi, line 20: Are these deletions simply relics from our older infrastructure?

sjsrey: Yes

jGaboardi (approved):

martinfleis:
I like that automatic comparison but in some cases, it may be worth adding a note explaining _why_ there's no match and what is recommended to do instead. E.g Graph does not have `islands` but has `isolates` instead etc.

I also suppose that merge shall wait until the notebook is complete, right?

sjsrey:
> I like that automatic comparison but in some cases, it may be worth adding a note explaining _why_ there's no match and what is recommended to do instead. E.g Graph does not have `islands` but has `isolates` instead etc.

Yes, this was intended to identify all the cases we need to surface in the recommendations. I wasn't sure if we should put the recommendations in these tables, or treat them in the migration notebook where we can have a bit more narrative for the user.


> 
> I also suppose that merge shall wait until the notebook is complete, right?

Definitely.

martinfleis:
>  treat them in the migration notebook where we can have a bit more narrative for the user.

That may be better. We just need to make a link to the notebook clear and visible from the RST with the tables.

knaaptime (approved):

sjsrey:
This is still WIP. I have split the migration material and member comparisons apart for different audiences (users,devs).
This also reorganizes the notebooks and changes the tutorial to a user-guide.

The idea is to keep the component comparison around to automatically generate the W-G tables. These will get updated each time the docs are built.

Meanwhile, the user migration guide is a notebook that can now be fleshed out for the current state of the Graph.
`docs/user-guide/graph/w_g_migration.ipynb` which can now be updated.

PR #739, opened 2024-07-05
fix higher_order contiguity including lower order

### LINKED ISSUE #738: BUG: higher_order(lower_order=True) not working in edge cases
I encountered super strange behaviour of `higher_order`, both in old W and new Graph implementation.

```py
g = graph.Graph.from_arrays([0, 1, 2, 3, 3, 4, 4], [0, 3, 4, 1, 4, 2, 3], [0, 1, 1, 1, 1, 1, 1])
g.adjacency

focal  neighbor
0      0           0
1      3           1
2      4           1
3      1           1
       4           1
4      2           1
       3           1
Name: weight, dtype: int64

g.higher_order(k=2, lower_order=True).adjacency

focal  neighbor
0      0           0
1      4           1
2      3           1
3      2           1
4      1           1
Name: weight, dtype: int64

```
As you can see, this includes **only**  neighbors of second order, not the first one as it should. The reason is that the underlying matrix power computation does not include them.

```
g.sparse.todense()

array([[0., 0., 0., 0., 0.],
       [0., 0., 0., 1., 0.],
       [0., 0., 0., 0., 1.],
       [0., 1., 0., 0., 1.],
       [0., 0., 1., 1., 0.]])

sparse.linalg.matrix_power(int_g.sparse, 2).todense()

array([[0., 0., 0., 0., 0.],
       [0., 1., 0., 0., 1.],
       [0., 0., 1., 1., 0.],
       [0., 0., 1., 2., 0.],
       [0., 1., 0., 0., 2.]])
```

The solution is to sum the original sparse with the new one. It seems to be an edge case resulting from extremely sparse matrix but could be real nevertheless. 

I _think_ that the solution is to sum the result of matrix power(s) with the original sparse to ensure a value is present. Here https://github.com/pysal/libpysal/blob/01c1f8026fe8b52913116af0300bba239d58e310/libpysal/graph/base.py#L1607-L1608

martinfleis:
Fixes #738 

Funnily enough, we have the same line of code in `weights/raster.py`, where it is correct.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1605,7 +1605,7 @@ class Graph(SetOpsMixin):
         sp = binary.sparse
 
         if lower_order:
-            wk = sum(sparse.linalg.matrix_power(sp, x) for x in range(2, k + 1))
+            wk = sum(sparse.linalg.matrix_power(sp, x) for x in range(1, k + 1))
             shortest_path = False
         else:
             wk = sparse.linalg.matrix_power(sp, k)
--- a/libpysal/graph/tests/test_base.py
+++ b/libpysal/graph/tests/test_base.py
@@ -983,7 +983,11 @@ class TestBase:
         lower = cont.higher_order(2, lower_order=True)
         assert lower == expected
 
-    def test_higher_order_inclusive(self): #GH738
+    @pytest.mark.skipif(
+        Version(scipy_version) < Version("1.12.0"),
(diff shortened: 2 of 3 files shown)


martinfleis:
(the dev failure is coming from incompatibility between pandas main and geopandas, nothing to solve here)

jGaboardi (approved):

PR #740, opened 2024-07-06
ENH: add xarray interface to Graph

martinfleis:
This attempts to mirror the functionality of the `weights.raster` module within the `Graph`. I tried to re-use as much of the code from `weights` as I could which required quite a bit of refactoring there to avoid too many unnecessary conversions back and forth.

Contrary to the original implementation, we don't need any new public attribute to facilitate using the graph within esda etc but I have added one private one. If it is not present it is not the end of the world either, you just get `level_0`, `level_1` etc as the dimension names in the DataArray. But I tried to preserve it in most common operations.

Compared to the original, there are some minor differences when putting it work. For example, computing LISA would look like this, using the tif from https://data.worldpop.org/GIS/Population/Global_2000_2020/2019/LUX/lux_ppp_2019.tif (which needs to be pulled locally).

```py
da = rioxarray.open_rasterio("/Users/martin/Downloads/lux_ppp_2019.tif")
g_queen = graph.Graph.build_raster_contiguity(da, rook=False)
data = da.to_series()[g_queen.unique_ids]  # here you would use W.index in weights.raster

# If you have missing data
data_non_missing = data[data != da.rio.nodata]
data_non_missing_index = data_non_missing.index.to_numpy() # flatten the multi index to tuples

# get subgraph encoding only non-missing
subgraph = g_queen.subgraph(data_non_missing_index)

# Measure LISA
lisa = Moran_Local(np.array(data_non_missing, dtype=np.float64), subgraph.to_W())

# Get result in a form of a DataArray that can be aligned with da
psim = subgraph.generate_da(lisa.p_sim)
psim.plot()
```

It is probably not as fast as the original WSP due to some overhead of Graph but it works. 

As a side note, I think that the detection of no data value in the constructor in the `weights.raster` module needs updating as it only checks for `da.attrs["nodatavals"]` but the example above uses `da.attrs["_FillValue"]`. It is generally better to try getting it via `da.rio.nodata` but I'll open an issue for that. 

Another side note - the Raster_Awareness notebook is not executable as it uses files I have no idea where from and some custom branch of `splot`.

This also includes one line from #739 to ensure this actually works as intended.


martinfleis:
Oh, another note - I think that the tests of this and of W implementation pass only if `numba` is installed which we don't catch in CI here as there's no env where we have `xarray` but don't have `numba`.

jGaboardi (approved):

knaaptime (approved):

jGaboardi:
After @martinfleis resolves the conflicts, are we OK to merge or should we get one more approval?

Issue #208, opened 2019-12-03
Weight Object Question


achapkowski:
I was reading the help and I was trying to figure out what some of the properties mean on the weight object.

1. What does `s0`, `s1`, `s2` mean?  I see the formula, but what do those number represent?  The doc is lacking there.  Could someone please shed some light on these properties?

jGaboardi:
Also, the formulas aren't being rendered in the docs.

  * [pysal.lib.weights.W](https://pysal.org/pysal/generated/pysal.lib.weights.W.html#pysal.lib.weights.W)
  * [libpysal.weights.W](https://pysal.org/libpysal/generated/libpysal.weights.W.html#libpysal.weights.W)

sjsrey:
@achapkowski thanks for raising the issue. We will be fixing the rendering of the docs and will add more information about those measures. For now they are functions of the `W` matrix that are used by a number of tests for [spatial autocorrelation](https://github.com/pysal/esda/blob/master/esda/moran.py#L192) and econometric tests/estimation methods.

achapkowski:
@sjsrey thanks for the information, so it's a convenience property for other calculations?  

Here is some doc on math in rst files: http://svn.python.org/projects/doctools/trunk/doc/ext/math.rst

ljwolf:
@achapkowski Yes, they're fixed quantities useful to compute standard errors for spatial autocorrelation estimators and are there for convenience. 

If you look at where they are in the code (e.g. [Moran](https://github.com/pysal/esda/blob/master/esda/moran.py#L196), [Geary](https://github.com/pysal/esda/blob/master/esda/geary.py#L164), [Getis-Ord](https://github.com/pysal/esda/blob/master/esda/getisord.py#L148)), you can see they have something to do with [analytical expressions about the variance of random variables](https://en.wikipedia.org/wiki/Variance#Definition).

achapkowski:
@ljwolf thanks for the follow up.  


Is there some way to compare n number of weights (thinking of AIC or BIC) in pysal?  So let's say I make 100 weights matrices off of the same dataset adjusting the input parameters accordingly, say a distance parameter, how can I assume one of better over another? Is there a metric for this?  

That is why I started looking at the s0, s1, etc.. properties.

sjsrey:
> @ljwolf thanks for the follow up.
> 
[...]

PR #742, opened 2024-07-06
ENH: add Graph.summary with s0, s1, s2 and similar properties to Graph

martinfleis:
I started to look into support of Graph in esda and it seems that s0, s1 and s2 properties will be quite handy for that purpose.

Btw, what are those? 🙃


libpysal/graph/base.py
```diff
@@ -2090,6 +2090,61 @@ def describe(
...
+            s1
+        """
+        t = self.sparse.transpose()
+        t = t + self.sparse
+        t2 = t.multiply(t)  # element-wise square
+        return t2.sum() / 2.0
+
+    @cached_property
+    def s2(self):
+        r"""What is this???
```
martinfleis, line 2132: And here?

jGaboardi:
@martinfleis Here's some explanation of those --> https://github.com/pysal/libpysal/issues/208

jGaboardi (approved):

martinfleis:
Okay. I went ahead and used spdep's terminology using _S1 sum of weights_ and _S2 sum of weights_.

libpysal/graph/base.py
```diff
@@ -2090,6 +2090,85 @@ def describe(
...
+    def s0(self):
+        r"""Global sum of weights
+
+        ``s0`` is defined as
+
+        .. math::
+
+               s0=\sum_i \sum_j w_{i,j}
+
+        :attr:`s0`, :attr:`s1`, and :attr:`s2` reflect interaction between observation
```
jGaboardi, line 2103:
```suggestion
        :attr:`s0`, :attr:`s1`, and :attr:`s2` reflect interaction between observations
```

libpysal/graph/base.py
```diff
@@ -2090,6 +2090,85 @@ def describe(
...
+    def s1(self):
+        r"""S1 sum of weights
+
+        ``s1`` is defined as
+
+        .. math::
+
+               s1=1/2 \sum_i \sum_j \Big(w_{i,j} + w_{j,i}\Big)^2
+
+        :attr:`s0`, :attr:`s1`, and :attr:`s2` reflect interaction between observation
```
jGaboardi, line 2128:
```suggestion
        :attr:`s0`, :attr:`s1`, and :attr:`s2` reflect interaction between observations
```

libpysal/graph/base.py
```diff
@@ -2090,6 +2090,85 @@ def describe(
...
+    def s2(self):
+        r"""S2 sum of weights
+
+        ``s2`` is defined as
+
+        .. math::
+
+                s2=\sum_j \Big(\sum_i w_{i,j} + \sum_i w_{j,i}\Big)^2
+
+        :attr:`s0`, :attr:`s1`, and :attr:`s2` reflect interaction between observation
```
jGaboardi, line 2156:
```suggestion
        :attr:`s0`, :attr:`s1`, and :attr:`s2` reflect interaction between observations
```

jGaboardi (approved):

ljwolf:
I'm kind of wary implementing these traits directly on `Graph`, but am open to discussion? 

I am planning for a revision of ESDA would avoid using these (and instead use functions to compute the quantities on the fly), or we could use something like a `.summary()` method to compute and return these in a tabular format? 

I would really like to avoid reproducing these notation-specific things in the API; instead, we should try to use descriptive names and document their mathematical origin in the doctstring.

martinfleis:
@ljwolf I don't mind either way, this was just easier to do the transition to Graph downstream. These are now used in esda, spreg and spint and not including them requires more changes of the actual code when adding support for Graph. But having these as functions that are called when needed is also fine with me but that should probably also live somewhere in lib so it can just be consumed rather than copied. And then it is a question whether it is not easier to have them as properties from the beginning.

We can also potentially not include these in the rendered API and indicate in the docstring that they are intended for temporary private use only but that is a bit shady.

knaaptime:
what about just having them as private methods? I agree with Levi that it's better to avoid notation-specific conventions--and also with Martin that this is the easiest way to get Graph used in practice.

so s0,1,2 are used a lot for the different estimators, but do users ever actually need them? I doubt it (I don't think R exposes them, do they?). So what about including them here with a deprecation to get the ball rolling, then moving them to a private method available for spreg and esda?

martinfleis:
> I don't think R exposes them, do they?

They do. spdep has them among [constants](https://github.com/r-spatial/spdep/blob/a0a7d030ff836c396e80d61a30a26ae477d62433/man/spweights.constants.Rd#L26-L28). 

> So what about including them here with a deprecation to get the ball rolling, then moving them to a private method available for spreg and esda?

Fine with me. We could mirror this on W to ensure this transition to private property does not affect support of W.

knaaptime:
alternatively, if we keep them around as public, what's a better name?

martinfleis:
I have added a private properties and included deprecation on public ones. That we can have easier time during the transition period and can eventually remove public properties once we catch all usage downstream.

If we want to do the same on W, I'd do it as a follow-up.

jGaboardi (approved):

sjsrey:
> I have added a private properties and included deprecation on public ones. That we can have easier time during the transition period and can eventually remove public properties once we catch all usage downstream.
> 
> If we want to do the same on W, I'd do it as a follow-up.

I'm -1 on changing anything in W (beyond bug fixes). New functionality should be in the graph - if graph enables things that W can't do, then this is more incentive for folks to adopt the graph. 

s0, s1, s2, are central to a lot of  lattice data statistics and while I'm all for using meaningful variable names, in these cases I think adding new names could actually increase confusion.

How would the proposed naming changes affect the [moment calculations](https://github.com/pysal/esda/blob/main/esda/moran.py#L215-L235) here for example?

martinfleis:
I have now updated this as per our discussion from yesterday. This PR now introduces a new method `Graph.summary()` that returns a `GraphSummary` class which has a nice repr and contains all summary statistics as properties.

To get S0 and like in downstream, you do `Graph.summary().s0`.

Example:

```py
In [1]: import geopandas as gpd
   ...: from geodatasets import get_path
   ...: from libpysal import graph, weights
   ...: 
   ...: nybb = gpd.read_file(get_path("nybb")).set_index("BoroName")
   ...: contiguity = graph.Graph.build_contiguity(nybb)

In [2]: contiguity
Out[2]: 
<Graph of 5 nodes and 10 nonzero edges indexed by
 ['Staten Island', 'Queens', 'Brooklyn', 'Manhattan', 'Bronx']>

In [3]: contiguity.summary()
Out[3]: 
Graph Summary Statistics
========================
Graph indexed by:
 ['Staten Island', 'Queens', 'Brooklyn', 'Manhattan', 'Bronx']
==============================================================
Number of nodes:                                             5
Number of edges:                                            10
Number of connected components:                              2
Number of isolates:                                          1
Number of non-zero edges:                                   10
Percentage of non-zero edges:                           44.00%
Number of asymmetries:                                       0
--------------------------------------------------------------
Cardinalities
==============================================================
Mean:                       2    25%:                        2
Standard deviation:         1    50%:                        2
Min:                        0    75%:                        3
Max:                        3
--------------------------------------------------------------
Weights
==============================================================
Mean:                       1    25%:                        1
Standard deviation:         0    50%:                        1
Min:                        0    75%:                        1
Max:                        1
--------------------------------------------------------------
Sum of weights
==============================================================
S0:                                                         10
S1:                                                         20
S2:                                                        104
--------------------------------------------------------------
Traces
==============================================================
GG:                                                         10
G'G:                                                        10
G'G + GG:                                                   20
```

If you are in a notebook, it uses HTML repr which looks like this in JLab:

Screenshot 2024-07-12 at 13 15 56

Before doing docs and tests for this, can you let me know if this is the right direction? Especially @sjsrey and @ljwolf.

knaaptime:
sweet

sjsrey (approved):

sjsrey (approved):

martinfleis:
It is not ready guys. There are no tests and not documentation. In progress!

martinfleis:
Now it is ready!

PR #743, opened 2024-07-08
BUG: fix handling of non-isolate self-weights of 0 and order preservation

martinfleis:
Work over in esda uncovered a series of bugs related to handling of self-weights with 0 weight which are **not** isolates at the same time. I hope that this fixes them all. They are all tested, either in their own or in related tests. Reverting each would cause CI to fail.

This is precisely the reason why I started doing the support of Graph in esda, in hope that it verifies that Graph works as intended.

This is required for https://github.com/pysal/esda/pull/293/ to pass, alongside #742.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -1473,13 +1473,7 @@ class Graph(SetOpsMixin):
         pandas.Index
             Index with a subset of observations that do not have any neighbor
         """
-        nulls = self._adjacency[self._adjacency == 0]
-        # since not all zeros are necessarily isolates, do the focal == neighbor check
-        return (
-            nulls[nulls.index.codes[0] == nulls.index.codes[1]]
-            .index.get_level_values(0)
-            .unique()
-        )
+        return self.cardinalities.index[self.cardinalities == 0]
 
     @cached_property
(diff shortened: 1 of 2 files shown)


jGaboardi (approved):

libpysal/graph/base.py
```diff
@@ -264,7 +264,12 @@ def to_W(self):  # noqa: N802
                 neighbors[ix] = chunk.index.get_level_values("neighbor").tolist()
                 weights[ix] = chunk.tolist()
 
-        return W(neighbors=neighbors, weights=weights, id_order=self.unique_ids)
+        return W(
+            neighbors=neighbors,
+            weights=weights,
+            id_order=self.unique_ids.tolist(),
+            silence_warnings=True,
```
martinfleis, line 271: Warnings are silenced as the connectivity is already created and there's no need to warn about components and isolates when converting it to a different representation.

martinfleis:
I have pushed one more fix ensuring that all outcomes of groupby over adjacency are in the correct order, not lexicographically sorted.

jGaboardi (approved):

sjsrey (approved):

PR #745, opened 2024-07-09
added examples to graph

Kryndlea:
added examples to docstrings

PR #746, opened 2024-07-09
BUG: fix categorical lag for custom index

martinfleis:
While doing #745 we have noticed that categorical lag works only for graphs based on default range index.

This fixes it.

### DIFF
--- a/libpysal/graph/tests/test_spatial_lag.py
+++ b/libpysal/graph/tests/test_spatial_lag.py
@@ -65,3 +65,9 @@ class TestLag:
         with pytest.raises(ValueError, match="There are 2 ties that must be broken"):
             self.yc[3] = "a"  # create ties
             _lag_spatial(self.gc, self.yc, categorical=True)
+
+    def test_categorical_custom_index(self):
+        expceted = np.array(["bar", "foo", "bar", "foo"])
+        np.testing.assert_array_equal(
+            expceted, self.g.lag(["foo", "bar", "foo", "foo"])
+        )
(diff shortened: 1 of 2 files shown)


libpysal/graph/tests/test_spatial_lag.py
```diff
@@ -65,3 +65,9 @@ def test_ties_raise(self):
         with pytest.raises(ValueError, match="There are 2 ties that must be broken"):
             self.yc[3] = "a"  # create ties
             _lag_spatial(self.gc, self.yc, categorical=True)
+
+    def test_categorical_custom_index(self):
+        expceted = np.array(["bar", "foo", "bar", "foo"])
```
jGaboardi, line 70:
```suggestion
        expected = np.array(["bar", "foo", "bar", "foo"])
```

libpysal/graph/tests/test_spatial_lag.py
```diff
@@ -65,3 +65,9 @@ def test_ties_raise(self):
...
+        np.testing.assert_array_equal(
+            expceted, self.g.lag(["foo", "bar", "foo", "foo"])
```
jGaboardi, line 72:
```suggestion
            expected, self.g.lag(["foo", "bar", "foo", "foo"])
```

jGaboardi:
@martinfleis Do we require BDFL approval for this PR?

cc @sjsrey

jGaboardi (approved):

martinfleis:
Probably not. I just tagged @sjsrey as he wrote the code I'm touching here.

sjsrey:
> While doing #745 we have noticed that categorical lag works only for graphs based on default range index.
> 
> This fixes it.

Nice catch.

It reminds me that we might want to surface the indexing options that the graph affords in the migration guide.

sjsrey (approved):

PR #753, opened 2024-07-16
make asymmetry computation in Graph summary optional

martinfleis:
For non-binary weights, asymmetry check can be pretty costly. If we compute that by default in `summary`, it comes with a massive overhead when we need just s0, s1, s2 in esda. Adding a keyword that disables it by default. That way the overhead of computing stuff we don't need in esda is relatively negligible (still some but that is probably okay).

### DIFF
--- a/libpysal/graph/_summary.py
+++ b/libpysal/graph/_summary.py
@@ -118,7 +118,7 @@ class GraphSummary:
     <Graph of 5 nodes and 10 nonzero edges indexed by
         ['Staten Island', 'Queens', 'Brooklyn', 'Manhattan', 'Bronx']>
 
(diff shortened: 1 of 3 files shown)


jGaboardi (commented):
Should the default behavior be `True` (at least for now) since that's what's happening before? And perhaps a notice that default will be `False` in the future? Or am I misunderstanding something?

martinfleis:
There's no before, this thing got merged yesterday :D

jGaboardi (approved):

jGaboardi:
I think good to merge.

knaaptime (approved):

PR #754, opened 2024-07-16
DOC: move Graph from experimental to stable API

martinfleis:
I am still not sure if this is not premature given the number of bugs I've found while doing the Graph support in esda but here it is. Once this is merged, `Graph` is considered stable.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -6,8 +6,24 @@
 libpysal API reference
 ======================
 
+Spatial Graph
+-------------
+
+Modern implementation of spatial graphs encoding spatial weights matrices.
+
+.. autosummary::
+   :toctree: generated/
+
+   libpysal.graph.Graph
+   libpysal.graph.GraphSummary
+   libpysal.graph.read_parquet
+   libpysal.graph.read_gal
+   libpysal.graph.read_gwt
+
 Spatial Weights
 ---------------
+
+Legacy implementation of spatial weights matrices.
 
 .. autosummary::
    :toctree: generated/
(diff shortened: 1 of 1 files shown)


jGaboardi (approved):
LGTM, but I think it may be prudent to get all steering council + @ljwolf & @sjsrey prior to merge.

jGaboardi:
xref #733

martinfleis:
> LGTM, but I think it may be prudent to get all steering council + @ljwolf & @sjsrey prior to merge.

Happy with that but I had a feeling that we also already agreed on this during the last dev call (though it is not mentioned in minutes).

jGaboardi:
> > LGTM, but I think it may be prudent to get all steering council + @ljwolf & @sjsrey prior to merge.
> 
> Happy with that but I had a feeling that we also already agreed on this during the last dev call (though it is not mentioned in minutes).

OK, since I missed half that call I'll take your word for it. Let's merge?

knaaptime:
not sure if serge wanted to add more to the migration guide before that's merged though

martinfleis:
> not sure if serge wanted to add more to the migration guide before that's merged though

I suppose that it does not matter much if the migration guide is finished before the next release.

jGaboardi:
Also, we should probably do at least 1 `v4.12.0-alpa` (or `rc`) release to make sure nothing breaks in the workflow, etc.

martinfleis:
I'll leave that to you but given the 4.11 was fairly recently, I don't expect any complications there.

PR #755, opened 2024-07-17
add graph from travel network

knaaptime:
```python
import geodatasets
import geopandas as gpd
from geosnap import io
from libpysal.graph import Graph

df = gpd.read_file(geodatasets.get_path("geoda cincinnati"))

net = io.get_network_from_gdf(df)

Graph.build_travel_distance(df.set_index('ID'), net, 500, kernel=None).adjacency
>>> focal     neighbor
726907.0  726907.0      0.000000
          709415.0    385.609009
          709472.0    309.471985
          738259.0    346.858002
          738277.0    247.363007
                         ...    
703708.0  703537.0    349.321014
          703574.0    190.669006
          703629.0    370.070007
          703686.0    447.006989
          703708.0      0.000000
Name: weight, Length: 8075, dtype: float64


Graph.build_travel_distance(df.set_index('ID'), net, 500, kernel='triangular').adjacency
>>> focal     neighbor
726907.0  726907.0    1.000000
          709415.0    0.228782
          709472.0    0.381056
          738259.0    0.306284
          738277.0    0.505274
                        ...   
703708.0  703537.0    0.301358
          703574.0    0.618662
          703629.0    0.259860
          703686.0    0.105986
          703708.0    1.000000
Name: weight, Length: 8075, dtype: float64


```

### DIFF
--- a/libpysal/graph/_network.py
+++ b/libpysal/graph/_network.py
@@ -36,7 +36,7 @@ def pdna_to_adj(origins, network, node_ids, threshold):
     node_ids:
         array of node_ids in the pandana.Network aligned with the input
         observations in ``origins``. This is created via a call like
-        `pandana.Network.get_node_ids(df.geometry.x, df.geometry.y)`
+        ``pandana.Network.get_node_ids(df.geometry.x, df.geometry.y)``
     threshold : int
         maximum travel distance (inclusive)
 
--- a/libpysal/graph/_network.py
+++ b/libpysal/graph/_network.py
@@ -6,6 +6,7 @@ from ._utils import _induce_cliques
 def _build_coplanarity_node_lookup(geoms):
     """
     Identify coplanar points and create a look-up table for the coplanar geometries.
+    Same function as in graph._utils, but need to keep the index to use as graph ids
     """
     # geoms = geoms.reset_index(drop=True)
     coplanar = []
@@ -72,13 +73,38 @@ def build_travel_graph(
     network,
     threshold,
 ):
+    """Compute the shortest path between gdf centroids via a pandana.Network
+    and return an adjacency list with weight=cost
+
+    Parameters
+    ----------
+    df : geopandas.GeoDataFrame
+        geodataframe of observations. CRS should be the same as the locations
+        of node_x and node_y in the pandana.Network (usually 4326 if network
+        comes from OSM, but sometimes projected to improve snapping quality).
+    network : pandana.Network
+        Network that encodes travel costs. See <https://udst.github.io/pandana/>
+    threshold : int
+        maximum travel cost to consider neighbors
+
+    Returns
+    -------
+    pandas.Series
+        adjacency formatted as multiindexed (focal, neighbor) series
+    """
     df = df.copy()
     df["node_ids"] = network.get_node_ids(
         df.geometry.centroid.x, df.geometry.centroid.y
     )
+    # depending on density of the graph nodes / observations, it is common to have
+    # multiple observations snapped to the same network node, so use the clique
+    # expansion logic to handle these cases
 
+    # get indices of multi-observations at unique nodes
     coplanar, nearest = _build_coplanarity_node_lookup(df["node_ids"])
+    # create adjacency on unique nodes
     adj = pdna_to_adj(df, network, threshold, reindex=True, drop_nonorigins=True)
+    # add clique members back to adjacency
     adj_cliques = _induce_cliques(
         adj.rename(
             columns={"origin": "focal", "destination": "neighbor", "cost": "weight"}
@@ -86,7 +112,7 @@ def build_travel_graph(
         coplanar=coplanar,
         nearest=nearest,
     )
-
+    # reorder, drop induced dupes, and return
     adj_cliques = (
         adj_cliques.groupby(["focal", "neighbor"])
         .first()
--- a/libpysal/graph/_network.py
+++ b/libpysal/graph/_network.py
@@ -6,7 +6,7 @@ from ._utils import _induce_cliques, _validate_geometry_input
 def _build_coplanarity_node_lookup(geoms):
     """
     Identify coplanar points and create a look-up table for the coplanar geometries.
-    Same function as in graph._utils, but need to keep the index to use as graph ids
+    Same function as in ``graph._utils``, but need to keep the index to use as graph ids
     """
     # geoms = geoms.reset_index(drop=True)
     coplanar = []
--- a/libpysal/graph/_network.py
+++ b/libpysal/graph/_network.py
@@ -35,7 +35,7 @@ def pdna_to_adj(origins, network, node_ids, threshold):
         pandana.Network instance that stores the local travel network
     node_ids:
         array of node_ids in the pandana.Network aligned with the input
-        observations in `origins`. This is created via a call like
+        observations in ``origins``. This is created via a call like
         `pandana.Network.get_node_ids(df.geometry.x, df.geometry.y)`
     threshold : int
         maximum travel distance (inclusive)
--- a/libpysal/graph/_network.py
+++ b/libpysal/graph/_network.py
@@ -1,6 +1,6 @@
 import numpy as np
-
-from ._utils import _induce_cliques
+import shapely
+from ._utils import _induce_cliques, _validate_geometry_input
 
 
 def _build_coplanarity_node_lookup(geoms):
@@ -23,35 +23,30 @@ def _build_coplanarity_node_lookup(geoms):
     return np.asarray(coplanar), np.asarray(nearest)
 
 
-def pdna_to_adj(origins, network, threshold, reindex=True, drop_nonorigins=True):
+def pdna_to_adj(origins, network, node_ids, threshold):
     """Create an adjacency list of shortest network-based travel between
        origins and destinations in a pandana.Network.
 
     Parameters
     ----------
     origins : geopandas.GeoDataFrame
-        Geodataframe of origin geometries to begin routing. If geometries are
-        polygons, they will be collapsed to centroids
+        Geodataframe of origin geometries to begin routing.
     network : pandana.Network
         pandana.Network instance that stores the local travel network
     threshold : int
         maximum travel distance (inclusive)
-    reindex : bool, optional
-        if True, use geodataframe index to identify observations in the
-        adjlist. If False, the node_id from the OSM node nearest each
-        observation will be used. by default True
-    drop_nonorigins : bool, optional
-        If True, drop any destination nodes that are not also origins,
-        by default True
 
     Returns
     -------
     pandas.DataFrame
         adjacency list with columns 'origin', 'destination', and 'cost'
     """
-    node_ids = network.get_node_ids(origins.centroid.x, origins.centroid.y).astype(int)
 
     # map node ids in the network to index in the gdf
+    _validate_geometry_input(
+        origins.geometry, ids=None, valid_geometry_types="Point"
+    )
+
     mapper = dict(zip(node_ids, origins.index.values, strict=False))
 
     namer = {"source": "origin", network.impedance_names[0]: "cost"}
@@ -59,11 +54,9 @@ def pdna_to_adj(origins, network, threshold, reindex=True, drop_nonorigins=True)
     adj = network.nodes_in_range(node_ids, threshold)
     adj = adj.rename(columns=namer)
     # swap osm ids for gdf index
-    if reindex:
-        adj = adj.set_index("destination").rename(index=mapper).reset_index()
-        adj = adj.set_index("origin").rename(index=mapper).reset_index()
-    if drop_nonorigins:
-        adj = adj[adj.destination.isin(origins.index.values)]
+    adj = adj.set_index("destination").rename(index=mapper).reset_index()
+    adj = adj.set_index("origin").rename(index=mapper).reset_index()
+    adj = adj[adj.destination.isin(origins.index.values)]
 
     return adj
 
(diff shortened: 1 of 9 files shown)


knaaptime:
i think that windows failure is [an issue in pandana's c lib](https://github.com/ansys/pymapdl/issues/14#issuecomment-364654285), if I had to guess

knaaptime:
probably better to call this build_travel_cost, since impedance isn't always distance

libpysal/graph/_network.py
```diff
@@ -0,0 +1,124 @@
...
+        maximum travel cost to consider neighbors
+
+    Returns
+    -------
+    pandas.Series
+        adjacency formatted as multiindexed (focal, neighbor) series
+    """
+    df = df.copy()
+    df["node_ids"] = network.get_node_ids(
+        df.geometry.centroid.x, df.geometry.centroid.y
```
martinfleis, line 97: Automatically falling to centroid goes against what we have done elsewhere (e.g. knn). If the builder requires point geometry, we raise if that is not fulfilled and it is up to user to determine the optimal conversion to points. I'd prefer we use the same logic here.

libpysal/graph/_network.py
```diff
@@ -0,0 +1,124 @@
...
+    drop_nonorigins : bool, optional
+        If True, drop any destination nodes that are not also origins,
+        by default True
+
+    Returns
+    -------
+    pandas.DataFrame
+        adjacency list with columns 'origin', 'destination', and 'cost'
+    """
+    node_ids = network.get_node_ids(origins.centroid.x, origins.centroid.y).astype(int)
```
martinfleis, line 52: Don't we already have these from the first call to `get_node_ids` below in `build_travel_graph`? We could just reuse it to avoid another call here.

libpysal/graph/_network.py
```diff
@@ -0,0 +1,124 @@
...
+        pandana.Network instance that stores the local travel network
+    threshold : int
+        maximum travel distance (inclusive)
+    reindex : bool, optional
+        if True, use geodataframe index to identify observations in the
+        adjlist. If False, the node_id from the OSM node nearest each
+        observation will be used. by default True
+    drop_nonorigins : bool, optional
+        If True, drop any destination nodes that are not also origins,
+        by default True
```
martinfleis, line 45: Why do we have these two options if they are not exposed in the API? Can we just drop it and remove the if/else clauses?

libpysal/graph/base.py
```diff
@@ -1454,6 +1455,50 @@ def build_h3(cls, ids, order=1, weight="distance"):
...
+        kernel : str or callable, optional
+            kernel transformation applied to the weights. See
+            libpysal.graph.Graph.build_kernel for more information on kernel
+            transformation options. Default is None, in which case the Graph weight
+            is pure distance between focal and neighbor
+
+        Returns
+        -------
+        Graph
+        """
```
martinfleis, line 1487: Can we add a simple example here, that includes building the pandana.Network? It can be quite opaque how to use this.

martinfleis (commented):
Cool!

libpysal/graph/base.py
knaaptime: yeah, i wasnt sure how best to handle that. The fastest way to get a network is [geosnap's getter](), though I didn't want to depend too heavily on geosnap. That function itself is pretty small, so it could just be defined in the example, but that still gets a little verbose. You can also use [osmnet](https://udst.github.io/osmnet/), but its much slower. (then obviously you can download an example from our bucket, but that doesn't have any pedagogical value for *creating* one, and those networks are quite old now

libpysal/graph/_network.py
knaaptime: yep, that occurred to me a few mins ago too. I think i can call the geomvalidator before here?

libpysal/graph/_network.py
knaaptime: yes, ported this directly from  geosnap where I need those options

knaaptime:
this linter check is the bane of my existence

jGaboardi:
> this linter check is the bane of my existence

Do you have pre-commit set up locally?

sjsrey:
> this linter check is the bane of my existence

I'm wondering if there is a way to set all of the lint checking locally, across all the projects without having to do it project by project? I'm thinking of a global setting somewhere?

knaaptime:
the local pre commit gives me the willies. Ive gotten lost in rebasemergefastrorewardrewind too many times to like git making changes on my behalf

jGaboardi:
> the local pre commit gives me the willies. Ive gotten lost in rebasemergefastrorewardrewind too many times to like git making changes on my behalf

The way we set up `pre-commit` formatting is automatically changed, but not linting changes

jGaboardi:
Your commit will fail and report linting errors if any linting violations are found.

knaaptime:
yeah but sometimes i need to get the idea saved in git history before its looking pretty and I don't want my commit to fail (sorry formatting/linting the same in my head)

jGaboardi:
in that case you can do `git commit -m "..." --no-verify`

libpysal/graph/base.py
knaaptime: what do you think of the example in the PR request?

knaaptime:
but if just avoid setting it up entirely, I don't have to type those excruciatingly long 11 characters :P

libpysal/graph/base.py
martinfleis: What I think we should show is how you build the network that can be used here. Reading some random hdf5 doesn't cut it. 

We could potentially use the data shipped with momepy as it contains street network and buildings of the same area.

knaaptime: im cool with that. though i meant [the example up here](https://github.com/pysal/libpysal/pull/755#issue-2414792747) that builds the network on the fly and is only three lines long, but requires geosnap

martinfleis: mmm, looking at it again it is indeed not super straightforward. But this works

```py
import momepy
import geopandas as gpd
import pandana

path = momepy.datasets.get_path("bubenec")
streets = gpd.read_file(path, layer="streets")
buildings = gpd.read_file(path, layer="buildings")

nodes, edges = momepy.nx_to_gdf(momepy.gdf_to_nx(streets))

network = pandana.Network(
    nodes.geometry.x,
    nodes.geometry.y,
    edge_from=edges["node_start"],
    edge_to=edges["node_end"],
    edge_weights=edges[["mm_len"]],
)

Graph.build_travel_cost(buildings.centroid, network, 200)
```

knaaptime: cool with me. though as a user, its probably not how i'd get a network for my own purposes. What about the simple getter [like here](https://knaaptime.com/urban_analysis/03_esda/spatial_graphs.html#transport-network-based-graphs) that just uses osmnx?

martinfleis: Happy with the one you just added! 

Btw, speaking as a user, the code above is exactly how I would do that. I pulled that straight from my own 3-years old research notebook :).

libpysal/graph/base.py
```diff
@@ -1484,6 +1484,42 @@ def build_travel_cost(cls, df, network, threshold, kernel=None):
...
         -------
         Graph
+
+        Examples
+        ---------
+        >>> import geodatasets
+        >>> import geopandas as gpd
+        >>> import osmnx as ox
+        >>> import pandana as pdna
+        >>> from libpysal.graph import Graph
```
martinfleis, line 1494:
```suggestion
```

We've been omitting this within the class docstrings.

libpysal/graph/base.py
```diff
@@ -1484,6 +1484,42 @@ def build_travel_cost(cls, df, network, threshold, kernel=None):
...
+
+        >>> # read an example geodataframe
+        >>> df = gpd.read_file(geodatasets.get_path("geoda Cincinnati")).to_crs(4326)
```
martinfleis, line 1497: Given the way this will get rendered, it may be easier to keep the comments out of the code blocks.

```suggestion
        Read an example geodataframe:

        >>> df = gpd.read_file(geodatasets.get_path("geoda Cincinnati")).to_crs(4326)
```

This way you get text-code_block-text-code_block sequence in the docs. Otherwise you just get code_block-code_block which is not as friendly.

libpysal/graph/base.py
```diff
@@ -1484,6 +1484,42 @@ def build_travel_cost(cls, df, network, threshold, kernel=None):
...
+
+        >>> # download a walk network using osmnx
+        >>> osm_graph = ox.graph_from_polygon(df.unary_union, network_type="walk")
```
martinfleis, line 1500:
```suggestion
        >>> osm_graph = ox.graph_from_polygon(df.union_all(), network_type="walk")
```

unary_union is deprecated

libpysal/graph/base.py
knaaptime: yeah, fair, i didnt mean your example not how *a user* might do it, but how a *really novice* user might do it. You already have the street data on hand and know that length is an actual impedance, not the straight-line distance between nodes or something. I figure most people are starting from scratch

libpysal/graph/base.py
```diff
@@ -1484,6 +1484,42 @@ def build_travel_cost(cls, df, network, threshold, kernel=None):
...
+        >>>     edge_from=edges["u"],
+        >>>     edge_to=edges["v"],
+        >>>     edge_weights=edges[["length"]],
+        >>>     node_x=nodes["x"],
+        >>>     node_y=nodes["y"],)
+
+        >>> # use the pandana network to compute shortest paths between gdf centroids
+        >>> G = Graph.build_travel_cost(df.set_geometry(df.centroid), network, 500)
+        >>> G.adjacency.head()
+
```
martinfleis, line 1515:
```suggestion
```

libpysal/graph/base.py
knaaptime: thanks i never know how to do this

libpysal/graph/base.py
knaaptime: 😬 keep forgetting that! lots of things I need to address

martinfleis (approved):

jGaboardi:
Having `pandana` as a hard dependency is a touch troublesome, no? For both Python 3.11 and Python 3.12 it took a considerably long time to get a compatible Conda Forge release, similar to `numba`. Maybe better to have as an optional dependency?

Also, it needs to be added to the `pyproject.toml` and `environment.yml`

libpysal/graph/_network.py
```diff
@@ -0,0 +1,119 @@
...
+        maximum travel distance (inclusive)
+
+    Returns
+    -------
+    pandas.DataFrame
+        adjacency list with columns 'origin', 'destination', and 'cost'
+    """
+
+    # map node ids in the network to index in the gdf
+
```
jGaboardi, line 50:
```suggestion
```

libpysal/graph/_network.py
```diff
@@ -0,0 +1,119 @@
...
+
+def build_travel_graph(
+    df,
+    network,
+    threshold,
+):
+    """Compute the shortest path between gdf centroids via a pandana.Network
+    and return an adjacency list with weight=cost. Note unlike distance_band,
+    G_ij and G_ji are often different because travel networks may be directed.
```
jGaboardi, line 72:
```suggestion
    :math:`G_ij` and :math:`G_ji` are often different because travel networks may be directed.
```

knaaptime:
it just consumes pandana--can only operate on a network if it's given, so no dependency (like xarray). so only changes are to the ci test files

libpysal/graph/_network.py
```diff
@@ -0,0 +1,119 @@
+import numpy as np
+
+from ._utils import _induce_cliques, _validate_geometry_input
+
+
+def _build_coplanarity_node_lookup(geoms):
+    """
+    Identify coplanar points and create a look-up table for the coplanar geometries.
+    Same function as in graph._utils, but need to keep the index to use as graph ids
```
jGaboardi, line 9:
```suggestion
    Same function as in ``graph._utils``, but need to keep the index to use as graph ids
```

libpysal/graph/_network.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Parameters
+    ----------
+    origins : geopandas.GeoDataFrame
+        Geodataframe of origin geometries to begin routing.
+    network : pandana.Network
+        pandana.Network instance that stores the local travel network
+    node_ids:
+        array of node_ids in the pandana.Network aligned with the input
+        observations in `origins`. This is created via a call like
```
jGaboardi, line 38:
```suggestion
        observations in ``origins``. This is created via a call like
```

libpysal/graph/_network.py
```diff
@@ -0,0 +1,119 @@
...
+        `pandana.Network.get_node_ids(df.geometry.x, df.geometry.y)`
```
jGaboardi, line 39:
```suggestion
        ``pandana.Network.get_node_ids(df.geometry.x, df.geometry.y)``
```

libpysal/graph/_network.py
```diff
@@ -0,0 +1,119 @@
...
+):
+    """Compute the shortest path between gdf centroids via a pandana.Network
+    and return an adjacency list with weight=cost. Note unlike distance_band,
+    G_ij and G_ji are often different because travel networks may be directed.
+
+    Parameters
+    ----------
+    df : geopandas.GeoDataFrame
+        geodataframe of observations. CRS should be the same as the locations
+        of node_x and node_y in the pandana.Network (usually 4326 if network
```
jGaboardi, line 78:
```suggestion
        of ``node_x`` and ``node_y`` in the pandana.Network (usually 4326 if network
```

jGaboardi (approved):

sjsrey (approved):
Great addition!

jGaboardi:
> this one was your fault james!!! :P

LOL It was on purpose.

PR #761, opened 2024-08-06
Add exponential kernel to Graph

knaaptime:


martinfleis (approved):

libpysal/graph/_kernel.py
```diff
@@ -106,6 +112,7 @@ def _kernel(
             - gaussian:
             - bisquare:
             - cosine:
+            - exponential:
```
knaaptime, line 115: should these all probably have formulas here?

martinfleis: Ideally

knaaptime:
no way to bound it at the bandwidth if you want though 🤷‍♂️. would require another argument

martinfleis:
Thinking about this more, I am actually not sure if I like passing alpha as `bandwidth`. It is inconsistent with the rest of the kernels then. Also, distance band has both bandwidth and alpha arguments (though not used together), so we shall aim for some degree of consistency between these.

PR #760, opened 2024-07-31
Fix for Graph.describe() when the graph has a string index (#759)

u3ks:
This PR fixes #759 .

Pandas aggregation functions pass the original string index to numba, and thats why it fails the type check.
To fix this we can either have two new numba mode calculation functions - one for  strings, one for numbers or we can explicitly recast the index to integers, carry out the calculations using the numeric index and then reindex the results before returning them.

PR #765, opened 2024-08-29
ENH: ensure lag_spatial is compatible with both W and Graph

martinfleis:
Ensures that when downstream packages (like splot https://github.com/pysal/splot/pull/184) use `lag_spatial` when a Graph is given, the actual computation produces correct lag. This will avoid silent introduction of bugs when using Graph as scipy does not raise but produces different output.

### DIFF
--- a/libpysal/weights/spatial_lag.py
+++ b/libpysal/weights/spatial_lag.py
@@ -85,7 +85,7 @@ def lag_spatial(w, y):
            [6.        , 2.        ],
            [6.        , 2.        ]])
     """
-    return w.sparse * y
+    return w.sparse @ y
 
 
 def lag_categorical(w, y, ties="tryself"):


sjsrey (approved):

jGaboardi (approved):

knaaptime:
i think i would support doing a bugfix release with this change

martinfleis:
Yeah, we could do one with this and #760 but we'll need to backport those commits. Do you want me to do that?

knaaptime:
im game

knaaptime:
😬 sorry. i vote we do it

martinfleis:
The release notes action didn't get what I was doing but the patch release is out https://github.com/pysal/libpysal/releases/tag/v4.12.1

PR #768, opened 2024-10-21
COMPAT: h3 v4 compatibility

martinfleis:
Requires https://github.com/pysal/tobler/pull/218 to work, so likely only dev and oldest will pass until Tobler gets a patch release.


knaaptime:
i can get the tobler release out today. Was gonna see if I can fix that issue with the docs along the way but can just do the patch if not

jGaboardi (approved):

PR #780, opened 2025-03-26
BUG: ensure validity of polygons in voronoi_frames

martinfleis:
On some of the locations where `coverage_union` failed, unary union does not fail but produces invalid polygons. If that happens, the subsequent intersection has a tendency to fail. This seems to fix it but I wasn't able to extract a sample small enough to be included in CI. So my proposal is not merge as is.

### DIFF
--- a/libpysal/cg/voronoi.py
+++ b/libpysal/cg/voronoi.py
@@ -463,6 +463,11 @@ def voronoi_frames(
     else:
         polygons = polygons.iloc[ids_polygons].reset_index(drop=True)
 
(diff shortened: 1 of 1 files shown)


martinfleis:
Not that easy, the result is wrong...

martinfleis:
The result is wrong by a coincidence and the wrongness is caused by https://github.com/libgeos/geos/issues/955. This is fine as is.

jGaboardi (approved):

jGaboardi:
Nice work!

PR #793, opened 2025-06-09
PERF: eliminate redundant sorting

martinfleis:
Within `gwlearn` bandwidth search, we generate quite large graphs and it turned out that the bottleneck is not model fitting but graph creation. Especially sorting and reindex we do, and which we do at multiple spots even though one shall be enough. Removing some obvious redundancies here but there's a likelihood that some other performance-focused PRs will follow.

### DIFF
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -32,7 +32,6 @@ def _sparse_to_arrays(sparray, ids=None, resolve_isolates=True, return_adjacency
     When we know we are dealing with cliques, we don't want to resolve
     isolates here but will do that later once cliques are induced.
     """
-    argsort_kwds = {"stable": True} if NUMPY_GE_2 else {}
     sparray = sparray.tocoo(copy=False)
     if ids is not None:
         ids = np.asarray(ids)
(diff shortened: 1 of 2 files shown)


libpysal/graph/_utils.py
```diff
@@ -42,15 +41,13 @@ def _sparse_to_arrays(sparray, ids=None, resolve_isolates=True, return_adjacency
...
+        tail = ids[sparray.col]
+        data = sparray.data
     else:
-        sorter = sparray.row.argsort(**argsort_kwds)
-        head = sparray.row[sorter]
-        tail = sparray.col[sorter]
-        data = sparray.data[sorter]
+        head = sparray.row
+        tail = sparray.col
+        data = sparray.data
```
martinfleis, line 50: Unnecessary as `_resolve_islands` does this. And if it does not go through there, it is done in clique resolution.

libpysal/graph/base.py
```diff
@@ -1085,7 +1085,7 @@ def build_kernel(
             coplanar=coplanar,
         )
 
-        return cls.from_arrays(head, tail, weight)
+        return cls.from_arrays(head, tail, weight, is_sorted=True)
```
martinfleis, line 1088: As mentioned above, output of `_kernel` is known to be sorted.

jGaboardi (approved):

PR #797, opened 2025-07-18
add benchmarks of Graph

martinfleis:
Starting with some simple benchmarks of Graph. My interest is to compare performance on main vs in a PR, not necessarily to compare it with W. Hence the assumption is that we will be comparing different JSON outputs. 

Given I need to disable caching in base.py, let's not merge this.

Any suggestions what to add? Something that could be sensitive to changes of the way we represent adjacency, since I am mostly interested in removing the overheads of reindexing we have now.


jGaboardi:
possible/beneficial to add in memory profiling?

knaaptime:
i think we need to test lags and such. The graph ordering is critical to the sparse representation/alignment, and we need to maintain the integrity of that for all the linear algebra stuff. i.e. if we want to test indexing stuff, we need to test not only creation of sparse, but the math that goes through it

knaaptime:
i guess what im getting at is, in my gut, i don't think we're going to be able to avoid the performance overhead from sorting the graph (sorting for sparse will be unavoidable, but maybe we'll have a faster way than a pandas reindex or something). The question is *when* the hit occurs (when the graph is instantiated, when the sparse attribute is called, etc) so might be nice to compare different options

knaaptime:
cool, i trust your lead then. I was imagining a scenario where we considered things like reimplementing lag or transform using groupby/apply instead of matrix operation which would have implications for when you need to build sparse  (i.e. try to skip re-ordering by avoiding sparse whenever possible), then what tasks/packages that burden falls hardest onto

martinfleis:
Ah, yeah. If that happens we should certainly include these in the benchmarks.

martinfleis:
I have updated this to use `asv` for benchmarks. Which also means that this is not mergeable.

@jGaboardi in the same way, `asv` can do memory profiling if we want to do it.

jGaboardi:
> Given I need to disable caching in base.py, let's not merge this.

> Which also means that this is not mergeable.

So we can keep this PR open in the docket and simply trigger a re-run of it after other PRs are merged to compare vs. `main`. Am I understanding correctly?

martinfleis:
Nope, those comments are now obsolete. This is completely mergeable now, following the same model as geopandas does with benchmarks (which do not work at the moment but that is not relevant here).

ljwolf:
I'm attempting to add some basic common operations on the resulting `graph.sparse` matrix that are used downstream, like `numpy.linalg.cholesky()`, `scipy.sparse.linalg.spsolve()`, `scipy.sparse.linalg.eigs()`, which are affected by the construction, sorting, and `elimitate_zeros()` decisions we make. I've done this upstream (to martinfleis/libpysal#41) but let me know if I should just push here.

PR #802, opened 2025-08-22
BUG: RNG pruning condition and loop bound

### LINKED ISSUE #801: Bug in the Delaunay prunning (RNG-2)
Hi!

I've recently started using libpysal for a project and while playing with the graphs from `gabriel.py` I realized that `Relative_Neighborhood` and `Delaunay` produce the same results. I looked into it and I think there is a problem with the condition for pruning the Delaunay edges: according to the paper listed in the doc ([Toussaint (1980), RNG-2](https://cgm.cs.mcgill.ca/~godfried/publications/relng.pdf)), the condition is equivalent to: 
- prune $(i,j)$ whenever there is a $k$ with $\max(d_{ik},d_{jk}) < d_{ij}$

I gave it a try myself here: https://github.com/dcodrut/libpysal/commit/16205dbe3fcca1cffa87d1cefdaef7a6fcf25411
Additionally, I included the last node in the for loop as `edges.max()` was returning n - 1: https://github.com/dcodrut/libpysal/commit/d39206309c58bb295df6df14adcca170f050b7b3

Let me know what you think, if it makes sense, I can open a PR. 

Thanks for the work behind the library!

dcodrut:
This PR fixes #801:

* Fix the Toussaint RNG-2 condition: prune $(i,j)$ if ∃ $k$: $\max(d_{ik}, d_{jk}) < d_{ij}$
* include last node: `n = len(coordinates)` and skip `k == i or k == j`
* Updated the test (the resulted graph in shown below)

image

### DIFF
--- a/libpysal/weights/gabriel.py
+++ b/libpysal/weights/gabriel.py
@@ -334,7 +334,7 @@ def _filter_relativehood(edges, coordinates, return_dkmax=False):
     3. for each edge of the delaunay (i,j), prune
        if any dkmax is smaller than d(i,j)
     """
-    n = edges.max()
+    n = len(coordinates)
     out = []
     r = []
     for edge in edges:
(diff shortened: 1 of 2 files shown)


jGaboardi (approved):
Great catch on the @dcodrut ! Thanks for the PR

ljwolf:
Thanks very much @dcodrut!

PR #815, opened 2025-12-11
ENH: add information about components and isolates to Graph.__repr__

### LINKED ISSUE #807: Connectivity warnings on Graph
The `W` builder warns on init if there are multiple components and islands present. This can be silenced by a keyword but it enabled by default. On Graph, we have nothing like this and we needs to check this ourselves.

I was debugging a weird failure of `esda.Join_Counts_Local` and realised it is caused by multiple components only when I tried passing the graph as W, at which point I got the warning and immediately understood the issue.

I am wondering whether we should do something about this on Graph as well. One option is to mimic the behaviour of `W` with the warning, other less intrusive option might be to expand the `__repr__` to include this information (as `Graph.summary()` does). I sort of came to a conclusion yesterday that it might be useful to avoid unforeseen issues.

martinfleis:
Closes #807

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -69,7 +69,6 @@ jobs:
 
           geodatasets.fetch("nybb")
           geodatasets.fetch("geoda liquor_stores")
-          geodatasets.fetch("eea large_rivers")
           geodatasets.fetch("geoda groceries")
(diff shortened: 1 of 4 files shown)


martinfleis:
No - https://github.com/pysal/libpysal/actions/runs/20118036076/job/57732032807

PR #824, opened 2026-01-06
REGR: Revert #793 (sorting in sparse to arrays)

martinfleis:
Reverting the rest of #793 as #822 apparently was not enough. I managed to cause two regressions by a single PR :tada:.

This was caught by momepy CI - https://github.com/pysal/momepy/actions/runs/20742499625/job/59552088014


martinfleis:
The issue in CI is linked to numpy 1.26, which is already out of SPEC 0 support window. Shall we just bump the pin?

jGaboardi (approved):

sjsrey:
I think tracking spec 0 is important. Trying to support outside of that window increases the maintenance cost.

PR #834, opened 2026-01-12
Improve `graph._kernely.py` performance 

Jiya873:
I have updated `_kernel.py` by using `_distance_band` instead of full pairwise distance matrices when a numeric `taper` is provided. This reduces memory usage on larger datasets.

### DIFF
--- a/libpysal/graph/_kernel.py
+++ b/libpysal/graph/_kernel.py
@@ -141,10 +141,20 @@ def _kernel(
             rows, cols = numpy.where(mask)
             values = coordinates[mask]
             d = sparse.csc_array((values, (rows, cols)), shape=coordinates.shape)
-    elif (type(taper) in (float, int)) and (metric == "euclidean"):
-        d = _distance_band(coordinates, taper)
+    elif isinstance(taper, int | float) and (metric == "euclidean"):
+        if taper is False:
+            threshold = numpy.inf
+        else:
+            threshold = bandwidth if taper is True else taper
+            if threshold is None:
+                threshold = numpy.inf
+        d = _distance_band(coordinates, threshold)
         if exclude_self_weights:
-            d = d - sparse.diags_array(d.diagonal())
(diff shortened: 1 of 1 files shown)


libpysal/graph/_kernel.py
```diff
@@ -141,6 +141,14 @@ def _kernel(
             rows, cols = numpy.where(mask)
             values = coordinates[mask]
             d = sparse.csc_array((values, (rows, cols)), shape=coordinates.shape)
+    elif (
+        isinstance(taper, (float, int))
+        and not isinstance(taper, bool)
+        and (metric in ("euclidean", "manhattan", "cityblock", "minkowski"))
```
martinfleis:
```suggestion
        and (metric == "euclidean")
```

Second `isinstance` check is unnecessary after the first one and `distance_band` works on euclidean distance only.

libpysal/graph/_kernel.py
```diff
@@ -141,6 +141,14 @@ def _kernel(
...
+    ):
+        d = _distance_band(coordinates, taper)
+        if exclude_self_weights:
+            d = d - sparse.diags(d.diagonal())
```
martinfleis: `sparse.diags` return sparse matrix. Use `diags_array` to return sparse array. Read the documentation, it is written in there.

martinfleis (commented):
Do you have some performance comparison of before and after?

libpysal/graph/_kernel.py
Jiya873: i missed that about `distance_band`. thanks for that and i have updated the code now.

libpysal/graph/_kernel.py
Jiya873: ohh... yess! i have read the documentation now and understood the concept.

Jiya873:
while doing changes, I encountered an issue in the kernel function.
Would you consider this a bug?
Screenshot 2026-01-22 234813

PR #835, opened 2026-01-12
fix: handle distance arrays when calculating bandwidth

Jiya873:
I added a check for empty distance arrays to avoid numpy warnings and ensure stable behavior when no neighbors are found.


martinfleis (commented):
Can you add a test for this? Something that would explicitly fail when the warning is raised. Or if current tests emit this, you can just add `with pytest.warns` around that (preferable).

PR #841, opened 2026-01-17
Allow Polygons in Voronoi and pass kwargs to voronoi_frames (Fix #688)

### LINKED ISSUE #688: Support all geom types in Voronoi Graph builder
With #678 in, I suppose we can allow Voronoi weights builder for any geometry type. We just need a documentation and pass the args to `voronoi_frames`.

Draavith-DNA:
- **Geometry Support:** Updated `_VALID_GEOMETRY_TYPES` to allow `Polygon` and `MultiPolygon` inputs.
- **Arg Passing:** Ensured `**kwargs` are passed through the `_voronoi` function to the `voronoi_frames` engine, allowing parameters like `shrink` or `buffer` to work correctly.
- **Refactoring:** Removed the `@_validate_coplanar` decorator for the Voronoi constructor. Since the decorator was optimized for Point coordinates, it caused indexing errors with Polygons. 
- **Ghost-Cell Filtering:** Implemented index masking within `_voronoi` to filter out auxiliary "frame" geometries created during Voronoi clipping, preventing `IndexError` during ID mapping.

Closes #688.

### DIFF
--- a/libpysal/graph/_utils.py
+++ b/libpysal/graph/_utils.py
@@ -214,7 +214,7 @@ def _validate_geometry_input(geoms, ids=None, valid_geometry_types=None):
         geoms = geoms.copy()
         geoms.index = ids
         return coordinates, ids, geoms
-    
+
     # Check for geometry dtype only if geoms has a .dtype attribute
     # This prevents AttributeError on GeoDataFrames
     elif hasattr(geoms, "dtype") and isinstance(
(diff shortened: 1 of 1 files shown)


Draavith-DNA:
"I am closing this PR as I've encountered several downstream test failures related to the internal geometry validation logic and the triangulation decorator.

While I successfully implemented the Ghost-Index filtering and the **kwargs pass-through for voronoi_frames, the current _validate_coplanar decorator and _validate_geometry_input utility are strictly optimized for Point geometries, causing AttributeError and ValueError when Polygons are introduced.

I hope the research and the code here serve as a starting point for someone more familiar with the core utility architecture to finalize the fix for #688. Thanks for the guidance!"

PR #846, opened 2026-01-18
Added from_networkx() to Graph constructors

Ramji-Purwar:
2. [x] This pull request is directed to the `pysal/main` branch.
4. [ ] You have [assigned a

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -523,12 +523,14 @@ class Graph(SetOpsMixin):
             raise ImportError("NetworkX is required.") from None
 
         nodes = list(graph.nodes())
-        
+
         # Check if the specified weight attribute exists on edges
         if weight is not None:
             for _u, _v, edge_data in graph.edges(data=True):
                 if weight not in edge_data:
-                    raise ValueError(f"The weight attribute '{weight}' does not exist on all edges.")
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


libpysal/graph/base.py
```diff
@@ -491,6 +491,48 @@ def from_dicts(cls, neighbors, weights=None):
         head, tail, weight = _neighbor_dict_to_edges(neighbors, weights=weights)
         return cls.from_arrays(head, tail, weight)
 
+    @classmethod
+    def from_networkx(cls, graph, weight_col="weight"):
```
martinfleis:
```suggestion
    def from_networkx(cls, graph, weight=None):
```

Make it `weight` as in the context of networkx, there is no column.

Make the `weight` specification opt-in. Do not assume the name of the edge attribute. I am fully aware this differs from the implementation in W but I feel this could lead to unforeseen consequences of users not realising that some attribute is being silently used.

martinfleis (commented):
I left a few comments in the code and have one question. Can you test if `nx.to_pandas_adjacency` would not be faster? Given we are using adjacency as our internal representation, it seems unnecessary to go via sparse. But I'd be keen to see the performance comparison on somewhat larger graphs.

martinfleis (approved):

PR #847, opened 2026-01-18
Represent islands as (focal, focal, 0) in adjacency lists

### LINKED ISSUE #320: Represent islands as (focal, focal, 0) in adjacency lists
A call to `.format()` is missing [here](https://github.com/pysal/libpysal/blob/master/libpysal/weights/weights.py#L308), which should inform the user about how many islands are in the weights matrix. 

Currently, the message prints without telling the user how many islands there are in the dataframe.

Piyush0000:
This PR addresses #320 by preserving islands (observations with no neighbors)
in adjacency list representations.

### Summary of changes
- Allow self-loops with zero weight in `from_adjlist`
- Preserve islands during W → adjacency list → W round-trips
- Add tests covering island representation and round-trip integrity

### Behavior
- `to_adjlist(drop_islands=False)` represents islands as `(focal, focal, 0)`
- `to_adjlist(drop_islands=True)` preserves existing behavior
- Self-loops with non-zero weights remain filtered out

### Tests
- Added tests under `libpysal/weights/tests`
- Tests pass locally on Python 3.12 (Windows)

Closes #320

### DIFF
--- a/libpysal/weights/tests/test_adjlist.py
+++ b/libpysal/weights/tests/test_adjlist.py
@@ -23,7 +23,7 @@ class TestAdjlist:
 
     def test_round_trip_drop_islands_false(self):
         adjlist = self.knownW.to_adjlist(
-            remove_symmetric=False, drop_islands=True
+            remove_symmetric=False, drop_islands=False
         ).astype(int)
         w_from_adj = weights.W.from_adjlist(adjlist)
         np.testing.assert_allclose(


sjsrey:
Thank you for this.
However, we are transitioning from `W` to `Graph`: https://pysal.org/libpysal/user-guide/graph/w_g_migration.html and will be focusing all development efforts on `Graph`.

PR #853, opened 2026-01-21
Robust nodata detection in raster builders

Omiii-215:

- Handle NaNs explicitly in masking logic for float rasters
- Fix fallback path to respect generated masks (e.g. from NaN or rio.nodata)
- Add regression tests for NaN handling and rioxarray compatibility

### DIFF
--- a/libpysal/weights/raster.py
+++ b/libpysal/weights/raster.py
@@ -358,9 +358,17 @@ def _da2wsp(
     mask = None
 
     # Handle NaN masking for float rasters (explicit or implicit nodata)
-    if np.issubdtype(values.dtype, np.floating) and (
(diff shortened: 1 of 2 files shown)


martinfleis (changes requested):
Not sure I follow this properly but is there actually anything that would fail without these changes? Seems to me that it essentially does anything no? Masks out np.nan to replace it with np.nan...

Omiii-215:
mask = values != nodata failed for nan because np.nan != np.nan evaluates to True? because of this only i think nan pixels were to be treated as valid data, np.isnan(values) handles nan. The test failure confirms that without this fix, nans incorrectly leak into the weights object.

Issue #628, opened 2023-11-03
Allow plotting weight in Graph.plot


martinfleis:
A side note - I also realised that we can directly plot weight. Neat.
![Screenshot 2023-11-03 at 23 08 10](https://github.com/pysal/libpysal/assets/36797143/8f1ce4e2-72b5-42d2-a073-a5a7bcb311c5)

_Originally posted by @martinfleis in https://github.com/pysal/libpysal/issues/627#issuecomment-1793159245_
            

In explore, we get ability to plot the actual weight on a map for free but with static plots, we need to get matplotlib colormap, retrieve the colours and pass them to the `LineCollection` ourselves. Might be worth exploring if there's an appetite.

jGaboardi:
xref #562

jGaboardi:
@martinfleis This can now be closed out, correct?

knaaptime:
once you're [here](https://github.com/pysal/libpysal/blob/8c4a2447685ebc9ec10d19cd04844dd04e862bc1/libpysal/graph/_plotting.py#L111), you have an array of edges keyed by id? if you could bring along the weight values by going to a dataframe then joining or something, you could grab an [array of colors](https://pysal.org/mapclassify/generated/mapclassify.util.get_color_array.html#mapclassify.util.get_color_array) in the proper alignment and send that?

knaaptime:
yeah, though could also let the user choose whether to plot incoming values or outgoing values  with a kwarg then keep j-i or i-j and join weight values accordingly

PR #858, opened 2026-01-26
feat: add weight based coloring to static plots

satyanvm:
It adds the feature of coloring edges by their weights in static plots which was initiated in #628 
The method discussed in the issue is what is indeed being used in this PR. We get weights of the edges, normalize them to [0, 1] and then pass them to a matplotlib colormap and use the colormap to plot the different colors for different edges. I have used a color_by_weight boolean parameter, thus enabling backward compatibility and have also added a colorbar for weight to color mapping.
AI assistance was used for code development; all communication and explanations are written by me.


martinfleis (changes requested):
Hey, I am not a fan of this solution of mapping colours and creating a ScalarMappable to get a colorbar. I'd very much prefer passing an array to the collection and let matplotlib do the job.

PR #868, opened 2026-02-05
FIX: support Polygon/MultiPolygon in Voronoi graph builder (#688)

AliaaNasser7:
### Description
I've updated the `_voronoi` builder to support polygonal geometries by using their centroids as generators. This proposes a practical resolution for #688 and bypasses the limitations of the current point-based validation.
### Changes & Technical Implementation
- *Validation Bypass:* The `@_validate_coplanar` decorator was causing indexing issues and attribute errors with Polygons. I moved the coplanar check inside the function and used centroid extraction to ensure the math stays consistent with the Voronoi logic.
- *Handling Geometry Types:* I added a check for `.centroid` to handle GeoDataFrames directly. For other inputs, I'm using `get_points_array` with a fallback to reshape the array, ensuring we always pass an (n, 2) array to `voronoi_frames`.
- *Ghost Cell Filtering:* Clipping often creates auxiliary cells that mess up the input-to-output mapping. I fixed this by explicitly mapping the `heads_ix` and `tails_ix` back to the original `ids` before returning the graph.
- *Kwargs:* Passed `**kwargs` down to `voronoi_frames` to support parameters like `shrink` or `buffer`.
### Verification
I tested this with a GeoDataFrame of adjacent polygons. The builder now correctly identifies neighbors and maintains the original index without crashing or throwing indexing errors.

Related to #688

### DIFF
--- a/libpysal/graph/_triangulation.py
+++ b/libpysal/graph/_triangulation.py
@@ -342,8 +342,13 @@ def _relative_neighborhood(coordinates, coplanar):
     return heads_ix, tails_ix, coplanar
 
 
-@_validate_coplanar
-def _voronoi(coordinates, coplanar, clip="bounding_box", rook=True):
+
+def _voronoi(coordinates, ids=None, clip="None", **kwargs):   
+    from ._utils import _validate_geometry_input, CoplanarError
+    from ._contiguity import _vertex_set_intersection
+    from ..weights.util import get_points_array
+    from ..cg import voronoi_frames
+    import numpy 
     """
     Compute contiguity weights according to a clipped
     Voronoi diagram.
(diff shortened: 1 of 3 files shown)


martinfleis (changes requested):
This completely missed the point of #688. Have a look at #678 which it refers to and try to understand what the aim was there. This PR does not support Polygons in Voronoi builders, it silently downcasts them to centroids.

AliaaNasser7:
> This completely missed the point of #688. Have a look at #678 which it refers to and try to understand what the aim was there. This PR does not support Polygons in Voronoi builders, it silently downcasts them to centroids.

I reworked the PR completely ,instead of downcasting to centroids, _voronoi now passes geometries directly to voronoi_frames (which handles all geometry types since #678) and forwards segment/shrink kwargs.

martinfleis (commented):
It might be easier to keep the current `_voronoi` function intact with its coplanarity solutions and write a separate one for non-point geometry types. And just channel to the proper one when needed.

libpysal/graph/tests/test_triangulation.py
```diff
@@ -498,3 +500,84 @@ def test_wrong_resolver(self):
...
+    polys = [
+        Polygon([(0, 0), (1, 0), (1, 1), (0, 1)]),
+        Polygon([(1, 0), (2, 0), (2, 1), (1, 1)]),
+        Polygon([(0, 1), (1, 1), (1, 2), (0, 2)]),
+        Polygon([(1, 1), (2, 1), (2, 2), (1, 2)]),
+    ]
+    gdf = geopandas.GeoDataFrame(geometry=polys)
+    ids = np.arange(4)
+    heads, tails, weights = _voronoi_polygon(gdf, ids=ids)
+    assert len(np.unique(heads)) == 4
```
martinfleis: The tests should look like those above. Check the exact arrays, all of them.

libpysal/graph/tests/test_triangulation.py
```diff
@@ -289,7 +291,7 @@ def test_coplanar_raise_voronoi(stores):
 
 
 @pytest.mark.network
-def test_coplanar_jitter_voronoi(stores, stores_unique):
+def test_coplanar_jitter_voronoi(stores, stores_unique):  # edit
```
martinfleis:
```suggestion
def test_coplanar_jitter_voronoi(stores, stores_unique):
```

libpysal/graph/_triangulation.py
```diff
@@ -431,6 +431,64 @@ def _voronoi(coordinates, coplanar, clip="bounding_box", rook=True):
 #### utilities
 
 
+def _voronoi_polygon(
```
martinfleis: this should not be under "utilities" but above.

libpysal/graph/_triangulation.py
```diff
@@ -431,6 +431,64 @@ def _voronoi(coordinates, coplanar, clip="bounding_box", rook=True):
...
+
+    valid_ids = set(ids)
+    mask = numpy.isin(heads_ix, list(valid_ids)) & numpy.isin(tails_ix, list(valid_ids))
+    heads_ix = heads_ix[mask]
+    tails_ix = tails_ix[mask]
+
+    heads = heads_ix
+    tails = tails_ix
+    weights = numpy.ones(len(heads_ix), dtype=numpy.int8)
+    return heads, tails, weights
```
martinfleis:
```suggestion
    weights = numpy.ones(len(heads_ix), dtype=numpy.int8)
    return heads_ix, tails_ix, weights
```

libpysal/graph/base.py
```diff
@@ -1442,6 +1449,9 @@ def build_triangulation(
             or negative) at some very large (possibly infinite) distance.
             Otherwise, kernel functions are treated as proper
             volume-preserving probability distributions.
+        **kwargs: Additional keyword arguments passed to ``voronoi_frames()`` when
+        ``method="voronoi"``. Supports ``segment`` (float) and ``shrink``
+        (float). Ignored for other methods.
```
martinfleis: Only for non-point inputs.

libpysal/graph/base.py
martinfleis, line 1393: This needs to be updated to also mention other geometry types.

libpysal/graph/base.py
```diff
@@ -1519,15 +1529,24 @@ def build_triangulation(
...
-            head, tail, weights = _voronoi(
-                data,
-                ids=ids,
-                clip=clip,
-                rook=rook,
-                coplanar=coplanar,
-                decay=decay,
-                taper=taper,
-            )
+            if hasattr(data, "geom_type") and not set(data.geom_type) <= {"Point"}:
```
martinfleis: I think we should use `shapely.get_type_id` in here to check the geometry types for all inputs, not just geopandas ones.

libpysal/graph/tests/test_triangulation.py
AliaaNasser7: > The tests should look like those above. Check the exact arrays, all of them.

@martinfleis 
Addressed all review comments.
-> `test_voronoi_polygon_string_ids` uses graph-based neighbor comparison instead of exact array equality because Python's hash randomization causes string-keyed arrays to return in different orders across process.

Issue #909, opened 2026-04-11
ENH: use sparse distance computation for compact-support kernels in `Graph.build_kernel()`


pastephens:
## Summary

  `Graph.build_kernel()` currently computes a full N×N dense distance matrix for all kernel types,
  then applies the kernel function and discards zero weights. For kernels with compact support
  (bisquare, boxcar, triangular, tricube, cosine, parabolic), all weights beyond `bandwidth` are
  exactly zero — so the full matrix is never needed. Using `scipy.spatial.KDTree.sparse_distance_matrix`
  instead reduces memory from O(N²) to O(N × avg_neighbors) and gives large speedups at realistic
  dataset sizes.

  ## Benchmarks

  Tested against the current `_kernel()` implementation on real GeoDa datasets:

  | Dataset | n | sparse (ms) | dense (ms) | speedup |
  |---|---|---|---|---|
  | Southern US counties | 1,412 | 11 | 40 | 3.8x |
  | King County home sales | 21,613 | 111 | 352,398 | **~3,200x** |

  Gaussian and other infinite-support kernels are unaffected — they continue to use the full
  distance matrix, as before.

  At n=21,613 the dense path takes nearly 6 minutes; the sparse path takes 111ms. At larger
  point datasets the dense path will exhaust memory entirely.

  ## Proposed change

  Add a fast path in `_kernel()` that routes to `KDTree.sparse_distance_matrix` when:
  - `kernel` has compact support (bisquare, boxcar, triangular, tricube, cosine, parabolic/discrete)
  - `bandwidth` is a fixed numeric value (not `None` or `"auto"`)
  - `metric == "euclidean"`
  - `k is None`

  All other cases fall through to the existing code path unchanged. A pre-built `tree` is reused
  if one is passed and it is a `scipy.spatial.KDTree`.

  Note: the `bandwidth=None` default path (which computes bandwidth as the 25th percentile of all
  pairwise distances) still requires the full matrix and is not affected by this change. A natural
  follow-on would be a KNN-based bandwidth estimator that avoids the full matrix entirely, but that
  would be a behavior change and is left for a separate discussion.

  ## Implementation

  A working implementation is available on a fork branch: `pastephens/libpysal` @
  `sparse-kernel-weights`. The branch is based on current `main` (as of today) and all 55 kernel
  tests pass.

  Happy to open a PR if this direction is of interest.

martinfleis:
[...]

PR #911, opened 2026-04-18
Sparse kernel weights

pastephens:
The justification for this PR is at https://github.com/pysal/libpysal/issues/909

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,14 +77,6 @@ tests = [
     "pytest-xdist",
 ]
 
-[dependency-groups]
-dev = [
-    "geodatasets>=2026.1.0",
-    "geopandas>=1.1.3",
-    "pytest>=9.0.3",
-    "ruff>=0.15.10",
-]
-
 [tool.setuptools.packages.find]
 include = ["libpysal", "libpysal.*"]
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -78,6 +77,13 @@ tests = [
     "pytest-xdist",
 ]
 
+[dependency-groups]
```
jGaboardi: It's true that currently we are not using `project.optional-dependencies` & `dependency-groups` [as they are intended](https://pydevtools.com/handbook/explanation/what-are-optional-dependencies-and-dependency-groups/), but we should stick with what we have until a bigger decision is made.

So let's leave out the additional `[dependency-groups]` section in this PR.

martinfleis (approved):
Apart from @jGaboardi's point on changes to `pyproject.toml `which are not related to the contents of the PR, I think this is good to go. Nice job!

knaaptime:
could we extend this approach to any kernel that's truncated? e.g. we should also use the sparse distance if you ask for a gaussian with [`taper=True`](https://github.com/pysal/libpysal/blob/main/docs/source/user-guide/kernels.ipynb)

jGaboardi (approved):

knaaptime (approved):
sweet

jGaboardi:
thanks, @pastephens !!!

PR #912, opened 2026-05-04
ENH: add Graph.from_lattice

martinfleis:
This brings a builder analogous to `lat2W` to Graph as `Graph.from_lattice`.

### DIFF
--- a/libpysal/graph/base.py
+++ b/libpysal/graph/base.py
@@ -349,7 +349,7 @@ class Graph(SetOpsMixin):
                 [focal, neighbor], names=["focal", "neighbor"]
             ),
         )
-        return cls(adjacency)
+        return cls(adjacency, is_sorted=True)
 
     def to_W(self):  # noqa: N802
         """Convert Graph to a libpysal.weights.W object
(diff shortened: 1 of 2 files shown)


libpysal/graph/base.py
```diff
@@ -246,6 +246,111 @@ def from_W(cls, w):  # noqa: N802
...
+            Type of index IDs to use in the final Graph object.
+
+        Returns
+        -------
+        Graph
+            Graph encoding lattice contiguity.
+
+        Examples
+        --------
+        >>> Graph.from_lattice()
```
jGaboardi, line 272: Should we make an explicit `from libpysal import Graph` here so these examples can be copy-pasted and run?

martinfleis: I am using the model of [geopandas](https://github.com/geopandas/geopandas/blob/531228837c2ce4f13a1880267710f301cebb7d44/geopandas/geoseries.py#L99-L100) and [momepy](https://github.com/pysal/momepy/blob/461c6be59326cb871f61a1f206f407fb1e43d238/momepy/distribution.py#L43-L45) here, neither of which explicitly imports itself. Same applies to pandas or xarray.

jGaboardi (approved):

ljwolf (approved):

PR #914, opened 2026-05-18
GHA: use dedicated environment for docs action

martinfleis:
As discussed during the dev call, this implements dedicated docs environment for GHA, split away from those used for tests. It should be easier to maintain the envs this way. I am not sure if we striclty need all of this but it certainly won't hurt. And trying it is a pain :).

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest"]
-        environment-file: [ci/314-latest.yaml]
+        environment-file: [ci/docs.yaml]
         experimental: [false]
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


jGaboardi (approved):
* failures seem only to do with fetching example data
* doc build action succeeded
* spot-checked doc artifacts -- all looks good

knaaptime (approved):

knaaptime:
although you could argue the environment file might be more appropriate in docs/ than ci/?

jGaboardi:
> although you could argue the environment file might be more appropriate in docs/ than ci/?

This is a good point @knaaptime. I second moving it there.

martinfleis:
I don't mind either way. I thought this keeps all yamls used in actions at one place. Easier for SPEC0 bumping and so on.

jGaboardi:
> I don't mind either way. I thought this keeps all yamls used in actions at one place. Easier for SPEC0 bumping and so on.

Also a good point... LOL

jGaboardi:
However, there should probably be no pinned dependencies in the docs env. So we really only need to worry about the latest minor release of Python.

martinfleis:
Alright, moved.

jGaboardi (approved):

PR #915, opened 2026-05-21
BUG: fix situations where linear geometries disappear in voronoi_frames

martinfleis:
The deduplication step means that some lines completely disappear. We need to clip them on both sides to ensure that they all have their unique endpoints that can go to the voronoi algorithm. The `shapely.ops.substring ` works but is a loop since there's no vectorized version. I'll see if I can quickly craft one, otherwise I'll leave it. It is not that slow.

Tests are missing but this already fixes the case I ran into.

### DIFF
--- a/libpysal/cg/tests/test_voronoi.py
+++ b/libpysal/cg/tests/test_voronoi.py
@@ -95,6 +95,24 @@ class TestVoronoi:
         )
         assert_geoseries_equal(geoms.simplify(0.1), expected, check_less_precise=True)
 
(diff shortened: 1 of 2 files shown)


jGaboardi (approved):

jGaboardi:
@martinfleis -- I think @sjsrey may be out for some days for a fancy conference in Paris. Shall we get another dev approval here and merge?

@ljwolf @knaaptime @gegen07

gegen07 (approved):

PR #921, opened 2026-06-11
update path for raster test data

sjsrey:
The gds book repo was recently refactored. 
This updates the path to reflect this.


martinfleis:
Rasterio seems to be unhappy. It might be that the protocol is different than when served from a rendered site? Direct Github link might not support range requests or something.

darribas:
I think you need to link to the raw file, not the rendered GH page of the file int he repo?

https://github.com/gdsbook/book/raw/refs/heads/main/data/ghsl/ghsl_sao_paulo.tif

martinfleis (approved):
Thanks @darribas for the fix!

PR #922, opened 2026-06-15
GHA: fix reverse dependency testing

martinfleis:
This ensures that the action properly installs the env. That is currently not the case.

In the [successful run](https://github.com/martinfleis/libpysal/actions/runs/27530337870), I see segregation and spint failing. It might be worth exploring why is that and whether it has been fixed on main in the meantime.

### DIFF
--- a/.github/workflows/reverse.yml
+++ b/.github/workflows/reverse.yml
@@ -1,7 +1,8 @@
 name: Test reverse dependencies
 on:
   push:
-    branches: [main]
+    branches:
+      - "*"
   schedule:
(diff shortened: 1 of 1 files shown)


martinfleis:
Failures being fixed in #921

jGaboardi:
@martinfleis -- So should we wait to merge this or go ahead and merge since #921 is closed out?

martinfleis:
@jGaboardi this can be merged now.

sjsrey (approved):

PR #924, opened 2026-06-15
feat: support array-like bandwidth in kernel functions

sjsrey:
- document that bandwidth can be float or ndarray
- add tests
- type annotations

### DIFF
--- a/libpysal/kernels.py
+++ b/libpysal/kernels.py
@@ -323,6 +323,8 @@ def kernel(
         k[distances > bandwidth] = 0.0
 
     elif isinstance(taper, (float, int)) and not isinstance(taper, bool):
+        # pandas 3.0 Copy-on-Write consequence
+        k = k.copy()
         k[distances > taper] = 0.0
 
     if decay:
--- a/libpysal/kernels.py
+++ b/libpysal/kernels.py
@@ -117,11 +117,10 @@ def _gaussian(distances: np.ndarray, bandwidth) -> np.ndarray:
     ndarray
         Gaussian kernel weights.
     """
-    if isinstance(bandwidth, (int, float)):
-        z = distances / bandwidth
-    else:
-        bandwidth = np.asarray(bandwidth)
-        z = distances / bandwidth
+    if not isinstance(bandwidth, (int, float)):
+       bandwidth = np.asarray(bandwidth)
+        
+    z = distances / bandwidth
     exponent_term = -0.5 * (z**2)
     c = 1 / np.sqrt(2 * np.pi)
     k = c * np.exp(exponent_term)
--- a/libpysal/kernels.py
+++ b/libpysal/kernels.py
@@ -317,11 +317,10 @@ def kernel(
     k = func(distances, bandwidth)
 
     if taper is True:
-        if isinstance(bandwidth, (int, float)):
-            k[distances > bandwidth] = 0.0
-        else:
+        if not isinstance(bandwidth, (int, float)):
             bandwidth = np.asarray(bandwidth)
-            k[distances > bandwidth] = 0.0
+
+        k[distances > bandwidth] = 0.0
 
     elif isinstance(taper, (float, int)) and not isinstance(taper, bool):
         k[distances > taper] = 0.0
--- a/libpysal/kernels.py
+++ b/libpysal/kernels.py
@@ -203,11 +203,10 @@ def _exponential(distances: np.ndarray, bandwidth) -> np.ndarray:
     ndarray
         Exponential kernel weights.
     """
-    if isinstance(bandwidth, (int, float)):
-        z = distances / bandwidth
-    else:
+    if not isinstance(bandwidth, (int, float)):
         bandwidth = np.asarray(bandwidth)
-        z = distances / bandwidth
+        
+    z = distances / bandwidth
     return np.exp(-z)
 
 
--- a/libpysal/kernels.py
+++ b/libpysal/kernels.py
@@ -56,11 +56,10 @@ def _trim(distances: np.ndarray, bandwidth) -> np.ndarray:
     ndarray
         Clipped and normalized distances.
     """
-    if isinstance(bandwidth, (int, float)):
-        return np.clip(np.abs(distances) / bandwidth, 0, 1)
-    else:
+    if not isinstance(bandwidth, (int, float)):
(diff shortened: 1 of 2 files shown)


libpysal/kernels.py
```diff
@@ -37,38 +37,42 @@
...
     -------
     ndarray
         Clipped and normalized distances.
     """
-    return numpy.clip(numpy.abs(d) / bandwidth, 0, 1)
+    if isinstance(bandwidth, (int, float)):
+        return np.clip(np.abs(distances) / bandwidth, 0, 1)
+    else:
+        bandwidth = np.asarray(bandwidth)
+        return np.clip(np.abs(distances) / bandwidth, 0, 1)
```
martinfleis:
```suggestion
    if not isinstance(bandwidth, (int, float)):
        bandwidth = np.asarray(bandwidth)

    return np.clip(np.abs(distances) / bandwidth, 0, 1)
```

This might be easier. Do I understand the logic here correctly that this is just an array-like sanitisation? Ensuring that if bandwidth is not a numpy.array but some other sequence, we use it properly anyway? Coz with array the original code works...

libpysal/kernels.py
```diff
@@ -98,130 +102,142 @@ def _parabolic(distances, bandwidth):
...
     -------
     ndarray
         Gaussian kernel weights.
     """
-    z = distances / bandwidth
+    if isinstance(bandwidth, (int, float)):
+        z = distances / bandwidth
+    else:
+        bandwidth = np.asarray(bandwidth)
+        z = distances / bandwidth
```
martinfleis:
```suggestion
    if not isinstance(bandwidth, (int, float)):
       bandwidth = np.asarray(bandwidth)
        
    z = distances / bandwidth
```

Same pattern as above. Removes the repetition.

libpysal/kernels.py
```diff
@@ -98,130 +102,142 @@ def _parabolic(distances, bandwidth):
...
     ndarray
         Exponential kernel weights.
     """
-    z = distances / bandwidth
-    return numpy.exp(-z)
+    if isinstance(bandwidth, (int, float)):
+        z = distances / bandwidth
+    else:
+        bandwidth = np.asarray(bandwidth)
+        z = distances / bandwidth
```
martinfleis:
```suggestion
    if not isinstance(bandwidth, (int, float)):
        bandwidth = np.asarray(bandwidth)
        
    z = distances / bandwidth
```

libpysal/kernels.py
```diff
@@ -98,130 +102,142 @@ def _parabolic(distances, bandwidth):
...
         Binary weights: 1 if distance < bandwidth, else 0.
     """
-    distances = numpy.asarray(distances)
-    return (distances < bandwidth).astype(float)
+    distances = np.asarray(distances)
+    if isinstance(bandwidth, (int, float)):
+        return (distances < bandwidth).astype(float)
+    else:
+        bandwidth = np.asarray(bandwidth)
+        return (distances < bandwidth).astype(float)
```
martinfleis:
```suggestion
    if not isinstance(bandwidth, (int, float)):
        bandwidth = np.asarray(bandwidth)
    
    return (distances < bandwidth).astype(float)
```

libpysal/kernels.py
```diff
@@ -255,15 +271,18 @@ def _identity(distances, _):
 }
 
 
-def kernel(distances, bandwidth, kernel="gaussian", taper=True, decay=False):
-    """Evaluate a kernel function over a distance array.
+def kernel(
+    distances: np.ndarray, bandwidth, kernel="gaussian", taper=True, decay=False
```
martinfleis, line 271: I think that this type hint is wrong as this shall work on any array-like object, like Series.

libpysal/kernels.py
```diff
@@ -286,33 +305,35 @@ def kernel(distances, bandwidth, kernel="gaussian", taper=True, decay=False):
...
+
     if taper is True:
-        # pandas 3.0 Copy-on-Write consequence
-        k = k.copy()
-        k[distances > bandwidth] = 0.0
+        if isinstance(bandwidth, (int, float)):
+            k[distances > bandwidth] = 0.0
+        else:
+            bandwidth = np.asarray(bandwidth)
+            k[distances > bandwidth] = 0.0
```
martinfleis:
```suggestion
        if not isinstance(bandwidth, (int, float)):
            bandwidth = np.asarray(bandwidth)

        k[distances > bandwidth] = 0.0
```

libpysal/kernels.py
```diff
@@ -286,33 +305,35 @@ def kernel(distances, bandwidth, kernel="gaussian", taper=True, decay=False):
...
+
     elif isinstance(taper, (float, int)) and not isinstance(taper, bool):
-        # pandas 3.0 Copy-on-Write consequence
-        k = k.copy()
         k[distances > taper] = 0.0
```
martinfleis, line 328:
```suggestion
        # pandas 3.0 Copy-on-Write consequence
        k = k.copy()
        k[distances > taper] = 0.0
```

This needs to be there.

libpysal/kernels.py
```diff
@@ -286,33 +305,35 @@ def kernel(distances, bandwidth, kernel="gaussian", taper=True, decay=False):
...
 
     if decay:
-        k /= func(0.0, bandwidth)
+        if isinstance(bandwidth, (int, float)):
+            k /= func(0.0, bandwidth)
+        else:
+            k /= func(0.0, np.mean(bandwidth))
```
martinfleis, line 334: Can you elaborate on this? That does not seem right, I would assume we normalise by the local bandwidth (as it was), not a fixed value across the entire dataset. At least that is how I expect it to behave within spatialml.

martinfleis (commented):
It is a bit unclear to me what is this doing. Is there a code path this fixes? I don't mind the changes, it seems to be more robust overall, but I am wondering what is the motivation. I am also not sure these changes are appropriately tested.

libpysal/kernels.py
sjsrey: Good catch. Reverting this one back to original treatment.

sjsrey: Actually, I think it is best to close this pr and have a discussion of what behavior is needed in spatialml (during the gsoc call?)

sjsrey:
Closing. Will open a new pr following further discussions.

Issue #932, opened 2026-07-17
coplanar does not work in adaptive bandwidths


ljwolf:
Working to simplify and modernize over in ljwolf/spml#2 and I discovered this bug. 

```python
import numpy
from libpysal.graph import Graph

# two coincident points at (0, 0), plus two distinct points
coords = numpy.array([[0., 0.], [0., 0.], [1., 0.], [5., 0.]])

Graph.build_kernel(
    coords, k=1, bandwidth="adaptive", kernel="gaussian", coplanar="clique"
)
```

```
Traceback (most recent call last):
  ...
  File ".../libpysal/graph/_kernel.py", line 266, in _kernel
    numpy.maximum.at(bw_per_row, rows, d_csr.data)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: array is not broadcastable to correct shape
```

Fixed `bandwidth` values combined with `coplanar="clique"` work fine still. 

```python
Graph.build_kernel(coords, k=1, bandwidth=2.0, kernel="gaussian", coplanar="clique")
# no error
```

and `bandwidth="adaptive"` works fine without coplanar points:

```python
Graph.build_kernel(
    numpy.array([[0., 0.], [1., 0.], [2., 0.], [10., 0.], [11., 0.], [12., 0.]]),
    k=2, bandwidth="adaptive", kernel="gaussian", coplanar="raise",
)
# no error
```

## Root cause

The `bandwidth == "adaptive"`computes each observation's per-row bandwidth using `.nonzero()`:

```python
d_csr = d.tocsr()
rows, _ = d_csr.nonzero()
bw_per_row = numpy.zeros(d_csr.shape[0])
numpy.maximum.at(bw_per_row, rows, d_csr.data)
bandwidth = bw_per_row[rows] * 1.0000001
```

This assumes that there are no structural zeros, which is generally incorrect. We will sometimes have structural zero weight values, as might happen when storing isolates *and* cliques. 

Hence, when we have `coplanar="clique"`, coincident points are clique-connected at distance 0. This means dges get an explicit **stored zero** in the sparse matrix:

```python
sparse_out.data[sparse_out.data < 0] = 0   # clique fill-value -1 -> 0
return sparse_out                          # correctly does not .eliminate_zeros()
```

[...]

PR #933, opened 2026-07-17
fix adaptive bandwidth crashes when combined with coplanar='clique'

ljwolf:
This addresses #932. See that issue for replication information and discussion.


sjsrey (approved):
Nice catch and fix.

martinfleis (approved):

jGaboardi (commented):
[postmortem approval]

Issue #475, opened 2022-07-27
libpysal.weights.DistanceBand returns warning the first time of use


Sieboldianus:
I am following the API docs for [esda.G_Local](https://pysal.org/esda/generated/esda.G_Local.html#esda.G_Local), to generate a weights object, to be used in Getis Ord Gi* Statistics, for classification of Hot and Cold Spots. It works, but what is strange is that I get a warning the first time I execute `libpysal.weights.DistanceBand()`. This warning disappears once executed again. If I restart python, the warning is back.

```
distance_threshold = 350
w = libpysal.weights.DistanceBand(points, threshold=distance_threshold, binary=False)
> /opt/conda/envs/worker_env/lib/python3.9/site-packages/scipy/sparse/_data.py:117: RuntimeWarning: divide by zero encountered > in reciprocal
>  return self._with_data(data ** n)
```

See the full notebook [here](https://ad.vgiscience.org/tagmaps-mapnik-jupyter/01_mapnik-tagmaps.html).

- Platform information:
- [JupyterLab from conda-forge in Docker](https://gitlab.vgiscience.de/lbsn/tools/jupyterlab)
```python
import os; print(os.name, os.sys.platform);print(os.uname())
> posix linux
> posix.uname_result(sysname='Linux', nodename='93b178f92cc0', release='5.10.104-linuxkit', version='#1 SMP Thu Mar 17 17:08:06 > UTC 2022', machine='x86_64')
```
- Python version: 
```python
import sys; print(sys.version)
> 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:56:21) 
[GCC 10.3.0]
```
- SciPy version:
```python
import scipy; print(scipy.__version__)
> 1.8.1
```
- NumPy version:
```python
import numpy; print(numpy.__version__)
> 1.22.4
```
- libpysal
```python
import libpysal; print(libpysal.__version__)
> 4.6.2
```

Sample data:
[shapefile.zip](https://github.com/pysal/libpysal/files/9196130/shapefile.zip)

```python
points = []
weights = []
with fiona.open('allLocationCluster_Size.shp', encoding='UTF-8', mode="r") as shapefile:
    for feature in locations:
        point = (feature["geometry"]['coordinates'][0], feature["geometry"]['coordinates'][1])
        points.append(point)
        weights.append(float(feature["properties"]["Join_Count"]))
points[:5]
> [(409779.71121669677, 5653978.9132233495),
[...]

PR #936, opened 2026-08-04
Fix/distance band divide by zero warning

sjsrey:
Explicit zero-valued self-distance entries in the sparse distance matrix were being raised to a negative power, producing inf and a RuntimeWarning before being discarded. Exclude those entries before exponentiating instead of after.

In _kernel.py this also fixes exclude_self_weights being silently ignored for metric="precomputed", which let self-loop weights leak through as inf. build_distance_band's diagonal-drop/isolate heuristic in base.py relied on that bug (self-loops present for every node) to tell isolates apart from real neighbors, so it's simplified now that _kernel already excludes self-loops correctly.

xref https://github.com/pysal/libpysal/issues/475#issuecomment-5181854370

### DIFF
--- a/libpysal/cg/tests/test_geoJSON.py
+++ b/libpysal/cg/tests/test_geoJSON.py
@@ -12,7 +12,7 @@ class TesttestMultiPloygon:
 
         """
 
-        shp = FileIO(pysal_examples.get_path("NAT.shp"), "r")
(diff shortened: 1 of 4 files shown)


ljwolf:
How can we formulate a test for this instance? I think we need a standard coincident point test set... I'd suggest one or both from #938?

